PR #11, opened 2020-07-20
add gbaloc ll

anporumb:


common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
+gballoc design
+=====
+
+`gballoc` is a layer that abstrats memory allocators. The need to have and be able to compare multiple memory allocators stems from performance requirements. Not all memory allocators have the same performance and for the sake of being able to compare them gballoc exists.
```
amarsa-msft, line 4: >abstrats [](start = 26, length = 8)

abstracts

common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
...
+
+`gballoc_hl` consists of a header file (`gballoc_hl.h`) and several possible implementation of the performance functions(i.e. `gballoc_hl_passthrough.c`, `gballoc_hl_buckets.c` etc)
+
+Both `gballoc_ll` and `gballoc_hl` have their own set of orthogonal configurations in CMake. That is - it is possible to use any combination of `gballoc_ll` implementation with any other implementation of `gballoc_ll`. 
+
+As far as Windows is concerned there are several CMake options.
+
+  1. `USE_SEGMENTED_HEAP` - this is a process wide option that replaces all the "regular" heaps with segmented heaps (https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#heaptype)
+
+  2. One a heap type is used, there are other several CMakeLists switches that further influence the SW behavior.
```
amarsa-msft, line 21: Once

common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
...
+
+`gballoc_ll` consists of a header file (`gballoc_ll.h`) and several possible implementation of its functions (i.e.: `gballoc_ll_crt.c`, `gballoc_ll_win.c`, `gballoc_ll_mimalloc.c` etc).
+
+`gballoc_hl` consists of a header file (`gballoc_hl.h`) and several possible implementation of the performance functions(i.e. `gballoc_hl_passthrough.c`, `gballoc_hl_buckets.c` etc)
+
+Both `gballoc_ll` and `gballoc_hl` have their own set of orthogonal configurations in CMake. That is - it is possible to use any combination of `gballoc_ll` implementation with any other implementation of `gballoc_ll`. 
+
+As far as Windows is concerned there are several CMake options.
+
+  1. `USE_SEGMENTED_HEAP` - this is a process wide option that replaces all the "regular" heaps with segmented heaps (https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#heaptype)
```
amarsa-msft, line 19: It is not nice that we have both CMAKE switches and `ll` layers for controlling the type of allocations

common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
...
+
+  2. One a heap type is used, there are other several CMakeLists switches that further influence the SW behavior.
+
+    a) `gballoc_ll` implementation is switched between its implementations by the CMake option `GBALLOC_LL_TYPE` (of type string) which can be either
+
+      i. "PASSTHROUGH" - `gballoc_ll` shall route its calls directly to C's routines.
+
+      ii. "WIN32HEAP" - `gballoc_ll` shall route its calls directly to Windows routines
```
amarsa-msft, line 27: >WIN32HEAP [](start = 11, length = 9)

I'd have a WIN32_SEGMENTD_HEAP implementation here instead of a separate `USE_SEGMENTED_HEAP` switch.

common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
...
+      i. "PASSTHROUGH" - `gballoc_ll` shall route its calls directly to C's routines.
+
+      ii. "WIN32HEAP" - `gballoc_ll` shall route its calls directly to Windows routines 
+        (`HeapAlloc`, `HeapReAlloc`, `HeapFree`)
+
+      iii. "MIMALLOC" - `gballoc_ll` shall route its calls directly to `mimalloc` APIs.
+
+    b) `gballoc_hl` implementation is governed by `GBALLOC_HL_TYPE` (of type string) which can be either:
+
+      i. "PASSTHROUGH" - `gballoc_hl` shall route all its APIs directly to `gballoc_ll`.
```
amarsa-msft, line 34: why is this a layer? it looks like we're just turning metrics on or off here.

common/devdoc/gballoc_design.md
```diff
@@ -0,0 +1,40 @@
...
+
+As far as Windows is concerned there are several CMake options.
+
+  1. `USE_SEGMENTED_HEAP` - this is a process wide option that replaces all the "regular" heaps with segmented heaps (https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#heaptype)
+
+  2. One a heap type is used, there are other several CMakeLists switches that further influence the SW behavior.
+
+    a) `gballoc_ll` implementation is switched between its implementations by the CMake option `GBALLOC_LL_TYPE` (of type string) which can be either
+
+      i. "PASSTHROUGH" - `gballoc_ll` shall route its calls directly to C's routines.
```
amarsa-msft, line 25: >PASSTHROUGH [](start = 10, length = 11)

NIT can we have this as CRT? there is also a `PASSTHROUGH` below with a totally different meaning ...

common/devdoc/gballoc_ll_mimalloc.md
```diff
@@ -0,0 +1,90 @@
...
+MOCKABLE_FUNCTION(, void, gballoc_ll_deinit);
+```
+
+`gballoc_ll_deinit` returns.
+
+**SRS_GBALLOC_LL_MIMALLOC_02_002: [** `gballoc_ll_deinit` shall return. **]**
+
+### gballoc_ll_malloc
+```c
+MOCKABLE_FUNCTION(, void*, gballoc_ll_malloc, size_t, size);
```
amarsa-msft, line 45: This is a golden opportunity to log the size of buffer to be allocated when this fails .. rather than logging it in every module that uses malloc. I'll buy you a coffee and an extra bike tire if you fix this :smile:

common/devdoc/gballoc_ll_win32heap.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+### gballoc_ll_init
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_init, void*, params);
+```
+
+`gballoc_ll_init` globally initializes the module. `params` exists as a placeholder and is ignored.
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_001: [** `gballoc_ll_init` shall call `CreateHeap(0,0,0)`. **]**
```
amarsa-msft, line 31: >CreateHeap [](start = 69, length = 10)

I think this is `HeapCreate`

common/devdoc/gballoc_ll_passthrough.md
```diff
@@ -0,0 +1,78 @@
...
+
+gballoc_ll_passthrough is a module that delegates all call of its APIs to the ones from C standard lib.
+
+## References
+
+
+## Exposed API
+
+```c
+    MOCKABLE_FUNCTION(, int, gballoc_ll_init, void*, params);
```
amarsa-msft, line 14: >gballoc_ll_init [](start = 29, length = 15)

This design assumes forces implementors and users to have one heap for the entire process. IMO, the interface should support creation and usage of multiple heaps. I'm thinking malloc, free, calloc, realloc can take as input a heap handle as well. If it is null, they can use the default heap.

common/devdoc/gballoc_ll_win32heap.md
```diff
@@ -0,0 +1,107 @@
...
+### gballoc_ll_deinit
+```c
+MOCKABLE_FUNCTION(, void, gballoc_ll_deinit);
+```
+
+`gballoc_ll_deinit` deinitializes the global state and frees all the used resources.
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_016: [** If the global state is not initialized then `gballoc_ll_deinit` shall return. **]**
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_004: [** `gballoc_ll_deinit` shall call `HeapDestroy`. **]**
```
amarsa-msft, line 46: >HeapDestroy [](start = 71, length = 11)

On what? who stores the handle?

common/devdoc/gballoc_ll_win32heap.md
```diff
@@ -0,0 +1,107 @@
...
+**SRS_GBALLOC_LL_WIN32HEAP_02_003: [** If `CreateHeap` fails then `gballoc_ll_init` shall fail and return a non-0 value. **]**
+
+### gballoc_ll_deinit
+```c
+MOCKABLE_FUNCTION(, void, gballoc_ll_deinit);
+```
+
+`gballoc_ll_deinit` deinitializes the global state and frees all the used resources.
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_016: [** If the global state is not initialized then `gballoc_ll_deinit` shall return. **]**
```
amarsa-msft, line 44: >If the global state is not initialized [](start = 39, length = 38)

Spec for "initializing global state" missing.

amarsa-msft, line 44: >initialized [](start = 66, length = 11)

spec for "deinitialization of global state" missing.

common/devdoc/gballoc_ll_win32heap.md
```diff
@@ -0,0 +1,107 @@
...
+MOCKABLE_FUNCTION(, void*, gballoc_ll_realloc, void*, ptr, size_t, size);
+```
+
+`gballoc_ll_realloc` reallocates `ptr` to have size `size`.
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_013: [** If the global state is not initialized then `gballoc_ll_realloc` shall return `NULL`. **]**
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_014: [** If `ptr` is `NULL` then `gballoc_ll_realloc` shall call `HeapAlloc` and return what `HeapAlloc` returns. **]**
+
+**SRS_GBALLOC_LL_WIN32HEAP_02_015: [** If `ptr` is not `NULL` then `gballoc_ll_realloc` shall call `HeapReAlloc` and return what `HeapAlloc` returns. **]**
```
amarsa-msft, line 99: `HeapReAlloc`

common/tests/gballoc_ll_mimalloc_int/gballoc_ll_mimalloc_int.c
```diff
@@ -0,0 +1,126 @@
...
+
+TEST_FUNCTION(gballoc_ll_deinit_works)
+{
+    ///act
+    gballoc_ll_deinit();
+
+    ///assert - doesn't crash
+}
+
+TEST_FUNCTION(gballoc_ll_malloc_works)
```
amarsa-msft, line 62: just for sanity, test alloc a 1MB buffer too

common/tests/gballoc_ll_mimalloc_ut/gballoc_ll_mimalloc_ut.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/*Tests_SRS_GBALLOC_LL_MIMALLOC_02_001: [ gballoc_ll_init shall return 0. ]*/
+TEST_FUNCTION(gballoc_ll_init_returns_0)
+{
+    ///arrange
+    int result;
+
+    ///act
+    result = gballoc_ll_init(NULL);
```
amarsa-msft, line 89: and what happens if the parameters are non-NULL?

common/devdoc/gballoc_design.md
anporumb: abstracted.

---
In reply to: [457670558](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457670558) [](ancestors = 457670558)

common/devdoc/gballoc_design.md
anporumb: Once'd

---
In reply to: [457672175](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457672175) [](ancestors = 457672175)

common/devdoc/gballoc_design.md
anporumb: And that is the precise intention of it: catching all the calls and constructing metrics out of them. (or not). So 2 layers:
1) no metrics - passthrough
2) with metrics

gballoc_hl can sit on top of any other gballoc_ll 

---
In reply to: [457675087](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457675087) [](ancestors = 457675087)

common/devdoc/gballoc_design.md
anporumb: We only use the switches. Switches dictate with of the LL/HL layers are to be used here. Switches will select one of the layers that gets compiled in the library. And it is one and only one of them.

---
In reply to: [457672996](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457672996) [](ancestors = 457672996)

common/devdoc/gballoc_design.md
anporumb: Those are different concerns at different layers.

"what type of heap" will influence (by manifest) all the heaps in the process (ours, SF's, and who else's). The process doesn't get to chose "for which parts of the process it uses segmented / for which it uses regular heaps". They are either ALL segmented or ALL regular/vanilla/old. 

We have here a heap implementation that separates all of our calls to our own heap (there's no Win32 APIs that lets you create a heap of a type or the other).
The heap type is given by manifest => SEGMENTED / REGULAR.



---
In reply to: [457674431](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457674431) [](ancestors = 457674431)

common/devdoc/gballoc_design.md
anporumb: I would keep it "as is" APIs are called gballoc_ll_bla_bla_bla which calls (by passthrough) bla_bla_bla. 

And the HL layer's APIs of gballoc_hl_bla_bla_bla call (when passthrough) gballoc_ll_bla_bla_bla (lower layer).

---
In reply to: [457676593](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457676593) [](ancestors = 457676593)

common/tests/gballoc_ll_mimalloc_ut/gballoc_ll_mimalloc_ut.c
anporumb: As explained in the .md "`gballoc_ll_init` returns. `params` exists as a placeholder and is ignored."

---
In reply to: [457702426](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457702426) [](ancestors = 457702426)

common/tests/gballoc_ll_mimalloc_int/gballoc_ll_mimalloc_int.c
anporumb: ok

---
In reply to: [457701994](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457701994) [](ancestors = 457701994)

anporumb: added here and in all the int tests.

---
In reply to: [457761934](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457761934) [](ancestors = 457761934,457701994)

common/devdoc/gballoc_ll_win32heap.md
anporumb: Correct. Replaced here and everywhere

---
In reply to: [457682925](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457682925) [](ancestors = 457682925)

common/devdoc/gballoc_ll_win32heap.md
anporumb: HeapRealloc'd here and everywhere

---
In reply to: [457688016](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457688016) [](ancestors = 457688016)

common/tests/gballoc_ll_mimalloc_ut/gballoc_ll_mimalloc_ut.c
amarsa-msft: Can we have a test with non-NULL?

anporumb: Yes, we can. Here and everywhere added.

---
In reply to: [457763832](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457763832) [](ancestors = 457763832)

common/devdoc/gballoc_ll_mimalloc.md
anporumb: I believe you're selling yourself cheap here because printing the size here bring little information. All it would print would be "failed to allocate 20 bytes". Now... what's the meaning of these 20 bytes, and where do they come from, that is left as an exercise to the user... I mean without a call stack / previous/next messages the message is meaningless. And even when you'd have the previous/next LogErrors, the context where it was allocated still matter. I suppose it is better to keep all the information in one LogError, but... 

I need that tire. I think I would like to try Conti Gatorskins :)

So. Added logerrors here and everywhere... when is my tire coming? :)

---
In reply to: [457679818](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457679818) [](ancestors = 457679818)

common/devdoc/gballoc_ll_passthrough.md
anporumb: "we" will have 1 heap per process. With this interface, that's all we're going to get. We might get more "heaps" in a later iteration of the code. Added a task to consider it

"others" are free to create/to not create their own heaps. Task 7745659: Consider adding "heap" as parameter to all allocations in gballoc_ll/hl

---
In reply to: [457685081](https://github.com/Azure/azure-c-pal/pull/11#discussion_r457685081) [](ancestors = 457685081)

common/devdoc/gballoc_ll_win32heap.md
anporumb: Added clarification

common/devdoc/gballoc_ll_win32heap.md
anporumb: Added verbiage that explains it is stored in a global variable

anporumb: Becomes clear with the addition of the global variable explictly stated.

common/devdoc/gballoc_ll_mimalloc.md
amarsa-msft: The user will log all failures in their code anyway, so it should be a pretty easy thread to follow. 
For e.g. malloc logs -> malloc 12345678910 bytes failed
User logs  "malloc failed" .. put two together and you have your data.

PR #14, opened 2020-07-23
have hl with metrics

anporumb:


win32/devdoc/gballoc_ll_passthrough.md
```diff
@@ -76,3 +78,13 @@ MOCKABLE_FUNCTION(, void*, gballoc_ll_realloc, void*, ptr, size_t, size);
...
 
 
+### gballoc_ll_size
+```c
+MOCKABLE_FUNCTION(, size_t, gballoc_ll_size, void*, ptr);
+```
+
+`gballoc_ll_size` returns 0 because stdlib has no API that provides back to the user the size of the allocated block.
+
+**SRS_GBALLOC_LL_PASSTHROUGH_02_007: [** `gballoc_ll_size` shall return what `_msize` returns. **]**
```
mattdurak, line 88: return 0 or return "what _msize returns"?

CMakeLists.txt
```diff
@@ -90,10 +90,10 @@ endif()
 
 if (
     (NOT TARGET mimalloc) AND 
-    (EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/mimalloc/CMakeLists.txt) AND
-    (${ARCHITECTURE} STREQUAL "x86_64")
+    (EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/mimalloc/CMakeLists.txt)
     )
-        set(MI_BUILD_TESTS OFF)
+        set(MI_BUILD_SHARED OFF CACHE BOOL "dont build a dll" FORCE) #not building a dll allows building on 32 bit, otherwise there's some errors on init.c about not finding a imported symbol
```
dcristoloveanu, line 95: >dont build a dll [](start = 44, length = 16)

Nitpick: I guess the description pertains to MI_BUILD_SHARED, so it should sound like "build a dll?"
Same applies for the below flag about building the tests.

win32/src/gballoc_hl_metrics.c
```diff
@@ -0,0 +1,435 @@
...
+
+static int wasInitialized = 0;
+
+int gballoc_hl_init(void* hl_params, void* ll_params)
+{
+    (void)hl_params;
+    int result;
+
+    /* Codes_SRS_GBALLOC_HL_METRICS_01_001: [ If the module is already initialized, gballoc_hl_init shall fail and return a non-zero value. ]*/
+    if (wasInitialized != 0)
```
mattdurak, line 136: >wasInitialized [](start = 8, length = 14)

I think the spec should mention that init/deinit are not thread-safe

interfaces/inc/azure_c_pal/gballoc_ll.h
```diff
@@ -25,6 +25,8 @@ extern "C"
     MOCKABLE_FUNCTION(, void*, gballoc_ll_calloc, size_t, nmemb, size_t, size);
     MOCKABLE_FUNCTION(, void*, gballoc_ll_realloc, void*, ptr, size_t, size);
 
+    MOCKABLE_FUNCTION(, size_t, gballoc_ll_size, void*, ptr);
```
dcristoloveanu, line 28: >gballoc_ll_size [](start = 32, length = 15)

Nice!

dcristoloveanu:
> `gballoc` is dvided in 2 layers (`ll` stand for `lower layer`, `hl` stands for `higher layer`):

stands

---
Refers to: win32/devdoc/gballoc_design.md:6 in 39f4b3f. [](commit_id = 39f4b3f10160f2d9e086919ea62ac74d751f9333, deletion_comment = False)

win32/src/gballoc_hl_passthrough.c
```diff
@@ -0,0 +1,166 @@
...
+        result = 0;
+        wasInitialized = 1;
+    }
+
+    return result;
+}
+
+void gballoc_hl_deinit(void)
+{
+    /*no work for this layer, it is passthough*/
```
mattdurak, line 40: >/*no work for this layer, it is passthough*/ [](start = 4, length = 44)

nit: Comment sort of applies to everything here (technically this pass-through layer does add logging and even this function enforces calling init/deinit once). Maybe the comment is not really needed

dcristoloveanu:
> `gballoc` is dvided in 2 layers (`ll` stand for `lower layer`, `hl` stands for `higher layer`):

divided

---
Refers to: win32/devdoc/gballoc_design.md:6 in 39f4b3f. [](commit_id = 39f4b3f10160f2d9e086919ea62ac74d751f9333, deletion_comment = False)

win32/tests/gballoc_hl_metrics_wout_init_ut/gballoc_hl_metrics_wout_init_ut.c
```diff
@@ -0,0 +1,165 @@
...
+{
+    int result;
+
+    if (TEST_MUTEX_ACQUIRE(test_serialize_mutex))
+    {
+        ASSERT_FAIL("Could not acquire test serialization mutex.");
+    }
+
+    umock_c_reset_all_calls();
+    result = umock_c_negative_tests_init();
```
dcristoloveanu, line 65: >result = umock_c_negative_tests_init(); [](start = 4, length = 39)

Nitpick: consider not having result and directly asserting.

win32/tests/gballoc_hl_metrics_ut/gballoc_hl_metrics_ut.c
```diff
@@ -0,0 +1,1491 @@
...
+#include "umock_c/umock_c_negative_tests.h"
+
+#define ENABLE_MOCKS
+#include "azure_c_pal/timer.h"
+#include "azure_c_pal/gballoc_ll.h"
+#undef ENABLE_MOCKS
+
+#include "azure_c_pal/gballoc_hl.h"
+
+#if defined(_MSC_VER)
```
mattdurak, line 27: already in the win32 folder

win32/tests/gballoc_hl_int/CMakeLists.txt
```diff
@@ -18,15 +18,18 @@ set(${theseTestsName}_test_files
 
 set(${theseTestsName}_c_files
     ../../src/gballoc_ll_${gballoc_ll_impl_lower}.c
-    ../../src/gballoc_hl_passthrough.c
+    ../../src/gballoc_hl_${gballoc_hl_impl_lower}.c
+    ../../src/timer.c #needed because gballoc_hl needs timer to compute "how much time it takes"
+    ../../../common/src/gballoc.c #because timer will allocate memory with gballoc... 
+    ../../src/lock_win32.c #because timer uses locking...
```
mattdurak, line 24: I think that all the int tests should link to the pal lib. Because (eventually) this int test should go in the interfaces folder and run on both Windows and Linux (of course just the subset of passthrough/crt might be the only combination that is supported on Linux).

win32/tests/gballoc_hl_perf/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+#Copyright (c) Microsoft. All rights reserved.
+
+cmake_minimum_required(VERSION 2.8.11)
```
mattdurak, line 3: Also think that perf tests should go in interfaces/tests (under if WIN32 for now until we have it working for Linux)

win32/tests/CMakeLists.txt
```diff
@@ -16,10 +16,55 @@ if(${run_unittests})
...
+        
+            #if(${GBALOC_LL_IMPL} STREQUAL "MIMALLOC")
+            #    if(${ARCHITECTURE} STREQUAL "x86_64")
+            #        build_test_folder(gballoc_hl_int BINARY_DIR ${gballoc_hl_impl_lower}_${gballoc_ll_impl_lower})
+            #    else()
+            #        # do nothing, MIMALLOC only on 64 bit
+            #    endif()
+            #else()
+            #    build_test_folder(gballoc_hl_int BINARY_DIR ${gballoc_hl_impl_lower}_${gballoc_ll_impl_lower})
+            #endif()
```
mattdurak, line 58: so this can be removed?

win32/tests/gballoc_hl_int/CMakeLists.txt
anporumb: Agree! However... easier to say than done. Here's why. Linking against the lib means there 1 (one) library. Well - one library means "one implementation of ll/hl". Which means we'd have to build the azure-c-pal for EVERY combination of hl/ll and link the tests against that. Not impossible, but at this moment I would like to do that with a different task or else this PR never gets checked in.

Notice how the library is a different target than here, and to use it here, it would need to exist predefined before running the test. As is now, it keep the madness of "foreach/foreach" (and frankly there should be "foreach" for every mltiple choices option we have).


Here's a task that tracks linking int tests against the library and not against individual files: Task 7781632: link int tests against the library and not against individual files:


---
In reply to: [459665962](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459665962) [](ancestors = 459665962)

win32/devdoc/gballoc_ll_passthrough.md
anporumb: Spec text is authoritative :)

line 86 shall change

---
In reply to: [459658810](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459658810) [](ancestors = 459658810)

win32/src/gballoc_hl_metrics.c
anporumb: None of the init/deinit are. Added here and everywhere that the function is not thread-safe.

---
In reply to: [459659999](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459659999) [](ancestors = 459659999)

win32/src/gballoc_hl_passthrough.c
anporumb: removed

---
In reply to: [459661312](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459661312) [](ancestors = 459661312)

win32/tests/gballoc_hl_metrics_wout_init_ut/gballoc_hl_metrics_wout_init_ut.c
anporumb: I do remember changing all these SOME TIME IN THE PAST. How did it creep back... 

Not used, so removed completely anyway. Here and everywhere

---
In reply to: [459663283](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459663283) [](ancestors = 459663283)

win32/tests/gballoc_hl_metrics_ut/gballoc_hl_metrics_ut.c
anporumb: removed altogether, doesn't actually apply to the file anymore

---
In reply to: [459663761](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459663761) [](ancestors = 459663761)

win32/tests/gballoc_hl_perf/CMakeLists.txt
anporumb: Moved

---
In reply to: [459666667](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459666667) [](ancestors = 459666667)

anporumb:
> `gballoc` is dvided in 2 layers (`ll` stand for `lower layer`, `hl` stands for `higher layer`):

now it does :)

---
In reply to: [663175956](https://github.com/Azure/azure-c-pal/pull/14#issuecomment-663175956) [](ancestors = 663175956)

---
Refers to: common/devdoc/gballoc_design.md:6 in 39f4b3f. [](commit_id = 39f4b3f10160f2d9e086919ea62ac74d751f9333, deletion_comment = False)

anporumb:
> `gballoc` is dvided in 2 layers (`ll` stand for `lower layer`, `hl` stands for `higher layer`):

divided now

---
In reply to: [663176028](https://github.com/Azure/azure-c-pal/pull/14#issuecomment-663176028) [](ancestors = 663176028)

---
Refers to: common/devdoc/gballoc_design.md:6 in 39f4b3f. [](commit_id = 39f4b3f10160f2d9e086919ea62ac74d751f9333, deletion_comment = False)

CMakeLists.txt
anporumb: Used the exact same description as the one from mimalloc

---
In reply to: [459659420](https://github.com/Azure/azure-c-pal/pull/14#discussion_r459659420) [](ancestors = 459659420)

PR #19, opened 2020-07-28
do not show pal md

anporumb:


CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ add_library(azure_c_pal INTERFACE)
 
 FILE(GLOB azure_c_pal_md_files "doc/*.md")
 SOURCE_GROUP(devdoc FILES ${azure_c_pal_md_files})
-target_sources(azure_c_pal INTERFACE ${azure_c_pal_md_files})
+target_sources(azure_c_pal INTERFACE)
```
mattdurak, line 134: Is this line needed at all then?

anporumb: I do not know the answer to that question, I suspect it is not needed. I'll run some experiments, and if "it still compiles" then i will remove it. brb.

---
In reply to: [461691695](https://github.com/Azure/azure-c-pal/pull/19#discussion_r461691695) [](ancestors = 461691695)

anporumb: Experiments revealed it is not needed, so removed altogether.

---
In reply to: [461704586](https://github.com/Azure/azure-c-pal/pull/19#discussion_r461704586) [](ancestors = 461704586,461691695)

PR #20, opened 2020-07-28
Add TryAcquire for srw_lock

dcristoloveanu:


interfaces/devdoc/srw_lock_requirements.md
```diff
@@ -83,6 +99,23 @@ MOCKABLE_FUNCTION(, void, srw_lock_acquire_shared, SRW_LOCK_HANDLE, handle);
 
 **SRS_SRW_LOCK_02_026: [** If `do_statistics` is `true` and the timer created has recorded more than `TIME_BETWEEN_STATISTICS_LOG` seconds then statistics will be logged and the timer shall be started again. **]**
 
+### srw_lock_try_acquire_shared
+```c
+MOCKABLE_FUNCTION(, int, srw_lock_try_acquire_shared, SRW_LOCK_HANDLE, handle);
+```
+
+`srw_lock_try_acquire_shared` attempts to acquires the SRWLOCK in shared (read) mode.
```
mattdurak, line 107: >acquires [](start = 42, length = 8)

acquire

win32/src/srw_lock_win32.c
```diff
@@ -212,6 +212,59 @@ void srw_lock_acquire_exclusive(SRW_LOCK_HANDLE handle)
...
+
+                /*piggyback on the lock itself to print the statistics "until now" - note - does not include the current Lock call*/
+                if (timer_get_elapsed(handle->timer) >= TIME_BETWEEN_STATISTICS_LOG)
+                {
+                    timer_start(handle->timer);
+                    LogStatistics(handle, "periodic logging almost every " MU_TOSTRING(TIME_BETWEEN_STATISTICS_LOG) " seconds");
+                }
+
+                (void)InterlockedExchange64(&handle->lastCount_AcquireSRWLockExclusive, stop.QuadPart);
+            }
```
mattdurak, line 258: Suggest moving these (the shared and exclusive versions of statistics) to helpers instead of copy-pasting

anporumb:
>     TIMER_HANDLE timer;

"while you are here" (ha ha ha).

REmove this handle, timer doesn't need a handle, timer_global_get_elapsed is as good.

./consnider :)

---
Refers to: win32/src/srw_lock_win32.c:46 in 9f8d038. [](commit_id = 9f8d0388faefedca4b26e99e955e575341a8e18a, deletion_comment = False)

win32/tests/srw_lock_win32_ut/srw_lock_win32_ut.c
```diff
@@ -477,6 +595,114 @@ TEST_FUNCTION(srw_lock_acquire_shared_with_do_statistic_false_succeeds)
...
+{
+    ///arrange
+    SRW_LOCK_HANDLE bsdlLock = TEST_srw_lock_create(true, "test_lock");
+
+    STRICT_EXPECTED_CALL(mocked_TryAcquireSRWLockShared(IGNORED_ARG));
+    STRICT_EXPECTED_CALL(timer_get_elapsed(IGNORED_ARG))
+        .SetReturn(0);
+
+    ///act
+    int result = srw_lock_try_acquire_shared(bsdlLock);
```
anporumb, line 626: >srw_lock_try_acquire_shared [](start = 17, length = 27)

 Ido have to ask why we need these APIs.

Because following this CR there's going to be another where they are used. So... a bit of look in the future, if you will. Just out of curiosity

win32/src/srw_lock_win32.c
```diff
@@ -285,6 +338,64 @@ void srw_lock_acquire_shared(SRW_LOCK_HANDLE handle)
...
+        if (handle->doStatistics)
+        {
+            (void)QueryPerformanceCounter(&start);
+        }
+
+        /*Codes_SRS_SRW_LOCK_01_002: [ Otherwise srw_lock_try_acquire_shared shall call TryAcquireSRWLockShared. ]*/
+        if (!TryAcquireSRWLockShared(&handle->lock))
+        {
+            /* Codes_SRS_SRW_LOCK_01_003: [ If TryAcquireSRWLockShared returns FALSE, srw_lock_try_acquire_shared shall return a non-zero value. ]*/
+            result = MU_FAILURE;
```
amarsa-msft, line 364: Looks like you need an ENUM here to differentiate between an actual error and the failure to acquire a lock. I suspect the user code will react differently in both the instances:
If the handle is NULL, the user code might error out
If the lock could not be acquired, the user is free to decide what she wants to do.

amarsa-msft: Same comment for the other function.

---
In reply to: [461973925](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461973925) [](ancestors = 461973925)

mattdurak: I think that acquiring a lock should never fail in any other way. I really don't like the old lock.h that had a return code just for taking the lock.

---
In reply to: [461974020](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461974020) [](ancestors = 461974020,461973925)

interfaces/devdoc/srw_lock_requirements.md
dcristoloveanu: Fixed.

---
In reply to: [461957008](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461957008) [](ancestors = 461957008)

win32/src/srw_lock_win32.c
amarsa-msft: This isn't acquiring a lock, it's trying. The underlying API itself has two return codes. I think it makes sense to be explicit about what the failure is when we don't acquire a lock.

---
In reply to: [461974555](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461974555) [](ancestors = 461974555,461974020,461973925)

win32/tests/srw_lock_win32_ut/srw_lock_win32_ut.c
dcristoloveanu: Ah, sure, sorry about that:
Currently when we do a change role we unwind all the operations that have failed (because they will fail due to NOT_PRIMARY).
We have set block epoch which literally takes a lock on the block to disallows new appends (appends acquire as shared, set block epoch acquires as exclusive).

the lock on the block is acquired exclusively when set block epoch starts and released exclusively when it completes.
So if the set block epoch starts, lock is acquired exclusively, append attempts to acquire shared and it blocks in there.
Then we have a NOT_PRIMARY and we want to unwind the operations, but unfortunately we are stuck in the acquire shared lock.

So the way I am thinking of fixing this is attempt to take the lock (if it fails bounce this up as busy). We do have a retry layer that will retry the operation if it fails so I am planning to simply reuse the retry in the bsi_user_retry layer.

---
In reply to: [461968384](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461968384) [](ancestors = 461968384)

win32/src/srw_lock_win32.c
dcristoloveanu: Ugh. I see where Amar is coming from: if non-zero is returned, does that mean that we did not acquire the lock - should we retry? or is it a critical failure?
Interestingly TryAcquireSRWLockExclusive also returns a BOOLEAN, so there is no difference between passing a NULL and not being able to acquire the lock.

---
In reply to: [461979716](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461979716) [](ancestors = 461979716,461974555,461974020,461973925)

amarsa-msft: hmm, that's true. I hope they do a null check :P

---
In reply to: [461982771](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461982771) [](ancestors = 461982771,461979716,461974555,461974020,461973925)

win32/src/srw_lock_win32.c
dcristoloveanu: I like this suggestion quite a lot, applied.

---
In reply to: [461957918](https://github.com/Azure/azure-c-pal/pull/20#discussion_r461957918) [](ancestors = 461957918)

dcristoloveanu:
>     TIMER_HANDLE timer;

I considered it, started it, then got into more changes because seconds vs ms and because would need to remember a start time rather than use start and it just felt I don't want to do it now.

---
In reply to: [665360736](https://github.com/Azure/azure-c-pal/pull/20#issuecomment-665360736) [](ancestors = 665360736)

---
Refers to: win32/src/srw_lock_win32.c:46 in 9f8d038. [](commit_id = 9f8d0388faefedca4b26e99e955e575341a8e18a, deletion_comment = False)

PR #31, opened 2020-08-07
gballoc hl metrics uses lazy init

anporumb:


win32/src/gballoc_hl_metrics.c
```diff
@@ -125,54 +126,60 @@ static void internal_copy_latency_data(GBALLOC_LATENCY_BUCKETS* latency_buckets_
     }
 }
 
-static int wasInitialized = 0;
+static call_once_t g_lazy = 0;
```
mattdurak, line 129: >0 [](start = 28, length = 1)

LAZY_INIT_NOT_DONE?

win32/src/gballoc_hl_metrics.c
```diff
@@ -125,54 +126,60 @@ static void internal_copy_latency_data(GBALLOC_LATENCY_BUCKETS* latency_buckets_
...
+    if (interlocked_add(&g_lazy, 0) != LAZY_INIT_NOT_DONE)
     {
-        /* Codes_SRS_GBALLOC_HL_METRICS_01_005: [ If gballoc_hl_deinit is called while not initialized, gballoc_hl_deinit shall return. ]*/
-        LogError("Not initialized");
-    }
-    else
-    {
-        /* Codes_SRS_GBALLOC_HL_METRICS_01_006: [ Otherwise it shall call gballoc_ll_deinit to deinitialize the ll layer. ]*/
-        // error ignored as we can't do much with it
+        interlocked_exchange(&g_lazy, LAZY_INIT_NOT_DONE);
```
mattdurak, line 177: I'm wondering if this should be interlocked_compare_exchange (except we shouldn't really know that the compare value is "2"...).

Is it okay if two callers could deinit and it would call gballoc_ll_deinit twice?

Should there be a lazy_init_reset helper that does the interlocked compare exchange to do a deinit like this exactly once?

anporumb: A very good question otherwise. Problem is with "threading" and the model we are considering.

Assume we use interlocked_compare_exchange. So code looks like: 
```c
if(interlocked_compare_exchange(&lazy, NOT_DONE, 2) {ll_deinit}
```

There's a change, after the interlocked is executed, AND state switched to NOT_DONE that another gballoc_malloc comes. Let's say that runs its course and ends up calling gballoc_ll => so the lower layer is initialized. Well then we return in this thread and call ll_deinit. Not cool.

So then it follows that we should not take other threaded calls in de_init. Which is fine. I think this is the model we should have. None of our HANDLEs are resilient to that, and we don't build them like that. C++ objects the same... all bets are off when destructor is called and there's threading partying on the object methods.

So to answer: it is not ok for 2 of the callers to call deinit at the same time. Just like _destroy/ ~class is not supposed to be called twice.

==

If we count init/deinit a la WSAStartup calls - we are still not good, becuase we are using lazy_init. And how should our own calls be counted? ...

==

Possibly the solution should look like SM in c utils. Or just use real handles (this is also something I'll have to do ... later).

---
In reply to: [467115165](https://github.com/Azure/azure-c-pal/pull/31#discussion_r467115165) [](ancestors = 467115165)

win32/src/gballoc_hl_metrics.c
anporumb: Of course :)

---
In reply to: [467113867](https://github.com/Azure/azure-c-pal/pull/31#discussion_r467113867) [](ancestors = 467113867)

PR #33, opened 2020-08-15
ref count type define supporting malloc/free being specified

dcristoloveanu:
ref count type define supporting malloc/free being specified

### DIFF
--- a/common/devdoc/refcount_requirements.md
+++ b/common/devdoc/refcount_requirements.md
@@ -17,7 +17,10 @@ It wraps the structure that needs to be ref counted into another structure that 
 #define DEC_REF(type, var) interlocked_decrement(&((REFCOUNT_TYPE(type)*)((unsigned char*)var - offsetof(REFCOUNT_TYPE(type), counted)))->count)
 #define INIT_REF(type, var) interlocked_exchange(&((REFCOUNT_TYPE(type)*)((unsigned char*)var - offsetof(REFCOUNT_TYPE(type), counted)))->count, 1)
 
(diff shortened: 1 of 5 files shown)


common/devdoc/refcount_requirements.md
```diff
@@ -13,14 +13,25 @@ It wraps the structure that needs to be ref counted into another structure that
...
+#define INIT_REF(type, var) interlocked_exchange(&((REFCOUNT_TYPE(type)*)((unsigned char*)var - offsetof(REFCOUNT_TYPE(type), counted)))->count, 1)
+
+#define DEFINE_REFCOUNT_TYPE(type, malloc_func, free_func) \
 ...
 ```
 
 ### DEFINE_REFCOUNT_TYPE
 
+```c
+#define DEFINE_REFCOUNT_TYPE(type, malloc_func, free_func) \
```
amarsa-msft, line 27: >DEFINE_REFCOUNT_TYPE [](start = 8, length = 20)

Will there be defaults for the malloc and free functions used? If so, Should we have a different macro `DEFINE_REFCOUNT_TYPE_CUSTOM_ALLOC` so that  this does not ripple through the repos?

dcristoloveanu: A very good idea. I do not remember why I discarded it initially. It sounds much better when you suggested it :-).

---
In reply to: [470943363](https://github.com/Azure/azure-c-pal/pull/33#discussion_r470943363) [](ancestors = 470943363)

PR #38, opened 2020-09-30
prepare for zc preprocessor

anporumb:


build/linux/build_linux.sh
```diff
@@ -15,7 +15,7 @@ CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.ncpu)
 rm -r -f $build_folder
 mkdir -p $build_folder
 pushd $build_folder
-cmake -Drun_valgrind:BOOL=ON $build_root -Drun_unittests:BOOL=ON -Drun_int_tests:BOOL=ON
+cmake -Drun_valgrind:BOOL=OFF $build_root -Drun_unittests:BOOL=ON -Drun_int_tests:BOOL=ON
```
dcristoloveanu, line 18: >OFF [](start = 26, length = 3)

I think we should keep valgrind. Just ssh into the machine and sudo apt-get install it IMO. Won't take long and we'll need it anyways

interfaces/tests/sync_int/sync_int.c
```diff
@@ -38,7 +38,7 @@ static int increment_on_odd_values(void* address)
         (void)interlocked_increment(ptr);
         wake_by_address_all(ptr);
     }
-    ThreadAPI_Exit(0);
```
dcristoloveanu, line 41: >ThreadAPI_Exit(0); [](start = 4, length = 18)

This is actually needed IIRC for pthreads, So it should stay?

dcristoloveanu (approved):
:shipit:

build/linux/build_linux.sh
anporumb: Made a task for this: [https://msazure.visualstudio.com/One/_workitems/edit/8444467](https://msazure.visualstudio.com/One/_workitems/edit/8444467) 

Task 8444467: enable valgrind in azure c pal

I don't think it ever worked. when run_valgrind is set to ON, the valgrind tests will be generated IFF the machine has valgrind installed.

I suspect that the hosted machines DO NOT have valgrind installed, therefore, even you enable it - it doesn't do anything. I can confirm my suispicion here: [https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md). Valgrind is not there.

And the reason why I disabled it is because OBVIOUSLY we do not pass it.

---
In reply to: [497239807](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497239807) [](ancestors = 497239807)

interfaces/tests/sync_int/sync_int.c
anporumb: Not according to the documentation.

[https://linux.die.net/man/3/pthread_exit](https://linux.die.net/man/3/pthread_exit) say: (also man 3 pthread_exit says the same thing):

```
Performing a **return** from the start function of any thread other than the main thread results in an implicit call to pthread_exit(), using the function's return value as the thread's exit status.
```

Here's another problem with "not returning" and actually exiting through ExitThread/pthread_exit. 

We've taken those functions and nicely thrown them in ThreadAPI_Exit. Well - that is confusing to the compiler (I've send you an email about this:) in Windows Because ExitThread is marked as "DECLSPEC_NORETURN" - (true that). So ... ThreadAPI_Exit is not marked as noreturn. So theeeeen... soooooome versions of visual studio will propagate the noreturn, and then real "return" cannot follow ThreadAPI_Exit. And soooooom ooooother versions of Visual Studio will not propagate NORETURN so then the compiler will complain about a function without a return value.

Having "return 0" solves both of these cases on Windows, because there's no ExitThread called.

Also return 0 is perfectly fine in Linux as far as I can tell. 0 being the same as NULL cause obviously LINUX needs to return a void* from its threads. How do we even... put a void* <--->DWORD, cause sizeof void* might be 4,8... depends... 

Anyway - I will make an item in **that** bucket about this discovery.

---
In reply to: [497240061](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497240061) [](ancestors = 497240061)

build/linux/build_linux.sh
dcristoloveanu: I understand. Sure, separate task is fine.
valgrind was being installed - it was installed through the bash step you removed:

- bash: |
      sudo apt-get update
      sudo apt-get install build-essential pkg-config git cmake uuid-dev valgrind
    displayName: 'Bash Script'

But maybe something else is missing for this repo.
For example ZRPC does use valgrind:

[https://msazure.visualstudio.com/One/_build/results?buildId=35000926&view=logs&j=ae9a2f62-3ee1-5984-579f-87b38be696e5&t=1b4a4e70-15a3-566b-e768-285a68c542e2&l=21](https://msazure.visualstudio.com/One/_build/results?buildId=35000926&view=logs&j=ae9a2f62-3ee1-5984-579f-87b38be696e5&t=1b4a4e70-15a3-566b-e768-285a68c542e2&l=21)

My comment was based on thinking that it does work and it seemed odd that we'd remove it. So separate task.

---
In reply to: [497241016](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497241016) [](ancestors = 497241016,497239807)

interfaces/tests/sync_int/sync_int.c
dcristoloveanu: Ok, makes sense.
I know it is confusing to the compiler because I hit it first :-). When I enabled LTCG we started having that problem. At the time I disabled the warning, but since return does an implicit pthread_exit of course we want to remove Thread_Exit.

Cheers!

---
In reply to: [497247378](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497247378) [](ancestors = 497247378,497240061)

dcristoloveanu: As a matter of fact we should remove ThreadAPI_Exit completely. It only "confuzles" to quote my daughter.

---
In reply to: [497251256](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497251256) [](ancestors = 497251256,497247378,497240061)

dcristoloveanu: Task created to remove it: [https://msazure.visualstudio.com/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=8444719](https://msazure.visualstudio.com/One/_backlogs/backlog/Azure%20Messaging%20Store/Backlog%20items/?workitem=8444719) 

---
In reply to: [497251649](https://github.com/Azure/azure-c-pal/pull/38#discussion_r497251649) [](ancestors = 497251649,497251256,497247378,497240061)

PR #43, opened 2020-10-07
Remove interlocked_hl (will be reborn in azure-c-util)

dcristoloveanu:
Remove interlocked_hl (will be reborn in azure-c-util)


mattdurak (approved):
:shipit:

PR #44, opened 2020-10-13
Avoid collision when building on Linux: timer_create already exists ...

dcristoloveanu:
Avoid collision when building on Linux: timer_create already exists ...


mattdurak (approved):
:shipit:

PR #49, opened 2020-10-29
Fix sync_int tests for Linux

dcristoloveanu:
Fix sync_int tests for Linux

### DIFF
--- a/linux/src/sync_linux.c
+++ b/linux/src/sync_linux.c
@@ -1,15 +1,23 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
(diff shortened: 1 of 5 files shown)


linux/src/sync_linux.c
```diff
@@ -1,15 +1,23 @@
...
+#include <stdbool.h>
 #include <unistd.h>
 #include <limits.h>
 #include <time.h>
+#include <errno.h>
+
+#include "sys/syscall.h"
+#include "linux/futex.h"
+
+#include "azure_c_logging/xlogging.h"
```
mattdurak, line 13: Not used (intended to add logging around line 46?)

mattdurak (approved):
:shipit:

PR #54, opened 2020-11-03
Add sysinfo abstraction (primarily for getting processor count)

dcristoloveanu:
Add sysinfo abstraction (primarily for getting processor count)
- On Linux it uses sysconf
- On Windows it uses GetSystemInfo (for now ...)


mattdurak (approved):
:shipit:

PR #58, opened 2020-11-18
Handle 0 bytes received as socket closed

mattdurak:
On receive of 0 bytes, treat it as socket disconnect


dcristoloveanu (approved):
:shipit:

PR #65, opened 2020-12-16
Adding cancel and restart to threadpool timer

mattdurak:
Make threadpool timers more re-usable by adding a cancel (which does part of stop without freeing resources) and restart (which changes the delay and period of a created timer) to them.

### DIFF
--- a/interfaces/devdoc/threadpool_requirements.md
+++ b/interfaces/devdoc/threadpool_requirements.md
@@ -9,7 +9,11 @@ A `threadpool` object receives an execution engine at creation time in order to 
 
 The `threadpool` interface supports:
  - Scheduling a single work item (`threadpool_schedule_work`)
- - Scheduling a timer function to execute at a regular interval (`threadpool_start_timer`/`threadpool_stop_timer`)
+ - Scheduling a timer function to execute at a regular interval
+   - `threadpool_start_timer`
+   - `threadpool_stop_timer`
+   - `threadpool_restart_timer`
+   - `threadpool_cancel_timer`
 
 The lifetime of the execution engine should supersede the lifetime of the `threadpool` object.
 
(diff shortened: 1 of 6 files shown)


interfaces/devdoc/threadpool_requirements.md
```diff
@@ -9,7 +9,11 @@ A `threadpool` object receives an execution engine at creation time in order to
 
 The `threadpool` interface supports:
  - Scheduling a single work item (`threadpool_schedule_work`)
- - Scheduling a timer function to execute at a regular interval (`threadpool_start_timer`/`threadpool_stop_timer`)
+ - Scheduling a timer function to execute at a regular interval
+   - `threadpool_start_timer`
+   - `threadpool_stop_timer`
+   - `threadpool_restart_timer`
+   - `threadpool_cancel_timer`
```
dcristoloveanu, line 16: >threadpool_cancel_timer [](start = 6, length = 23)

How is this different than stop?

dcristoloveanu: Oh I see this does not free the resources. Do we need this now? It looked a bit confusing having cancel and stop. Took me a second to regroup to figure out which one does what.
And I think that confusion will happen every time we use it :-).

---
In reply to: [544611101](https://github.com/Azure/c-pal/pull/65#discussion_r544611101) [](ancestors = 544611101)

dcristoloveanu: Thus I'd say if we don't need it now maybe let's not add it.

---
In reply to: [544611918](https://github.com/Azure/c-pal/pull/65#discussion_r544611918) [](ancestors = 544611918,544611101)

dcristoloveanu (approved):
:shipit:

interfaces/devdoc/threadpool_requirements.md
mattdurak: I do need it now (stop the timer without freeing resources, which I want for the watchdog changes)

---
In reply to: [544612115](https://github.com/Azure/c-pal/pull/65#discussion_r544612115) [](ancestors = 544612115,544611918,544611101)

mattdurak: Renaming it to threadpool_timer_destroy though to clear up confusion

---
In reply to: [544649895](https://github.com/Azure/c-pal/pull/65#discussion_r544649895) [](ancestors = 544649895,544612115,544611918,544611101)

PR #78, opened 2021-04-06
New deps

dcristoloveanu:


jebrando (approved):
:shipit:

common/tests/call_once_int/main.c
```diff
@@ -7,6 +7,6 @@
 int main(void)
 {
     size_t failedTestCount = 0;
-    RUN_TEST_SUITE(call_once_inttests, failedTestCount);
+    RUN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE, failedTestCount);
```
mattdurak, line 10: > RUN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE, failedTestCount); [](start = 3, length = 61)

interesting...I guess now (besides those tests with the leak check retries) this file is identical boilerplate everywhere and maybe it could just be referenced from some shared place....

jebrando: I was just talking to Dan about this same thing.

---
In reply to: [608209577](https://github.com/Azure/c-pal/pull/78#discussion_r608209577) [](ancestors = 608209577)

mattdurak (approved):
:shipit:

common/tests/call_once_int/main.c
dcristoloveanu: I actually had an initial c test runner switcher with a main_ctest.c. But the "with leaks" and the perf tests have custom main.
So I could do another flavor of the build test functions with custom main.
Aaaaand then I said I will do it as a separate PR, but now I am inclined to do it now.

---
In reply to: [608209775](https://github.com/Azure/c-pal/pull/78#discussion_r608209775) [](ancestors = 608209775,608209577)

PR #100, opened 2021-09-27
Helper for refcount reals

mattdurak:
Adds a header for using refcount types in tests with mocks without generating interlocked mock calls.
Unit tests should include real_refcount.h before `DEFINE_REFCOUNT_TYPE_WITH_CUSTOM_ALLOC`

### DIFF
--- a/common/reals/real_refcount.h
+++ b/common/reals/real_refcount.h
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+#ifndef REAL_REFCOUNT_H
+#define REAL_REFCOUNT_H
+
+#include "c_pal/refcount.h"
+
+#include "real_interlocked.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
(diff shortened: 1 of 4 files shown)


common/reals/real_refcount.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "real_interlocked.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// When this file is included, test types defined as refcount types will not generate mock calls to interlocked
+
+#undef INC_REF
```
dcristoloveanu, line 16: It looks like including refcount and then real_refcount would get the real_ macros.
A subsequent include of refcount would yield an error, right?

Is there a ause case where we want to force the real_ defines by undefing and the redefining them? If there is such a use case we can discuss. Otherwise I would say this should be an error if already defined.

dcristoloveanu: Marking as request change for now.

dcristoloveanu (changes requested):

dcristoloveanu (approved):
:shipit:

PR #105, opened 2021-10-08
fix test lib names

parth21999:
Appending `_int` to gballoc_int tests so that they get filtered out for real check.


jebrando (approved):
:shipit:

PR #113, opened 2021-10-15
use new gballoc functions

anporumb:


win32/src/async_socket_win32.c
```diff
@@ -645,153 +646,144 @@ int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BU
...
-                {
-                    total_buffer_bytes += payload[i].length;
-                }
-            }
-
-            if (i < buffer_count)
-            {
-                LogError("Invalid buffers passed to async_socket_receive_async");
+                /* Codes_SRS_ASYNC_SOCKET_WIN32_01_098: [ If async_socket is not OPEN, async_socket_receive_async shall fail and return a non-zero value. ]*/
+                LogWarning("Not open");
```
jebrando, line 689: >"Not open" [](http://example.com/codeflow?start=27&length=10)

It would be really nice to add the current state in here so it can give a clue as to where we are when we failed.

anporumb: Change here and everywhere where "not open" could be found.

win32/src/string_utils.c
```diff
@@ -31,7 +31,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, char*, vsprintf_char, const char*, format, va_list
     }
     else
     {
-        result = (char*)malloc((neededSize + 1) * sizeof(char));
+        result = (char*)malloc((neededSize + 1U) * sizeof(char));
         if (result == NULL)
         {
             LogError("failure in malloc");
```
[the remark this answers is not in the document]
anporumb: I missed it,added  now

PR #140, opened 2022-01-11
fix linux string utils

anporumb:


linux/inc/c_pal/string_utils.h
```diff
@@ -1,62 +1,64 @@
...
+#include "macro_utils/macro_utils.h"
+
+#include "umock_c/umock_c_prod.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*produces a string as if printed by printf*/
+extern char* sprintf_char_function(const char* format, ...);
```
dcristoloveanu, line 31: >extern [](http://example.com/codeflow?start=0&length=6)

Likely not needed

anporumb: gzzzzzzzz

dcristoloveanu (approved):
:shipit:

linux/inc/c_pal/string_utils.h
anporumb: removed from here and everywhere

PR #147, opened 2022-04-08
Async socket closesocket

mattdurak:
Change ownership of SOCKET in async_socket so that async_socket calls closesocket


dcristoloveanu (approved):
:shipit:

PR #150, opened 2022-06-02
Use GetActiveProcessorCount in sysinfo

parth21999:


win32/devdoc/sysinfo_win32_requirements.md
```diff
@@ -19,6 +19,6 @@ MOCKABLE_FUNCTION(, uint32_t, sysinfo_get_processor_count);
 
 `sysinfo_get_processor_count` returns the processor count as returned by the operating system.
 
-**SRS_SYSINFO_WIN32_01_001: [** `sysinfo_get_processor_count` shall call `GetSystemInfo` to obtain the system information. **]**
+**SRS_SYSINFO_WIN32_43_001: [** `sysinfo_get_processor_count` shall call `GetActiveProcessorCount(ALL_PROCESSOR_GROUPS)` to obtain the number of processors. **]**
 
-**SRS_SYSINFO_WIN32_01_002: [** `sysinfo_get_processor_count` shall return the processor count as returned by `GetSystemInfo`. **]**
+**SRS_SYSINFO_WIN32_43_002: [** `sysinfo_get_processor_count` shall return the processor count as returned by `GetActiveProcessorCount`. **]**
```
dcristoloveanu, line 24: >sysinfo_get_processor_count` shall return the p [](http://example.com/codeflow?start=33&length=47)

Interestingly nothing is mentioned about failure cases.

GetActiveProcessorCount can fail, so we should at least spec what we return when it does fail:

"If the function fails, the return value is zero. To get extended error information, use GetLastError."

win32/tests/sysinfo_win32_ut/sysinfo_win32_ut.c
```diff
@@ -46,7 +46,9 @@ TEST_SUITE_INITIALIZE(suite_init)
     ASSERT_ARE_EQUAL(int, 0, umock_c_init(on_umock_c_error), "umock_c_init failed");
     ASSERT_ARE_EQUAL(int, 0, umocktypes_stdint_register_types(), "umocktypes_stdint_register_types failed");
 
-    REGISTER_UMOCK_ALIAS_TYPE(LPSYSTEM_INFO, void*);
+    REGISTER_UMOCK_ALIAS_TYPE(WORD, uint16_t);
+    REGISTER_UMOCK_ALIAS_TYPE(DWORD, uint32_t);
```
dcristoloveanu, line 50: I think there is a helper that registers windows types, you should use that one.

dcristoloveanu (approved):
:shipit:

PR #175, opened 2022-11-10
Add execution_engine_inc_ref to threadpool

mattdurak:
Previously, this code would get stuck:

```c
EXECUTION_ENGINE_HANDLE execution_engine = execution_engine_create(NULL);
THREADPOOL_HANDLE threadpool = threadpool_create(execution_engine);
execution_engine_dec_ref(execution_engine); // this is the last reference even though threadpool needs it!
(void)threadpool_open_async(threadpool, on_open_complete, &open_event);
// wait for open
threadpool_schedule_work(threadpool, work_function, NULL); // gets stuck!
```

Really, the threadpool needs to hold a reference to the execution engine and not rely on the caller to do that.

### DIFF
--- a/win32/tests/threadpool_win32_int/threadpool_win32_int.c
+++ b/win32/tests/threadpool_win32_int/threadpool_win32_int.c
@@ -203,6 +203,47 @@ TEST_FUNCTION(one_work_item_schedule_works)
     execution_engine_dec_ref(execution_engine);
 }
 
+TEST_FUNCTION(threadpool_owns_execution_engine_reference_and_can_schedule_work)
+{
+    // arrange
+    // create an execution engine
+    volatile LONG call_count;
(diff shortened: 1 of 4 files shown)


jebrando (approved):
:shipit:

darobs (approved):
:shipit:

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -203,6 +203,47 @@ TEST_FUNCTION(one_work_item_schedule_works)
     execution_engine_dec_ref(execution_engine);
 }
 
+TEST_FUNCTION(threadpool_owns_execution_engine_reference_and_can_schedule_work)
```
anporumb, line 206: >threadpool_owns_execution_engine_reference_and_can_schedule_work [](http://example.com/codeflow?start=14&length=64)

nice test

anporumb (approved):
:shipit:

PR #177, opened 2022-11-14
feature: adding async sockets for linux

jebrando:


linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+    ASYNC_SOCKET_IO_TYPE_RECEIVE
+
+MU_DEFINE_ENUM(ASYNC_SOCKET_IO_TYPE, ASYNC_SOCKET_IO_TYPE_VALUES)
+MU_DEFINE_ENUM_STRINGS(ASYNC_SOCKET_IO_TYPE, ASYNC_SOCKET_IO_TYPE_VALUES)
+
+MU_DEFINE_ENUM_STRINGS(ASYNC_SOCKET_SEND_SYNC_RESULT, ASYNC_SOCKET_SEND_SYNC_RESULT_VALUES)
+
+// These will be in a separate file maybe along with the 
+// thread function?
+volatile_atomic int32_t g_threads_num;
```
dcristoloveanu, line 50: >g_threads_num [](http://example.com/codeflow?start=24&length=13)

static

dcristoloveanu: Here and below

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+#include "c_pal/gballoc_hl_redirect.h"
+#include "c_pal/execution_engine.h"
+#include "c_pal/interlocked.h"
+#include "c_pal/sync.h"
+#include "c_pal/socket_handle.h"
+#include "c_pal/threadapi.h"
+
+#include "c_pal/async_socket.h"
+
+#define MAX_EVENTS_NUM      64
```
dcristoloveanu, line 27: erm, does this mean we can only have 64 sockets?
That will not really work, will it?

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+#define EVENTS_TIMEOUT      2*1000
```
dcristoloveanu, line 28: >2*1000 [](http://example.com/codeflow?start=28&length=6)

I'd state /* ms */ next to it if it is milliseconds

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+    (void)parameter;
+
+    // Loop while true
+    do
+    {
+        struct epoll_event events[MAX_EVENTS_NUM];
+        int num_ready = epoll_wait(g_epoll, events, MAX_EVENTS_NUM, EVENTS_TIMEOUT);
+        for(int index = 0; index < num_ready; index++)
+        {
+            if (events[index].events & EPOLLRDHUP)
```
dcristoloveanu, line 114: >if (events[index].events & EPOLLRDHUP) [](http://example.com/codeflow?start=12&length=38)

would be nice to have some comments of what this masks.
Here and below.

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+                ASYNC_SOCKET_RECV_CONTEXT* recv_context = events[index].data.ptr;
+
+                if (recv_context != NULL)
+                {
+                    uint32_t index = 0;
+                    ssize_t recv_size;
+                    do
+                    {
+
+                        recv_size = recv(recv_context->socket_handle, recv_context->recv_buffers[index].buffer, recv_context->recv_buffers[index].length, 0);
```
dcristoloveanu, line 135: Nit:  extra line

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+                        if (recv_size < 0)
+                        {
+                            if (errno == EAGAIN || errno == EWOULDBLOCK)
+                            {
+                                // No more data to recv
+                                receive_result = ASYNC_SOCKET_RECEIVE_OK;
```
dcristoloveanu, line 141: Doesn't this actually say that there will be more data?

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+                                break;
+                            }
+                            else
+                            {
+                                if (errno == ECONNRESET)
+                                {
+                                    receive_result = ASYNC_SOCKET_RECEIVE_ABANDONED;
```
dcristoloveanu, line 148: > receive_result = ASYNC_SOCKET_RECEIVE_ABANDONED; [](http://example.com/codeflow?start=35&length=49)

We should log error here

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+                        {
+                            if (index + 1 >= recv_context->total_buffer_count || recv_size < recv_context->recv_buffers[index].length)
+                            {
+                                // We gotten all the data we need to get or we don't have any more recv space
+                                receive_result = ASYNC_SOCKET_RECEIVE_OK;
+                                break;
+                            }
+                            else
+                            {
+                                index++;
```
dcristoloveanu, line 174: This is interesting.
I think we should actually indicate the data we got, right?
We're sort of trying to fill more buffers and now the data we got is lingering in our stack instead of being processed.

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+                    do
+                    {
+                        void* send_pos = send_context->buffers[index].buffer + send_size;
+                        uint32_t send_length = send_context->buffers[index].length - send_size;
+                        send_size += send(send_context->socket_handle, send_pos, send_length, 0);
+                        if (send_size < 0 && (errno != EAGAIN && errno != EWOULDBLOCK) )
+                        {
+                            if (errno == ECONNRESET)
+                            {
+                                send_result = ASYNC_SOCKET_SEND_ABANDONED;
```
dcristoloveanu, line 204: >SOCKET_SEND_A [](http://example.com/codeflow?start=52&length=13)

We should log error here too

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+    {
+        g_epoll = epoll_create(MAX_EVENTS_NUM);
+        if (g_epoll == -1)
+        {
+            LogError("failure epoll_create error num: %d", errno);
+        }
+        else
+        {
+            result = g_epoll;
+            for (int32_t index = 0; index < THREAD_COUNT; index++)
```
dcristoloveanu, line 240: >index < THREAD_COUNT [](http://example.com/codeflow?start=36&length=20)

Ah, I see, you are creating multiple threads with 64 events each.

dcristoloveanu: Or not.
THREAD_COUNT is 1. Soooo this won;t work for more than 64 sockets.

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,695 @@
...
+static void internal_close(ASYNC_SOCKET_HANDLE async_socket)
+{
+    int32_t value;
+    while ((value = interlocked_add(&async_socket->pending_api_calls, 0)) != 0)
+    {
+        (void)wait_on_address(&async_socket->pending_api_calls, value, UINT32_MAX);
+    }
+
+    // Remove the socket from the epoll
+    epoll_ctl(async_socket->epoll, EPOLL_CTL_DEL, async_socket->socket_handle, NULL);
```
dcristoloveanu, line 286: >epoll_ctl [](http://example.com/codeflow?start=4&length=9)

Can this fail? We should log if it does

linux/tests/async_socket_linux_int/async_socket_linux_int.c
```diff
@@ -0,0 +1,412 @@
...
+
+    async_socket_close(server_async_socket);
+    async_socket_close(client_async_socket);
+    close(listen_socket);
+    async_socket_destroy(server_async_socket);
+    async_socket_destroy(client_async_socket);
+    execution_engine_dec_ref(execution_engine);
+}
+
+END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
dcristoloveanu, line 412: >TE_NAME_ [](http://example.com/codeflow?start=23&length=8)

We have to add a test with 1000 sockets.

dcristoloveanu:
We need to have more than 64 sockets, waiting for that mainly.

dcristoloveanu (changes requested):
:clock1:

linux/src/async_socket_linux.c
jebrando: No this means that the on older system (This has been fixed with new kernel, the doc didn't say which kernel version has been fixed) this value is the number of spots in the array it starts with.  It grows as needed.

linux/src/async_socket_linux.c
jebrando: This is interesting because the async_socket_receive_async takes in an array of ASYNC_SOCKET_BUFFER.  So you can only receive half the size at a time.  I'm guessing we could call the callback after each recv, but then how to tell the callback to look at the 2nd array position?

linux/src/async_socket_linux.c
jebrando: The MAX_EVENTS_NUM is ignored since kernel 2.6.8, but on on earlier versions this is the note: 

The kernel used this information as a hint for the amount of space to initially allocate in internal data structures describing events

So technically I could just put 1 and it would figure it out.

[https://man7.org/linux/man-pages/man2/epoll_create.2.html](https://man7.org/linux/man-pages/man2/epoll_create.2.html)

linux/src/async_socket_linux.c
jebrando: The documentation says that you'll get this if "The socket's file descriptor is marked O_NONBLOCK and no data is waiting to be received", so no data is available, so I think we should return on this ... or should we loop again assuming data is coming on the next iteration?  I'll need to thin about this.

jebrando:
See other comment:

The MAX_EVENTS_NUM is ignored since kernel 2.6.8, but on on earlier versions this is the note: 

The kernel used this information as a hint for the amount of space to initially allocate in internal data structures describing events

So technically I could just put 1 and it would figure it out.

[https://man7.org/linux/man-pages/man2/epoll_create.2.html](https://man7.org/linux/man-pages/man2/epoll_create.2.html)

---
In reply to: [1316452859](https://github.com/Azure/c-pal/pull/177#issuecomment-1316452859) [](http://example.com/codeflow?ancestors=1316452859,1316452859)

linux/tests/async_socket_linux_int/async_socket_linux_int.c
```diff
@@ -0,0 +1,520 @@
...
+    *listen_socket = socket(AF_INET, SOCK_STREAM, 0);
+    ASSERT_ARE_NOT_EQUAL(int, INVALID_SOCKET, *listen_socket);
+
+    const int enable = 1;
+    ASSERT_ARE_EQUAL(int, 0, setsockopt(*listen_socket, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(int)));
+
+    int bind_res;
+    uint32_t counter = 0;
+    // The gate machines fails on bind due to 3 different
+    // Process running, so we have to figure out the bind
```
dcristoloveanu, line 104: >figure out the bind [](http://example.com/codeflow?start=38&length=19)

I think we should state that we're running through a set of ports and picking the first one that succeeds

linux/tests/async_socket_linux_int/async_socket_linux_int.c
```diff
@@ -0,0 +1,520 @@
...
+    } while (bind_res != 0 && counter < 10);
+    ASSERT_ARE_EQUAL(int, 0, bind_res, "Failure attempting to bind (%d) to socket %d error (%d): %s", bind_res, g_port_num, errno, strerror(errno));
+
+    // set it to async IO
+    set_nonblocking(*listen_socket);
+
+    // start listening
+    ASSERT_ARE_EQUAL(int, 0, listen(*listen_socket, SOMAXCONN), "Failure on listen socket");
+
+}
```
dcristoloveanu, line 128: Nit: extra line

linux/tests/async_socket_linux_int/async_socket_linux_int.c
```diff
@@ -0,0 +1,520 @@
...
+
+static void setup_client_sockets(int port_num, SOCKET_HANDLE* client_socket, SOCKET_HANDLE* listen_socket, SOCKET_HANDLE* accept_socket)
```
dcristoloveanu, line 130: >setup_client_sockets [](http://example.com/codeflow?start=12&length=20)

Sounds like it is setting up just one client socket.
But also seems like it is setting up both client and server sockets.

So maybe it should be named setup_seockets or setup_test_sockets

linux/src/async_socket_linux.c
```diff
@@ -0,0 +1,707 @@
...
+    if (async_socket == NULL)
+    {
+        LogError("Invalid arguments: ASYNC_SOCKET_HANDLE async_socket=%p", async_socket);
+    }
+    else
+    {
+        do
+        {
+            int32_t current_state = interlocked_compare_exchange(&async_socket->state, (int32_t)ASYNC_SOCKET_LINUX_STATE_CLOSING, (int32_t)ASYNC_SOCKET_LINUX_STATE_OPEN);
+            if (current_state == (int32_t)ASYNC_SOCKET_LINUX_STATE_OPEN)
```
dcristoloveanu, line 363: > (int32_t) [](http://example.com/codeflow?start=32&length=10)

Likely the cast not needed

dcristoloveanu (approved):
:shipit:

PR #178, opened 2022-11-28
chore: fix async socket send method

jebrando:


mattdurak (approved):
:shipit:

linux/src/async_socket_linux.c
```diff
@@ -205,7 +205,7 @@ static int thread_worker_func(void* parameter)
                     {
                         void* send_pos = send_context->buffers[index].buffer + send_size;
                         uint32_t send_length = send_context->buffers[index].length - send_size;
-                        send_size += send(send_context->socket_handle, send_pos, send_length, 0);
+                        send_size += send(send_context->socket_handle, send_pos, send_length, MSG_NOSIGNAL);
```
dcristoloveanu, line 208: >MSG_NOSIGNAL [](http://example.com/codeflow?start=94&length=12)

I would love a comment about why we need this. No-one except The Jelani will know in a couple of weeks.

linux/src/async_socket_linux.c
```diff
@@ -298,6 +298,8 @@ static void internal_close(ASYNC_SOCKET_HANDLE async_socket)
         LogError("Failure removing socket from epoll_ctrl");
     }
 
+    LogInfo("async socket is closing socket: %" PRI_MU_SOCKET "", async_socket->socket_handle);
```
dcristoloveanu, line 301: >ogInfo("asyn [](http://example.com/codeflow?start=5&length=12)

Love the logging

dcristoloveanu, line 301: >PRI_MU_SOCKET [](http://example.com/codeflow?start=48&length=13)

This should be renamed please.
MU stands for macro utils. So this, not having anything to do with macro utils should be named PRI_SOCKET

dcristoloveanu: I noticed this when I picked it up last time in zrpc, but did not have the time to go back and fix it. But now that you are changing anyway c-pal and will have to propagate it makes sense to fix it.

dcristoloveanu:
>                 result = ASYNC_SOCKET_SEND_SYNC_OK;

I don't quite like this setting blindly of the result.
The reason we don't do this usually is because if we miss one place where we should be setting the result to something else, the compiler will not cry.

Now imagine that we do not have this line and we miss to set the result to something in the lines that follow.
The compiler will cry that we're returning an uninitialized variable!

So let's refactor it please.

---
Refers to: linux/src/async_socket_linux.c:522 in 560df9e. [](commit_id = 560df9ef2138040fba74009f21869a5c022d54a0, deletion_comment = False)

linux/src/async_socket_linux.c
jebrando: This would be a good thing for a spec, but all in due time.

PR #179, opened 2022-11-29
chore: adding a few fixes to threadpool and socket

jebrando:


linux/src/threadpool_linux.c
```diff
@@ -24,7 +24,8 @@
 
 #include "c_pal/threadpool.h"
 
-#define DEFAULT_TASK_ARRAY_SIZE 2048
+#define DEFAULT_TASK_ARRAY_SIZE         2048
+#define THREADPOOL_SEMEPHORE_TIMEOUT    1
```
mattdurak, line 28: >THREADPOOL_SEMEPHORE_TIMEOUT [](http://example.com/codeflow?start=8&length=28)

THREADPOOL_SEMAPHORE_TIMEOUT_S or ...SECONDS?

mattdurak: Is 1 second small enough? We could use `tv_nsec` instead and make it like 100ms or something

mattdurak (approved):
:shipit:

linux/src/threadpool_linux.c
jebrando: I can try that out, I wasn't sure how small an interval it should be.

linux/src/threadpool_linux.c
```diff
@@ -133,8 +134,8 @@ static int threadpool_work_func(void* param)
             }
             else
             {
-                // Setup the timeout for the semaphore to 2 sec
-                ts.tv_sec += 2;
+                // Setup the timeout for the semaphore
+                ts.tv_nsec += THREADPOOL_SEMEPHORE_MS;
```
mattdurak, line 138: >tv_nsec [](http://example.com/codeflow?start=19&length=7)

this is ns not ms so you have to multiply by 1000000

PR #180, opened 2022-11-30
chore: need to move pending api calls back to read function

jebrando:
Followed more towards the win32 implementation and move the pending_api_call back into the read function

### DIFF
--- a/linux/src/async_socket_linux.c
+++ b/linux/src/async_socket_linux.c
@@ -117,10 +117,6 @@ static int thread_worker_func(void* parameter)
                 if (recv_context != NULL)
                 {
                     recv_context->on_receive_complete(recv_context->on_receive_complete_context, ASYNC_SOCKET_RECEIVE_ABANDONED, 0);
-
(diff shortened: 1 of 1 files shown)


linux/src/async_socket_linux.c
```diff
@@ -117,10 +117,6 @@ static int thread_worker_func(void* parameter)
                 if (recv_context != NULL)
                 {
                     recv_context->on_receive_complete(recv_context->on_receive_complete_context, ASYNC_SOCKET_RECEIVE_ABANDONED, 0);
-
-                    (void)interlocked_decrement(&recv_context->async_socket->pending_api_calls);
```
anporumb, line 121: >pending_api_calls [](http://example.com/codeflow?start=77&length=17)

as there are no specs, what's there to review...


signing off

anporumb (approved):
:shipit:

mattdurak (approved):
:shipit:

linux/src/async_socket_linux.c
jebrando: Specs are coming soon.

PR #182, opened 2022-12-01
Changing return type of wait on address

ipg2013:


interfaces/inc/c_pal/sync.h
```diff
@@ -18,7 +18,15 @@
 extern "C" {
 #endif
 
-MOCKABLE_FUNCTION(, bool, wait_on_address, volatile_atomic int32_t*, address, int32_t, compare_value, uint32_t, timeout_ms);
+#define WAIT_ON_ADDRESS_RESULT_VALUES \
+    WAIT_ON_ADDRESS_OK, \
+    WAIT_ON_ADDRESS_ERROR, \
+    WAIT_ON_ADDRESS_TIMEOUT, \
+    WAIT_ON_ADDRESS_CHANGED
```
dcristoloveanu, line 25: Oh wow, this will RIPPLE!

common/src/call_once.c
```diff
@@ -51,7 +51,7 @@ CALL_ONCE_RESULT call_once_begin(call_once_t* state)
         {
             /*Codes_SRS_CALL_ONCE_02_003: [ If interlocked_compare_exchange returns 1 then call_once_begin shall call wait_on_address(state) with timeout UINT32_MAX and call again interlocked_compare_exchange(state, 1, 0). ]*/
             /*state cannot be "2" because while condition, so that means state is "1", and that means, need to wait until it is not 1*/
-            if (!wait_on_address(state, CALL_ONCE_STATE_CALLING, UINT32_MAX))
+            if (wait_on_address(state, CALL_ONCE_STATE_CALLING, UINT32_MAX) != WAIT_ON_ADDRESS_OK)
```
dcristoloveanu, line 54: >wait_on_address [](http://example.com/codeflow?start=16&length=15)

I'm a bit scared of all placed where we'll have to change the return type. We'll have to check really hard that we don't make a mistake.
wait_on_address spurious wakes could mean that maybe sometimes a bad change of the result could go unnoticed in tests ...

Just saying that you'll have to take extra care. I don't know of a better automated way to replace it.

interfaces/reals/real_sync.h
```diff
@@ -28,7 +28,7 @@ extern "C" {
 #endif
 
 
-bool real_wait_on_address(volatile_atomic int32_t* address, int32_t compare_value, uint32_t timeout_ms);
+WAIT_ON_ADDRESS_RESULT real_wait_on_address(volatile_atomic int32_t* address, int32_t compare_value, uint32_t timeout_ms);
```
dcristoloveanu, line 31: >uint32_t [](http://example.com/codeflow?start=101&length=8)

This needs an stdint include

dcristoloveanu (approved):
:shipit:

PR #184, opened 2022-12-02
Cpal: Add assert in UT/Int test wherever malloc/malloc_2 /malloc_flex call

ACKPaul:


interfaces/tests/interlocked_int/interlocked_int.c
```diff
@@ -289,6 +290,7 @@ TEST_FUNCTION(interlocked_compare_exchange_128_does_not_exchange_when_not_equal)
     volatile_atomic int64_t* destination;
     interlocked_exchange_pointer((void* volatile_atomic*)&destination, malloc(2 * sizeof(int64_t)));
     int64_t* comperand_result = (int64_t*)malloc(2 * sizeof(int64_t));
+    ASSERT_IS_NOT_NULL(comperand_result);
```
dcristoloveanu, line 293: I have to wonder if you did this with some regexp or manual.

ACKPaul: After finding all `= malloc` call, it was manual.

PR #187, opened 2022-12-04
feature: adding threadpool timer

jebrando:


linux/src/async_socket_linux.c
```diff
@@ -109,6 +110,13 @@ static int thread_worker_func(void* parameter)
     {
         struct epoll_event events[MAX_EVENTS_NUM];
         int num_ready = epoll_wait(g_epoll, events, MAX_EVENTS_NUM, EVENTS_TIMEOUT_MS);
+        if (num_ready == -1)
+        {
+            char err_msg[128] = {0};
+            (void)strerror_r(errno, err_msg, 128);
+            LogError("Failure epoll wait. Error: %s", MU_P_OR_NULL(err_msg));
```
anporumb, line 117: >err_msg [](http://example.com/codeflow?start=67&length=7)

this is never NULL.

linux/src/async_socket_linux.c
```diff
@@ -109,6 +110,13 @@ static int thread_worker_func(void* parameter)
     {
         struct epoll_event events[MAX_EVENTS_NUM];
         int num_ready = epoll_wait(g_epoll, events, MAX_EVENTS_NUM, EVENTS_TIMEOUT_MS);
+        if (num_ready == -1)
+        {
+            char err_msg[128] = {0};
```
anporumb, line 115: >{0} [](http://example.com/codeflow?start=32&length=3)

initialization is unnecessary.

linux/src/async_socket_linux.c
anporumb, line 117: >LogError [](http://example.com/codeflow?start=12&length=8)

just in care strerror fails... it should print errno too.

linux/src/async_socket_linux.c
```diff
@@ -109,6 +110,13 @@ static int thread_worker_func(void* parameter)
...
+            (void)strerror_r(errno, err_msg, 128);
```
anporumb, line 116: >strerror_r [](http://example.com/codeflow?start=18&length=10)

From `man strerror_r`:

`POSIX.1-2001  and  POSIX.1-2008 require that a successful call to strerror() or strerror_l() shall leave errno
       unchanged, and note that, since no function return value is reserved to indicate an error, an application that
       wishes  to  check  for  errors should initialize errno to zero before the call, and then check errno after the
       call.`

Looks like errno NEEDS to be initialized to 0 before epoll_wait.

linux/src/async_socket_linux.c
anporumb: Maybe it cannot fail... on a second reading of documentation.

linux/src/sync_linux.c
```diff
@@ -53,6 +52,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, WAIT_ON_ADDRESS_RESULT, wait_on_address, volatile_
         }
         else
         {
+            LogError("failure in syscall, errno=%d (%s)", errno, strerror(errno));
```
anporumb, line 55: >strerror [](http://example.com/codeflow?start=65&length=8)

we should decide on one or the other (strerror/strerror_r) functions to use.

errno is always threadsafe. strerror... "maybe".

anporumb: Also errno needs to be reset before the call and observed after the call. Or otherwise you can inherit some "old" errno.

This is sadly here and everywhere

anporumb:
>         result = WAIT_ON_ADDRESS_OK;

So... spurious wakeups are userland?

---
Refers to: linux/src/sync_linux.c:39 in 7968855. [](commit_id = 79688553cb4dd2e288a6cf5f93c05ed07195f44d, deletion_comment = False)

linux/src/threadpool_linux.c
```diff
@@ -77,6 +94,19 @@ typedef struct THREADPOOL_TAG
     THREAD_HANDLE* thread_handle_array;
 } THREADPOOL;
 
+void on_timer_callback(sigval_t timer_data)
```
anporumb, line 97: >void [](http://example.com/codeflow?start=0&length=4)

is this supposed to be static?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+
+            sigev.sigev_notify          = SIGEV_THREAD;
+            sigev.sigev_notify_function = on_timer_callback;
+            sigev.sigev_value.sival_ptr = timer_instance;
+
+            if (timer_create(CLOCK_REALTIME, &sigev, &time_id) != 0)
+            {
+                char err_msg[128] = {0};
+                (void)strerror_r(errno, err_msg, 128);
+                LogError("Failure calling timer_create. Error: %s", MU_P_OR_NULL(err_msg));
```
anporumb, line 572: looks like copy&paste from async_socket_linux.c, so all comments from there apply here too.

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+
+            // Setup the timer
+            struct sigevent sigev = {0};
+            timer_t time_id = 0;
+
+            sigev.sigev_notify          = SIGEV_THREAD;
+            sigev.sigev_notify_function = on_timer_callback;
+            sigev.sigev_value.sival_ptr = timer_instance;
+
+            if (timer_create(CLOCK_REALTIME, &sigev, &time_id) != 0)
```
anporumb, line 568: >CLOCK_REALTIME [](http://example.com/codeflow?start=29&length=14)

So it apeear that CLOCK_REALTIME is "settable" (according to man page).

What happens when the time is set forward? or backward? Such as when the time is adjusted to match daylight savings? Or NTP will bring a new time from the server? 

Will then the timer_period be observed agains the REALTIME? Or against what? 

Do we have tests for these cases?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+            {
+                char err_msg[128] = {0};
+                (void)strerror_r(errno, err_msg, 128);
+                LogError("Failure calling timer_create. Error: %s", MU_P_OR_NULL(err_msg));
+            }
+            else
+            {
+                struct itimerspec its = {0};
+                its.it_value.tv_sec = start_delay_ms / 1000;
```
anporumb, line 577: >start_delay_ms / 1000; [](http://example.com/codeflow?start=38&length=22)

I keep seeing this where we divide with N and get the quotient of N too at the same time.

Can "div" (or similar from stdlib.h) be used here?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+
+            if (timer_create(CLOCK_REALTIME, &sigev, &time_id) != 0)
+            {
+                char err_msg[128] = {0};
+                (void)strerror_r(errno, err_msg, 128);
+                LogError("Failure calling timer_create. Error: %s", MU_P_OR_NULL(err_msg));
+            }
+            else
+            {
+                struct itimerspec its = {0};
```
anporumb, line 576: >itimerspec [](http://example.com/codeflow?start=23&length=10)

man 3 itimerspec doesn't return anything, where's the general documentation for this?

anporumb: nvm, I found it with man 3 timer_settime.

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+                    *timer_handle = timer_instance;
+                    result = 0;
+                    goto all_ok;
+                }
+            }
+            free(timer_instance);
+        }
+        result = MU_FAILURE;
+        (void)interlocked_decrement(&threadpool->pending_call_count);
+        wake_by_address_all(&threadpool->pending_call_count);
```
anporumb, line 602: >wake_by_address_all [](http://example.com/codeflow?start=8&length=19)

How many threads are waiting (possibly) here?

And are they waiting for ANY value, or just "0" or "1" is interesting to them?

linux/src/threadpool_linux.c
anporumb, line 568: >timer_create [](http://example.com/codeflow?start=16&length=12)

Can these timers be cached/reused across calls to threadpool_timer_start?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+                else
+                {
+                    (void)interlocked_decrement(&threadpool->pending_call_count);
+                    wake_by_address_all(&threadpool->pending_call_count);
+                    timer_instance->time_id = time_id;
+                    *timer_handle = timer_instance;
+                    result = 0;
+                    goto all_ok;
+                }
+            }
```
anporumb, line 597: >} [](http://example.com/codeflow?start=12&length=1)

assume line 586 gets called. Then isn't the timer created in line 568 leaking?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+                {
+                    char err_msg[128] = {0};
+                    (void)strerror_r(errno, err_msg, 128);
+                    LogError("Failure calling timer_settime. Error: %s", MU_P_OR_NULL(err_msg));
+                }
+                else
+                {
+                    (void)interlocked_decrement(&threadpool->pending_call_count);
+                    wake_by_address_all(&threadpool->pending_call_count);
+                    timer_instance->time_id = time_id;
```
anporumb, line 592: >timer_instance->time_id = time_id; [](http://example.com/codeflow?start=20&length=34)

Does thsi impliy that between lines 591 and 592 a threadpool_destroy can come and it should work???

anporumb: What's the meaning of pending_call_count here?

linux/src/threadpool_linux.c
anporumb, line 97: >on_timer_callback [](http://example.com/codeflow?start=5&length=17)

how come this function doesn't decrement pending_api_call??

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+                }
+            }
+            free(timer_instance);
+        }
+        result = MU_FAILURE;
+        (void)interlocked_decrement(&threadpool->pending_call_count);
+        wake_by_address_all(&threadpool->pending_call_count);
+    }
+all_ok:
+    return result;
```
anporumb, line 605: >return result; [](http://example.com/codeflow?start=4&length=14)

I've read all of this function. I fail to see ANY connection between `threadpool` and the timer. Effectively starting 100 of the timers, will create 100 of the `timer_t`. 

`timer_t` will fire on "some thread", not on one of the threads of the threadpool.

Note: `SIGEV_THREAD
              Upon timer expiration, invoke sigev_notify_function as if it were the start function of a  new  thread.
              See sigevent(7) for details.`

I would expect SOME limitation of the thread executions here.

As in... if the threadpool was created with some `max_thread_count`... then... we'd better not go over that limit.

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
 {
-    (void)timer;
+    if (timer == NULL)
+    {
+        LogError("Invalid args: TIMER_INSTANCE_HANDLE timer = %p", timer);
+    }
+    else
+    {
+        struct itimerspec its;
+        memset(&its, 0, sizeof(its));
```
anporumb, line 651: can we use a global here initialized to "all 0" as opposed to build it every time?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+        }
+        else
+        {
+            result = 0;
+        }
+    }
+    return result;
 }
 
 void threadpool_timer_cancel(TIMER_INSTANCE_HANDLE timer)
```
anporumb, line 642: >threadpool_timer_cancel [](http://example.com/codeflow?start=5&length=23)

what's the relation between _cancel and pending_api_call?

linux/src/threadpool_linux.c
```diff
@@ -8,6 +8,14 @@
 #include <semaphore.h>
 #include <time.h>
 #include <string.h>
+#include <signal.h>
+#include <errno.h>
+
+#include <bits/types/__sigval_t.h>         // for __sigval_t
```
anporumb, line 14: >__sigval_t [](http://example.com/codeflow?start=50&length=10)

404 not found.

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+        {
+            LogError("Failure allocating Timer Instance");
+        }
+        else
+        {
+            timer_instance->work_function = work_function;
+            timer_instance->work_function_ctx = work_function_ctx;
+
+            // Setup the timer
+            struct sigevent sigev = {0};
```
anporumb, line 561: >sigevent [](http://example.com/codeflow?start=19&length=8)

documentation says sigevent comes from signal.h.

man sigevent

linux/src/threadpool_linux.c
```diff
@@ -8,6 +8,14 @@
...
 #include <time.h>
 #include <string.h>
+#include <signal.h>
+#include <errno.h>
+
+#include <bits/types/__sigval_t.h>         // for __sigval_t
+#include <bits/types/sigevent_t.h>         // for sigevent, sigev_notify_fun...
+#include <bits/types/sigval_t.h>           // for sigval_t
+#include <bits/types/struct_itimerspec.h>  // for itimerspec
+#include <bits/types/timer_t.h>            // for timer_t
```
anporumb, line 18: please reconsider these bits/types headers which i suspect are and all APis should be accessed through documented headers.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -230,4 +276,302 @@ TEST_FUNCTION(threadpool_chaos_knight)
...
+
+    EXECUTION_ENGINE_HANDLE execution_engine = execution_engine_create(&params);
+    ASSERT_IS_NOT_NULL(execution_engine);
+
+    // create the threadpool
+    THREADPOOL_HANDLE threadpool = threadpool_create(execution_engine);
+    ASSERT_IS_NOT_NULL(threadpool);
+
+    // open
+    ASSERT_ARE_EQUAL(int, 0, threadpool_open_async(threadpool, on_threadpool_open_complete, NULL));
```
anporumb, line 296: >on_threadpool_open_complete [](http://example.com/codeflow?start=63&length=27)

we are going to wait for this asynchronous function to be called or look the other way?

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
 
 int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms, uint32_t timer_period_ms, THREADPOOL_WORK_FUNCTION work_function, void* work_function_ctx, TIMER_INSTANCE_HANDLE* timer_handle)
```
anporumb, line 532: >threadpool_timer_start [](http://example.com/codeflow?start=4&length=22)

why isn't this function returning a NULL when it fails or otherwise a non-NULL value when it doesn't?

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -230,4 +276,302 @@ TEST_FUNCTION(threadpool_chaos_knight)
...
+        {
+            LogInfo("Waiting for timer to execute after short delay of no execution");
+
+            // Should eventually run once (wait up to 2.5 seconds, but it should run in 1.5 seconds)
+            wait_for_equal(&call_count, 1, 5000);
+            LogInfo("Timer completed, make sure it doesn't run again");
+
+            // And should not run again
+            ThreadAPI_Sleep(5000);
+            ASSERT_ARE_EQUAL(uint32_t, 1, (uint32_t)interlocked_add(&call_count, 0));
```
anporumb, line 331: >uint32_t [](http://example.com/codeflow?start=43&length=8)

why is this cast needed?

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -230,4 +276,302 @@ TEST_FUNCTION(threadpool_chaos_knight)
...
+        if (interlocked_add(&call_count, 0) != 0)
+        {
+            LogWarning("Timer ran after sleeping 500ms, we just got unlucky, try test again");
+        }
+        else
+        {
+            LogInfo("Waiting for timer to execute after short delay of no execution");
+
+            // Should eventually run once (wait up to 2.5 seconds, but it should run in 1.5 seconds)
+            wait_for_equal(&call_count, 1, 5000);
```
anporumb, line 326: >wait_for_equal [](http://example.com/codeflow?start=12&length=14)

Can this fail?

anporumb: ok ok ok it asserts if it fails. ok.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
anporumb: here and everywhere.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -230,4 +276,302 @@ TEST_FUNCTION(threadpool_chaos_knight)
...
+    interlocked_increment(&wait_work_context.wait_event);
+
+    // cleanup
+    threadpool_timer_destroy(timer);
+    threadpool_close(threadpool);
+    threadpool_destroy(threadpool);
+    execution_engine_dec_ref(execution_engine);
+}
+
 END_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
```
anporumb, line 577: >TEST_SUITE_NAME_FROM_CMAKE [](http://example.com/codeflow?start=15&length=26)

I am not going to stop progress with a wait.

"where specs?"

"where UTs"?

./signed off

anporumb (approved):
:shipit:

linux/src/async_socket_linux.c
```diff
@@ -109,6 +110,13 @@ static int thread_worker_func(void* parameter)
     {
         struct epoll_event events[MAX_EVENTS_NUM];
         int num_ready = epoll_wait(g_epoll, events, MAX_EVENTS_NUM, EVENTS_TIMEOUT_MS);
+        if (num_ready == -1)
+        {
```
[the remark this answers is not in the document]
jebrando: This is coming after we give this to, the MQTT guys.

linux/src/async_socket_linux.c
jebrando: Yes, it shouldn't fail, but having the errno is not a bad idea.

jebrando:
>         result = WAIT_ON_ADDRESS_OK;

I'm not understanding, this code too well.  I just had to move the LogError because it was spamming on any timeout.

---
In reply to: [1337618933](https://github.com/Azure/c-pal/pull/187#issuecomment-1337618933) [](http://example.com/codeflow?ancestors=1337618933,1337618933)

---
Refers to: linux/src/sync_linux.c:39 in 7968855. [](commit_id = 79688553cb4dd2e288a6cf5f93c05ed07195f44d, deletion_comment = False)

linux/src/threadpool_linux.c
jebrando: iwyu demands all the bits/types includes.  I'm not sure if this is correct or not.

linux/src/threadpool_linux.c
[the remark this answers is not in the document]
jebrando: These can be stack allocated the return value of time_id is stored in the timer_instance and that's the only thing that is needed.

jebrando: Yes, this time_id is stored and used again.

linux/src/threadpool_linux.c
jebrando: Yes, docs for this is not easily found

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+            {
+                char err_msg[128] = {0};
+                (void)strerror_r(errno, err_msg, 128);
+                LogError("Failure calling timer_create. Error: %s", MU_P_OR_NULL(err_msg));
+            }
+            else
+            {
+                struct itimerspec its = {0};
+                its.it_value.tv_sec = start_delay_ms / 1000;
+                its.it_value.tv_nsec = start_delay_ms * MILLISEC_TO_NANOSEC % 1000000000;
```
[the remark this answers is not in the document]
jebrando: I'm open to suggestions.  Win deals in milliseconds, but linux deals in seconds and microseconds.  It's the cost of doing business.

linux/src/threadpool_linux.c
jebrando: This is a hold over from windows, I don't think it's needed here.

linux/src/threadpool_linux.c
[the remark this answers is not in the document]
jebrando: Yes, this will determine how long between fires.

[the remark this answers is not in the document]
jebrando: Because this was written for windows first and linux as an afterthought.

linux/src/threadpool_linux.c
jebrando: iwyu is the culprit.

linux/src/threadpool_linux.c
```diff
@@ -501,29 +531,156 @@ int threadpool_schedule_work(THREADPOOL_HANDLE threadpool, THREADPOOL_WORK_FUNCT
...
+                its.it_interval.tv_sec = 0;
```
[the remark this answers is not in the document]
jebrando: I believe I need to use the timer_period_ms/1000 to get seconds.

linux/src/threadpool_linux.c
jebrando: this has been removed because it was a hold over from windows and not needed here.

linux/src/threadpool_linux.c
jebrando: In Win I believe there is a connection between the timer and the threadpool.  In linux they are separate entities.  The system takes care of the timer and the threadpool is controlled by us.  So this is just another instance of this API being written for windows.

linux/src/threadpool_linux.c
jebrando: All removed.

linux/src/threadpool_linux.c
jebrando: removed pending_api_call for this timer"

linux/tests/threadpool_linux_int/threadpool_linux_int.c
jebrando: I'm removing the open, it's not needed, this is a hold over from windows.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
jebrando: specs and UT will come soon.

PR #188, opened 2022-12-07
feature: adding doublylinklist and slist to repo

jebrando:


common/inc/c_pal/containing_record.h
```diff
@@ -0,0 +1,27 @@
...
+//
+
+#ifdef WIN32
+/*windows.h brings CONTAINING_RECORD*/
+#include "windows.h"
+#else
+#define CONTAINING_RECORD(address, type, field) ((type *)((uintptr_t)(address) - offsetof(type,field)))
+#endif
+
+#endif /* CONTAINING_RECORD_H */
```
mattdurak, line 27: ha I guess this could have already been in the PAL with separate headers in the win32 and linux folders. Maybe it should just be two headers that way instead /consider

common/devdoc/doublylinkedlist_requirements.md
```diff
@@ -0,0 +1,154 @@
+DoublyLinkedList Requirements
```
mattdurak, line 1: >DoublyLinkedList [](http://example.com/codeflow?start=0&length=16)

I guess as discussed this stays in c-util

mattdurak:
assuming dlist will move back, signing off otherwise

mattdurak (dismissed):
:shipit:

linux/src/async_socket_linux.c
```diff
@@ -101,6 +109,67 @@ typedef struct ASYNC_SOCKET_SEND_CONTEXT_TAG
...
+    if (s_list_remove(&async_socket->recv_data_head, &recv_context->link) != 0)
+    {
+        LogError("failure removing receive data to list");
+        result = MU_FAILURE;
+    }
+    else
+    {
+        result = 0;
+    }
+    (void)interlocked_exchange(&async_socket->recv_data_access, 0);
```
mattdurak, line 138: >interlocked_exchange [](http://example.com/codeflow?start=10&length=20)

missing wake after this line (here and below)

linux/src/async_socket_linux.c
```diff
@@ -300,12 +386,25 @@ static void internal_close(ASYNC_SOCKET_HANDLE async_socket)
...
     }
 
-    LogInfo("async socket is closing socket: %" PRI_SOCKET "", async_socket->socket_handle);
-
     // Close the socket
     (void)close(async_socket->socket_handle);
     async_socket->socket_handle = INVALID_SOCKET;
 
+    // Ensure that all the cleanup on the epoll is complete
+    if ((value = interlocked_add(&async_socket->epoll_cleanup, 0)) != 1)
```
mattdurak, line 394: > if [](http://example.com/codeflow?start=3&length=3)

interlocked+wait_on_address seems like it should be a while loop....

mattdurak (approved):
:shipit:

PR #194, opened 2022-12-20
Remove cyclic dependences

ACKPaul:


common/reals/real_s_list.h
```diff
@@ -5,7 +5,6 @@
 #define REAL_S_LIST_H
 
 #include "macro_utils/macro_utils.h"
-#include "c_util/s_list.h"
```
mattdurak, line 8: >c_util [](http://example.com/codeflow?start=10&length=6)

I think we should switch to `c_pal` instead of removing because this header references `PS_LIST_ENTRY` from that file.

I saw in other reals here we aren't consistent about doing that (like execution_engine), but we should really include what is used by the file so there isn't a weird dependency in the includes.

mattdurak (approved):
:shipit:

common/reals/real_s_list.h
ACKPaul: Thank you.Added.

PR #197, opened 2022-12-26
chore: fix helgrind race conditions

jebrando:


linux/src/async_socket_linux.c
```diff
@@ -50,6 +54,7 @@ MU_DEFINE_ENUM_STRINGS(ASYNC_SOCKET_SEND_SYNC_RESULT, ASYNC_SOCKET_SEND_SYNC_RES
 #define THREAD_COUNT    1
 
 static volatile_atomic int32_t g_thread_access_cnt;
+static volatile_atomic int32_t g_epoll_access = 0;
 static int g_epoll = -1;
 static THREAD_HANDLE g_thread_array[THREAD_COUNT];
```
anporumb, line 59: >THREAD_COUNT [](http://example.com/codeflow?start=36&length=12)

This seems to be just 1. (line 54) and it also seems to be referenced as such for example in `initialize_global_thread`. Note singular "thread".

anporumb:
>     EXECUTION_ENGINE_HANDLE execution_engine;

I cannot understand why this is used.

It is only ever inc_ref'd once and dec_ref'd once (in _create/_destroy).

Looks like it can be removed.

Also the line from .md: "[...]in order to be able to schedule all asynchronous work using the execution engine[...]" should also be removed.

---
Refers to: linux/src/async_socket_linux.c:64 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>             LogError("failure allocating asyn socket %zu", sizeof(ASYNC_SOCKET));

missing "c"

---
Refers to: linux/src/async_socket_linux.c:532 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>     int result = -1;

should not initialized the function result. Suspicion initialization can be postponed for later.

---
Refers to: linux/src/async_socket_linux.c:376 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -50,6 +54,7 @@ MU_DEFINE_ENUM_STRINGS(ASYNC_SOCKET_SEND_SYNC_RESULT, ASYNC_SOCKET_SEND_SYNC_RES
 #define THREAD_COUNT    1
 
 static volatile_atomic int32_t g_thread_access_cnt;
```
anporumb, line 56: >g_thread_access_cnt [](http://example.com/codeflow?start=31&length=19)

I find this VERY weird that it is not initialized to "anything".

I find it with interlocked_increment and interlocked_decrement in the file all over the place, but what's its initial value? 

And by the way, HOW did this pass valgrind ? I might be wrong and it COULD be initialized some place, just that I cannot find where that place is.

anporumb:
> static int initialize_global_thread(void)

There seem to be multiple mechanisms here (in this function) that are targetting similar effects:

1) looks like `current_count` wants to limit "something" (lines 383-425) to just 1 thread.
2) g_epoll_access looks similar, as it incremens in line 383 and decrements in line 421.
3) lines 424-433 are missing any sort of Codes_something, but looks like errr... it is waiting for the value of g_epoll_access to be 0, at which point it returns whatever was last stored in g_epoll.

Can I please beg for consideration for call_once/lazy_init APIs here? They might already do what this code is doing. Specially seeing as the code is dealing with `g_` called variables.

---
Refers to: linux/src/async_socket_linux.c:374 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 }

this code is not needed because we ever only have 1 thread.

---
Refers to: linux/src/async_socket_linux.c:416 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 }

I am going to say "here and everywhere" and be done with that remark.

---
In reply to: [1366197403](https://github.com/Azure/c-pal/pull/197#issuecomment-1366197403) [](http://example.com/codeflow?ancestors=1366197403)

---
Refers to: linux/src/async_socket_linux.c:416 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>         LogErrorNo("Failure epoll_ctrl with EPOLL_CTL_DEL");

function name is epoll_ctl

---
Refers to: linux/src/async_socket_linux.c:475 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>     ThreadAPI_Sleep(1);

why not 42.

I don't think the mechanism to close epoll/socket should/is/can ever be Sleep based.

At this moment, the socket file descripotor is detached from the "interest list" (see `man 7 epoll`).


---
Refers to: linux/src/async_socket_linux.c:480 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>     ThreadAPI_Sleep(1);

also missing specs, if this is REALLY required.

---
Refers to: linux/src/async_socket_linux.c:480 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>             LogWarning("Not open, current_state=%" PRI_MU_ENUM "", MU_ENUM_VALUE(ASYNC_SOCKET_LINUX_STATE, current_state));

it can be printed with PRI_MU_ENUM.

---
Refers to: linux/src/async_socket_linux.c:683 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>             LogWarning("Not open, current_state=%" PRI_MU_ENUM "", MU_ENUM_VALUE(ASYNC_SOCKET_LINUX_STATE, current_state));

Looks like a case of "here and everywhere".

---
In reply to: [1366202955](https://github.com/Azure/c-pal/pull/197#issuecomment-1366202955) [](http://example.com/codeflow?ancestors=1366202955)

---
Refers to: linux/src/async_socket_linux.c:683 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -944,6 +964,10 @@ int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BU
                         struct epoll_event ev = {0};
                         ev.events = EPOLLIN | EPOLLRDHUP | EPOLLONESHOT;
                         ev.data.ptr = recv_context;
+                        // This is telling valgrind that this will happne before the epoll gets called
```
anporumb, line 967: >happne [](http://example.com/codeflow?start=67&length=6)

happen

anporumb:
>     while ((value = interlocked_add(&async_socket->pending_api_calls, 0)) != 0)

So assume there's a `async_socket_receive_async` ongoing which did get to execute interlocked_increment(pending_api_calls).

What makes it go to zero so that line 467 passes?

---
Refers to: linux/src/async_socket_linux.c:467 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>         int32_t current_state = interlocked_compare_exchange(&async_socket->state, (int32_t)ASYNC_SOCKET_LINUX_STATE_OPENING, (int32_t)ASYNC_SOCKET_LINUX_STATE_CLOSED);

why is this cast necessary??

---
Refers to: linux/src/async_socket_linux.c:622 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                                 send_context->async_socket = async_socket;

why does send_context need to remember "socket_handle" AND "async_socket" when "async_socket" already contains "socket_handle"?

---
Refers to: linux/src/async_socket_linux.c:785 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 result = ASYNC_SOCKET_SEND_SYNC_ABANDONED;

Interesting choice, I was under the impression that "abandoned" is when "it was OK once and then devolved into ABANDONED because something else happened" (such as request to disconnect or a natural disconnect from the peer).

Why not return immediately ERROR here?

---
Refers to: linux/src/async_socket_linux.c:752 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
> static int send_data(ASYNC_SOCKET* async_socket, const ASYNC_SOCKET_BUFFER* buff_dat)

This doesn't work properly.

Assume there's 1000 bytes to be send.

Assume the loop between lines 181-197 is executed 5 times each time returning "100 bytes have been accepted for sending" (so 500 bytes of the 1000 are handed over to the socket for sending).

Then the loop dies because E_WOULDBLOCK.

---
Refers to: linux/src/async_socket_linux.c:176 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
> static int send_data(ASYNC_SOCKET* async_socket, const ASYNC_SOCKET_BUFFER* buff_dat)

Line 765 picks this error code up and REQUEUES all the bytes to be send (because the information carried by the local `ssize_t data_sent` is lost).

---
In reply to: [1366225362](https://github.com/Azure/c-pal/pull/197#issuecomment-1366225362) [](http://example.com/codeflow?ancestors=1366225362)

---
Refers to: linux/src/async_socket_linux.c:176 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>         ssize_t send_size = send(async_socket->socket_handle, buff_dat->buffer, buff_dat->length + data_sent, MSG_NOSIGNAL);

line 187 should capture errno "as is" before the code drifts and there's something else inserted between "send" and the place where errno is consumed, which is 600 lines of code down in the file.

---
Refers to: linux/src/async_socket_linux.c:184 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                                 // Codes_SRS_ASYNC_SOCKET_LINUX_11_057: [ The the context shall then be added to the epoll system by calling epoll_ctl with EPOLL_CTL_MOD. ]

The the

---
Refers to: linux/src/async_socket_linux.c:793 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                                 if (epoll_ctl(async_socket->epoll, EPOLL_CTL_MOD, async_socket->socket_handle, &ev) < 0)

So why do we need to re-run epoll_ctl every time when the send failed with EAGAIN or EWOULDBLOCK?

My understanding is that "once registered, it stays registered"...?

---
Refers to: linux/src/async_socket_linux.c:794 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 wake_by_address_single(&async_socket->pending_api_calls);

If any value of interest for wake_by_address_single or just "0"? If only 0 (or 0 and 1... )consider calling wake_by_address only when the "interesting" values happen.

---
Refers to: linux/src/async_socket_linux.c:857 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 wake_by_address_single(&async_socket->pending_api_calls);

Here and everywhere

---
In reply to: [1366229695](https://github.com/Azure/c-pal/pull/197#issuecomment-1366229695) [](http://example.com/codeflow?ancestors=1366229695)

---
Refers to: linux/src/async_socket_linux.c:857 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

build_functions/c-pal_suppressions.sup
```diff
@@ -1,9 +1,3 @@
-{
-   <epoll thread suppression>
-   Helgrind:Race
-   fun:thread_worker_func
```
anporumb, line 4: >thread_worker_func [](http://example.com/codeflow?start=7&length=18)

Cannot help but noticing that .c, ut and int are changed but no changes in the .md...

anporumb:
> #include <unistd.h>

there's no mention of "unistd.h" in C11, so I think it should stay with the other Linux headers.

---
Refers to: linux/src/async_socket_linux.c:6 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>             // Codes_SRS_ASYNC_SOCKET_LINUX_11_081: [ If the events value contains EPOLLRDHUP (hang up), thread_worker_func shall the following: ]

shall `do` the following:

---
Refers to: linux/src/async_socket_linux.c:218 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
> int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BUFFER* payload, uint32_t buffer_count, ON_ASYNC_SOCKET_RECEIVE_COMPLETE on_receive_complete, void* on_receive_complete_context)

what is the threading model here?

Just asking because 
1) if receive_async can be called concurrently then it should be explicitly stated so in requirements (and would be nice to have some Codes_SRS... around `add_item_to_list` and co
2) if it should not be called concurrently then there's no mitigation (critical_section?) against that.
3) if it cannot be called concurrently then documentation makes no mention of that.

So one of the three needs to happen. Likely "here and everywhere" and seeing as there are 

`interlocked_add(&async_socket->state, 0)` and `interlocked_increment(&async_socket->pending_api_calls);` 

it looks like we are in case 1).

---
Refers to: linux/src/async_socket_linux.c:866 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                 (void)interlocked_increment(&async_socket->pending_api_calls);

traditionally we increment the pending_api_call as the 1st line in the code and AFTER we look at the state. There's threading reasons to do that.

Here and everywhere.

---
Refers to: linux/src/async_socket_linux.c:930 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>         int num_ready = epoll_wait(g_epoll, events, MAX_EVENTS_NUM, EVENTS_TIMEOUT_MS);

shouldn't g_epoll be protected somehow here by the same mechanism as in other places with `g_epoll_access` interlocked?

---
Refers to: linux/src/async_socket_linux.c:210 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>             LogErrorNo("Failure epoll_wait, MAX_EVENTS_NUM: %d, EVENTS_TIMEOUT_MS: %d", MAX_EVENTS_NUM, EVENTS_TIMEOUT_MS);

If this ever fails, it is likely it will fail (again) when the loop goes around, so a Sleep of some sort would be appreciated, otherwise the logging will suffer.

---
Refers to: linux/src/async_socket_linux.c:213 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -215,6 +220,9 @@ static int thread_worker_func(void* parameter)
             {
                 // Codes_SRS_ASYNC_SOCKET_LINUX_11_082: [ thread_worker_func shall receive the ASYNC_SOCKET_RECV_CONTEXT value from the ptr variable from the epoll_event data ptr. ]
                 ASYNC_SOCKET_RECV_CONTEXT* recv_context = events[index].data.ptr;
+                #ifdef USE_VALGRIND
+                    ANNOTATE_HAPPENS_AFTER(recv_context);
+                #endif
                 if (recv_context != NULL)
```
anporumb, line 226: >recv_context != NULL [](http://example.com/codeflow?start=20&length=20)

can this ever be `NULL`? If not, then why check it? 

Also noticing a lack of `else` here...

anporumb: Same in line 252

anporumb:
>                         recv_size = recv(recv_context->socket_handle, recv_context->recv_buffers[index].buffer, recv_context->recv_buffers[index].length, 0);

So if there 100 of the receive_async "parked", each asking for 2 bytes to be received

And 1 byte comes on the wire.

then do we:

1) do nothing
2) call 1 callback with 1 byte
3) still loop over all the rest of the 99 parked receives?

I find this function very hard to follow :(


---
Refers to: linux/src/async_socket_linux.c:259 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>         // Codes_SRS_ASYNC_SOCKET_LINUX_11_080: [ Onced signaled thread_worker_func shall loop through the signaled epolls. ]

Once

---
Refers to: linux/src/async_socket_linux.c:215 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>                         recv_size = recv(recv_context->socket_handle, recv_context->recv_buffers[index].buffer, recv_context->recv_buffers[index].length, 0);

Is there an INT test for that?

---
In reply to: [1366257613](https://github.com/Azure/c-pal/pull/197#issuecomment-1366257613) [](http://example.com/codeflow?ancestors=1366257613)

---
Refers to: linux/src/async_socket_linux.c:259 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>     } while (*accept_socket == INVALID_SOCKET && errno == 11);

11?

---
Refers to: linux/tests/async_socket_linux_int/async_socket_linux_int.c:154 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
>     } while (*accept_socket == INVALID_SOCKET && errno == 11);

Is this some Stranger Things reference that I am not familiar with?

---
In reply to: [1366258331](https://github.com/Azure/c-pal/pull/197#issuecomment-1366258331) [](http://example.com/codeflow?ancestors=1366258331)

---
Refers to: linux/tests/async_socket_linux_int/async_socket_linux_int.c:154 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb:
> }

I will be waiting because I don't think this works as intended. And cheerfully look on the next iteration of the file.



---
Refers to: build_functions/c-pal_suppressions.sup:30 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb (changes requested):
:clock1:

dcristoloveanu:
>         LogError("EXECUTION_ENGINE_HANDLE execution_engine:%p, SOCKET_HANDLE socket_handle:%" PRI_SOCKET "", execution_engine, socket_handle);

The typical way we log invalid args is 
"Invalid arguments EXECUTION_ENGINE_HANDLE execution_engine=%p, SOCKET_HANDLE socket_handle=% ..."

Note the "Invalid arguments" and usage of "="

---
Refers to: linux/src/async_socket_linux.c:524 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
jebrando: I believe that you're correct it would not be NULL

jebrando:
>                 }

This code is here because we may add more threads if the perf is not sufficient.  We will have to wait to get more information from the mqtt runs.

---
In reply to: [1366197547](https://github.com/Azure/c-pal/pull/197#issuecomment-1366197547) [](http://example.com/codeflow?ancestors=1366197547,1366197403)

---
Refers to: linux/src/async_socket_linux.c:416 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
jebrando: We may have multiple threads depending on perf performance.

jebrando:
>             LogWarning("Not open, current_state=%" PRI_MU_ENUM "", MU_ENUM_VALUE(ASYNC_SOCKET_LINUX_STATE, current_state));

Not sure what you mean by this one, it is using PRI_MU_ENUM to print the value

---
In reply to: [1366203507](https://github.com/Azure/c-pal/pull/197#issuecomment-1366203507) [](http://example.com/codeflow?ancestors=1366203507,1366202955)

---
Refers to: linux/src/async_socket_linux.c:683 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>     while ((value = interlocked_add(&async_socket->pending_api_calls, 0)) != 0)

This is decremented at 1015, it so we don't close while we're in the recv.

---
In reply to: [1366209175](https://github.com/Azure/c-pal/pull/197#issuecomment-1366209175) [](http://example.com/codeflow?ancestors=1366209175,1366209175)

---
Refers to: linux/src/async_socket_linux.c:467 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>         int32_t current_state = interlocked_compare_exchange(&async_socket->state, (int32_t)ASYNC_SOCKET_LINUX_STATE_OPENING, (int32_t)ASYNC_SOCKET_LINUX_STATE_CLOSED);

It is not probably a carry over from the win32 version.

---
In reply to: [1366217286](https://github.com/Azure/c-pal/pull/197#issuecomment-1366217286) [](http://example.com/codeflow?ancestors=1366217286,1366217286)

---
Refers to: linux/src/async_socket_linux.c:622 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>                                 if (epoll_ctl(async_socket->epoll, EPOLL_CTL_MOD, async_socket->socket_handle, &ev) < 0)

Normally send would not setup epoll_ctl, it just sends the data, but if we get a EAGAIN then it has to use epoll to send later. This makes sure that in the event we have EPOLLOUT telling epoll that we have a send waiting.

---
In reply to: [1366229044](https://github.com/Azure/c-pal/pull/197#issuecomment-1366229044) [](http://example.com/codeflow?ancestors=1366229044,1366229044)

---
Refers to: linux/src/async_socket_linux.c:794 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>                 result = ASYNC_SOCKET_SEND_SYNC_ABANDONED;

This is a hold over from the WIN32 version.  I think you are correct that this should be just an ERROR.  I will change it here.

---
In reply to: [1366223915](https://github.com/Azure/c-pal/pull/197#issuecomment-1366223915) [](http://example.com/codeflow?ancestors=1366223915,1366223915)

---
Refers to: linux/src/async_socket_linux.c:752 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
dcristoloveanu, line 56: >nt; [](http://example.com/codeflow?start=48&length=3)

We should be explicit (or consistent - below we do initialize with 0 explicitly)

linux/src/async_socket_linux.c
```diff
@@ -370,12 +380,14 @@ static int initialize_global_thread(void)
     // Codes_SRS_ASYNC_SOCKET_LINUX_11_009: [ If the g_thread_access_cnt count is 1, initialize_global_thread shall do the following: ]
     if (current_count == 1)
     {
+        interlocked_increment(&g_epoll_access);
```
dcristoloveanu, line 383: >interlocked_increment [](http://example.com/codeflow?start=8&length=21)

(void)

linux/src/async_socket_linux.c
```diff
@@ -370,12 +380,14 @@ static int initialize_global_thread(void)
...
+
```
dcristoloveanu, line 384: I think generally this reference counting is not needed. What is the case that fails if we don't have it?

linux/src/async_socket_linux.c
```diff
@@ -409,9 +421,17 @@ static int initialize_global_thread(void)
                 interlocked_decrement(&g_thread_access_cnt);
```
dcristoloveanu, line 421: >interlocked_decrement [](http://example.com/codeflow?start=16&length=21)

(void)

dcristoloveanu:
>             interlocked_exchange(&async_socket->state, ASYNC_SOCKET_LINUX_STATE_CLOSED);

(void)

---
Refers to: linux/src/async_socket_linux.c:661 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

dcristoloveanu:
> static int thread_worker_func(void* parameter)

I think we should definitely have a handle that holds together an epoll object and a thread that is polling it.
That is like a completion port. SO maybe we should name it as such.

Once we do that we would:
- split the complexity out of async_socket. async_socket should only implement the API and the workhorse for epoll polling should be done by this new to be developed unit.

I think developing it as a separate task would be fine, does not have to be bound to this PR.

---
Refers to: linux/src/async_socket_linux.c:201 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

dcristoloveanu:
>     async_socket->socket_handle = INVALID_SOCKET;

This should never be needed

---
Refers to: linux/src/async_socket_linux.c:484 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

dcristoloveanu:
>     async_socket->socket_handle = INVALID_SOCKET;

If it is needed it means we're accessing the socket after we closed it - which we should never do, should we?

---
In reply to: [1367730576](https://github.com/Azure/c-pal/pull/197#issuecomment-1367730576) [](http://example.com/codeflow?ancestors=1367730576)

---
Refers to: linux/src/async_socket_linux.c:484 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -409,9 +421,17 @@ static int initialize_global_thread(void)
...
+        (void)interlocked_decrement(&g_epoll_access);
+        wake_by_address_single(&g_epoll_access);
     }
     else
     {
+        int32_t value;
+        while ((value = interlocked_add(&g_epoll_access, 0)) != 0)
+        {
+            (void)wait_on_address(&g_epoll_access, value, UINT32_MAX);
+        }
```
dcristoloveanu, line 433: In the light of "let's move the epoll+thread to another unit I think this counting would not be needed at all (would be handled by this other unit fully.

dcristoloveanu: So maybe we should remove this counting altogether.

dcristoloveanu:
>             break;

If the send fails, we should not send again the same bytes.
We really have to fix this before we do more work, as this WILL fail for sure in prod usage.

---
Refers to: linux/src/async_socket_linux.c:189 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

dcristoloveanu (changes requested):
:clock1:

jebrando:
> #include <unistd.h>

unistd.h is where close is defined: [https://linux.die.net/man/3/close](https://linux.die.net/man/3/close)

---
In reply to: [1366240264](https://github.com/Azure/c-pal/pull/197#issuecomment-1366240264) [](http://example.com/codeflow?ancestors=1366240264,1366240264)

---
Refers to: linux/src/async_socket_linux.c:6 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>     } while (*accept_socket == INVALID_SOCKET && errno == 11);

EAGAIN, will add the word.

---
In reply to: [1366258444](https://github.com/Azure/c-pal/pull/197#issuecomment-1366258444) [](http://example.com/codeflow?ancestors=1366258444,1366258331)

---
Refers to: linux/tests/async_socket_linux_int/async_socket_linux_int.c:154 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>                         recv_size = recv(recv_context->socket_handle, recv_context->recv_buffers[index].buffer, recv_context->recv_buffers[index].length, 0);

We do have an INT test with 1000 receives, but we can't tell epoll how many bytes to give us, it does it on its own.  What we will do is when we get bytes we send it up the chain, the recv will give us all the bytes that it has, so if there are 100 bytes to recv, we will get as many as the buffer will allow.

---
In reply to: [1366257892](https://github.com/Azure/c-pal/pull/197#issuecomment-1366257892) [](http://example.com/codeflow?ancestors=1366257892,1366257613)

---
Refers to: linux/src/async_socket_linux.c:259 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

jebrando:
>                 (void)interlocked_increment(&async_socket->pending_api_calls);

We probably should look at the WIN32 version also since this is a port from that version, then its probably wrong there too.

---
In reply to: [1366253916](https://github.com/Azure/c-pal/pull/197#issuecomment-1366253916) [](http://example.com/codeflow?ancestors=1366253916,1366253916)

---
Refers to: linux/src/async_socket_linux.c:930 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

anporumb (approved):
:shipit:

linux/src/async_socket_linux.c
jebrando: Yes, I will move this, but I'd like to move it in the next PR.

jebrando:
> static int initialize_global_thread(void)

This will be moved to a new module

---
In reply to: [1366193562](https://github.com/Azure/c-pal/pull/197#issuecomment-1366193562) [](http://example.com/codeflow?ancestors=1366193562,1366193562)

---
Refers to: linux/src/async_socket_linux.c:374 in 29def71. [](commit_id = 29def7169ece4e317350d56a7821a6caf39011a3, deletion_comment = False)

linux/src/async_socket_linux.c
jebrando: We will move this to a different module, task has been set

linux/src/async_socket_linux.c
```diff
@@ -168,17 +170,18 @@ static int add_item_to_list(ASYNC_SOCKET* async_socket, ASYNC_SOCKET_RECV_CONTEX
...
+static int send_data(ASYNC_SOCKET* async_socket, const ASYNC_SOCKET_BUFFER* buff_data, ssize_t* total_data_sent, int* error_no)
 {
     int result;
 
     ssize_t data_sent = 0;
     do
     {
         // Codes_SRS_ASYNC_SOCKET_LINUX_11_052: [ async_socket_send_async shall attempt to send the data by calling send with the MSG_NOSIGNAL flag to ensure an exception is not generated. ]
-        ssize_t send_size = send(async_socket->socket_handle, buff_dat->buffer, buff_dat->length + data_sent, MSG_NOSIGNAL);
+        ssize_t send_size = send(async_socket->socket_handle, buff_data->buffer+data_sent, buff_data->length-data_sent, MSG_NOSIGNAL);
```
dcristoloveanu, line 181: >send [](http://example.com/codeflow?start=28&length=4)

I think this can return EWOULDBLOCK, can't it?

linux/src/async_socket_linux.c
```diff
@@ -406,12 +411,22 @@ static int initialize_global_thread(void)
...
                 (void)close(g_epoll);
                 g_epoll = -1;
-                interlocked_decrement(&g_thread_access_cnt);
+                (void)interlocked_decrement(&g_thread_access_cnt);
             }
         }
+        if (interlocked_decrement(&g_epoll_access) == 0)
+        {
+            wake_by_address_single(&g_epoll_access);
+        }
```
dcristoloveanu, line 420: > } [](http://example.com/codeflow?start=7&length=2)

Wouldn't the else branch of this if be horribly and utterly bad?
It would mean someone has incremented the g_epoll_access while we're initializing? That's just pure evil

linux/src/async_socket_linux.c
```diff
@@ -406,12 +411,22 @@ static int initialize_global_thread(void)
                 result = -1;
                 (void)close(g_epoll);
                 g_epoll = -1;
-                interlocked_decrement(&g_thread_access_cnt);
+                (void)interlocked_decrement(&g_thread_access_cnt);
             }
         }
+        if (interlocked_decrement(&g_epoll_access) == 0)
```
dcristoloveanu, line 417: >g_epoll_access [](http://example.com/codeflow?start=35&length=14)

So, what happens if we delete g_epoll_access completely?
We should talk about this, I have a very uneasy feeling  introducing it when I don't understand really why we need it.

PR #198, opened 2022-12-28
Minor linux build fix.

darobs:
Use sys/syscall.h and sys/types.h

### DIFF
--- a/linux/src/sync_linux.c
+++ b/linux/src/sync_linux.c
@@ -7,8 +7,7 @@
 #include <time.h>
 #include <string.h>
 #include <errno.h>
-#include <sys/syscall.h>
-#include <sys/types.h>
+#include <syscall.h>
 
 #include "c_logging/xlogging.h"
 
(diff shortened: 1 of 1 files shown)


linux/src/sync_linux.c
```diff
@@ -7,10 +7,10 @@
 #include <time.h>
 #include <string.h>
 #include <errno.h>
+#include <syscall.h>
```
dcristoloveanu, line 10: Cannot really find syscall in C11, this is a Linux thing, so it should be where it was, next to the linux one in line 14 me thinks.

Aaaaand when I look at it : [https://man7.org/linux/man-pages/man2/syscall.2.html](https://man7.org/linux/man-pages/man2/syscall.2.html)

it seems to be used with the sys/ "namespacing"

What prompted this change?

dcristoloveanu (changes requested):
:clock1:

linux/src/sync_linux.c
darobs: I was having build problems, and syscall was one of the things that was giving me trouble. Now that I've got everything else resolved, this change doesn't look needed.

PR #199, opened 2023-01-05
feature: adding completion port linux spec and header

jebrando:


linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+{
+    int epoll_op;
+    SOCKET_HANDLE socket;
+    ON_COMPLETION_PORT_EVENT_COMPLETE event_callback;
+    void* event_callback_ctx;
+    EPOLL_BUFFER_DATA buffer_data;
+} COMPLETION_PORT_INGRESS_DATA;
+
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, completion_port_create);
+MOCKABLE_FUNCTION(, void, completion_port_dec_ref, COMPLETION_PORT_HANDLE, completion_port);
```
anporumb, line 44: >completion_port_dec_ref [](http://example.com/codeflow?start=26&length=23)

would be SOOO nice when we wouldn't have to dec_ref and inc_ref when we have THANDLE. Maybe move THANDLE here in c-pal?

Trigger happy comment, as I haven't perused the rest of the document.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+`completion_port_dec_ref` handles the decrement of the reference count and freeing the memory if necessary
+
+`completion_port_dec_ref` shall decrement the reference count for `completion_port`.
+
+If the reference count reaches 0, `completion_port_dec_ref` shall do the following:
+
+- increment the flag signaling that the threads can complete.
```
anporumb, line 80: >increment the flag [](http://example.com/codeflow?start=2&length=18)

This is weird, reading until here, this is the first reference to "flag". Also the only one, so I don't know what this "flag" is.

Also semantically talking, "flags" cannot really be "incremented". I always have this model where the "flag" is the little thingie attached to my mail box that I raise when there's mail inside, and the mailman lowers once he picks my mail up.

anporumb, line 80: >the threads [](http://example.com/codeflow?start=36&length=11)

which "threads" (note: plural)?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+`completion_port_post_queue` adds an ingress item to the post queue for epoll signaling
+
+If `completion_port` is `NULL`, `completion_port_post_queue` shall return a non-NULL value.
+
+If `ingress_data` is `NULL`, `completion_port_post_queue` shall return a non-NULL value.
+
+`completion_port_post_queue` shall allocate a thread data object
+
+`completion_port_post_queue` shall allocate a thread data object
```
anporumb, line 114: suppose allocating it twice (once here, once in line 112) is unintentional victim of copy&paste?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+`completion_port_post_queue` shall add the socket in the epoll system by calling `epoll_ctl` with `EPOLL_CTL_MOD` along with the `COMPLETION_PORT_INGRESS_DATA` `epoll_op` variable.
```
anporumb, line 116: >along with the `COMPLETION_PORT_INGRESS_DATA` `epoll_op` variable [](http://example.com/codeflow?start=114&length=65)

along with what... ? Just saying that this does not compute for me, as 

```c
int epoll_ctl(int epfd, int op, int fd, struct epoll_event *event);
```

epoll_ctl needs some `epoll_event` argument. epoll_op is "int" type. 

I am completely confused at this moment, because where is `event` coming from?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+    int epoll_op;
+    SOCKET_HANDLE socket;
+    ON_COMPLETION_PORT_EVENT_COMPLETE event_callback;
+    void* event_callback_ctx;
+    EPOLL_BUFFER_DATA buffer_data;
+} COMPLETION_PORT_INGRESS_DATA;
+
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, completion_port_create);
+MOCKABLE_FUNCTION(, void, completion_port_dec_ref, COMPLETION_PORT_HANDLE, completion_port);
+MOCKABLE_FUNCTION(, void, completion_port_inc_ref, COMPLETION_PORT_HANDLE, completion_port);
```
anporumb, line 45: >completion_port_inc_ref [](http://example.com/codeflow?start=26&length=23)

What's the threading model of this module?

Can any API be called from any thread? Can multiple threads call "post_queue" at the same time?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, completion_port_create);
+```
+
+`completion_port_create` creates and initializes the completion port module
+
+`completion_port_create` shall create a completion port object.
+
+`completion_port_create` shall create the epoll module by calling `epoll_create`.
+
+`completion_port_create` shall create a thread to handle the epoll events.
```
anporumb, line 62: >thread [](http://example.com/codeflow?start=40&length=6)

would be nice if the spec would say that the thread runs "epoll_worker_func"

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+`epoll_worker_func` shall call `epoll_wait` to wait for an epoll event to become signaled.
+
+If the events value contains `EPOLLRDHUP` (hang up), `epoll_worker_func` shall the following:
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- Then call the `event_callback` callback with the `event_callback_ctx` and `COMPLETION_PORT_RESULT_ABANDONED`.
+
+If the events value contains `EPOLLIN` (recv), `epoll_worker_func` shall the following:
```
anporumb, line 152: >shall the [](http://example.com/codeflow?start=67&length=9)

shall do the? (missing "do")

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- Then `epoll_worker_func` shall call `recv` and do the following:
+
+- If the recv size < 0, then:
+
+  - If `errno` is `EAGAIN` or `EWOULDBLOCK`, then unlikely errors will continue
```
anporumb, line 160: >then unlikely errors will continue [](http://example.com/codeflow?start=45&length=34)

stack overflow

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+### epoll_worker_func
+
+```c
+static int epoll_worker_func(void* parameter)
+```
+
+If `parameter` is `NULL`, `epoll_worker_func` shall do nothing.
+
+`epoll_worker_func` shall call `epoll_wait` to wait for an epoll event to become signaled.
```
anporumb, line 144: >epoll_wait [](http://example.com/codeflow?start=32&length=10)

If this "wait" has some sort of timeout, it should be spelled here how much it is.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+Otherwise `completion_port_inc_ref` shall increment the internally maintained reference count.
+
+### completion_port_post_queue
+
+```C
+MOCKABLE_FUNCTION(, int, completion_port_post_queue, COMPLETION_PORT_HANDLE, completion_port, const COMPLETION_PORT_INGRESS_DATA*, ingress_data);
+```
+
+`completion_port_post_queue` adds an ingress item to the post queue for epoll signaling
```
anporumb, line 106: >completion_port_post_queue [](http://example.com/codeflow?start=1&length=26)

shoudl 2 calls with the same SOCKET_HANDLE succeed? Would the second one replace the parameters of the first one? What happens then?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+- If the recv size < 0, then:
+
+  - If `errno` is `EAGAIN` or `EWOULDBLOCK`, then unlikely errors will continue
+
+  - If `errno` is `ECONNRESET`, then `epoll_worker_func` shall call the `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_ABANDONED`
+
+- If the recv size equal 0, then `epoll_worker_func` shall call `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_OK`
+
+- If the recv size > 0, if we have another buffer to fill then we will attempt another read, otherwise we shall call `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_OK`
```
anporumb, line 166: >we have another buffer to fill [](http://example.com/codeflow?start=27&length=30)

How do "we" get new buffers? What's the mechanism?

anporumb, line 166: >if we have another buffer to fill [](http://example.com/codeflow?start=24&length=33)

so what happens if we "recv" and we DO NOT have another buffer? Data is dropped?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `epoll_worker_func` shall loop through the total buffers and send the data.
+
+- if send returns value is < 0 `epoll_worker_func` shall do the following:
+
+  - if `errno` is `ECONNRESET`, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ABANDONED`.
+
+  - if `errno` is anything else, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ERROR`.
+
+If the thread_access_cnt variable is not 0, `epoll_worker_func` continue, otherwise it shall exit.
```
anporumb, line 180: >`epoll_worker_func` continue [](http://example.com/codeflow?start=44&length=28)

missing "shall"?

And continue... how?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+`completion_port_inc_ref` handles the incrementing of the reference count of the module
+
+If `completion_port` is `NULL`, `completion_port_inc_ref` shall return.
+
+Otherwise `completion_port_inc_ref` shall increment the internally maintained reference count.
+
+### completion_port_post_queue
+
+```C
+MOCKABLE_FUNCTION(, int, completion_port_post_queue, COMPLETION_PORT_HANDLE, completion_port, const COMPLETION_PORT_INGRESS_DATA*, ingress_data);
```
anporumb, line 103: >queue [](http://example.com/codeflow?start=46&length=5)

I will take poetry here.

I don't think anything is "queue"d here. What it does... as `man` explains is it `add`s some file descriptor to a list of "interesting file descriptors" that are served by the epoll.

anporumb: Same with "dequeue". There's nothing "queued" so there cannot be anything "dequeueud". ueueueueueueued

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+```C
+MOCKABLE_FUNCTION(, void, completion_port_dequeue, COMPLETION_PORT_HANDLE, completion_port, SOCKET_HANDLE, socket);
+```
+
+If `completion_port` is `NULL`, `completion_port_dequeue` shall return.
+
+If `socket` is `INVALID_SOCKET`, `completion_port_dequeue` shall return.
+
+`completion_port_dequeue` shall remove the underlying socket from the epoll by calling  `epoll_ctl` with `EPOLL_CTL_DEL`.
```
anporumb, line 134: >   [](http://example.com/codeflow?start=86&length=2)

generous spaces detected.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- `epoll_worker_func` shall loop through the total buffers and send the data.
+
+- if send returns value is < 0 `epoll_worker_func` shall do the following:
+
+  - if `errno` is `ECONNRESET`, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ABANDONED`.
+
+  - if `errno` is anything else, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ERROR`.
```
anporumb, line 178: >is anything else [](http://example.com/codeflow?start=15&length=16)

I don't think this is right here.

I imagine a world where the socket file descriptor indicates availability of "EPOLLOUT" (writes).

And there's 100MB of data to be send.

And **some** of it might make it (send returns > 0). And some of it will not make it, because the socket's internal buffers are full.

socket internal buffers full maybe would return EWOULDBLOCK? Supposedly.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,180 @@
...
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- Then call the `event_callback` callback with the `event_callback_ctx` and `COMPLETION_PORT_RESULT_ABANDONED`.
+
+If the events value contains `EPOLLIN` (recv), `epoll_worker_func` shall the following:
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- Then `epoll_worker_func` shall call `recv` and do the following:
```
anporumb, line 156: >recv [](http://example.com/codeflow?start=39&length=4)

I see this module doing the `recv` and `send` - how's that different from ASYNC_SOCKET which I remember having reviewd a recent while ago...

anporumb (approved):
:shipit:

linux/devdoc/completion_port_linux_requirements.md
jebrando: I probably didn't communicate this well enough and I'll reword this to make it more understandable.

linux/devdoc/completion_port_linux_requirements.md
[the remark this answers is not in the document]
jebrando: All the data will be stored in EPOLL_THREAD_DATA.  I will update the spec to say as such.

linux/devdoc/completion_port_linux_requirements.md
jebrando: Yes, this module does support multiple threads calling into it.

jebrando: Will add a blurb.

linux/devdoc/completion_port_linux_requirements.md
jebrando: This is one thread per epoll, so it will only be one thread.

linux/devdoc/completion_port_linux_requirements.md
jebrando: This is interesting.  That would require multiple PR's and conversation.  I will look at moving it, but not in this PR.

linux/devdoc/completion_port_linux_requirements.md
jebrando: Will change.

linux/devdoc/completion_port_linux_requirements.md
jebrando: Will do.

linux/devdoc/completion_port_linux_requirements.md
jebrando: ASYNC_SOCKET will use this module so it won't have to do recv and sends anymore.

linux/devdoc/completion_port_linux_requirements.md
jebrando: It changes the event associated with the socket.  We keep a list of memory so we would not link

linux/devdoc/completion_port_linux_requirements.md
jebrando: rewording

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
+# completion_port_linux requirements
+
+## Overview
+
+The completion_port_linux module handles the threading and control of the epoll socket system
```
dcristoloveanu, line 5: >completion_port_linux [](http://example.com/codeflow?start=4&length=21)

Missing backticks

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+typedef struct EPOLL_BUFFER_DATA_TAG
+{
+    void* buffer;
+    uint32_t length;
+} EPOLL_BUFFER_DATA;
+
+typedef struct COMPLETION_PORT_INGRESS_DATA_TAG
+{
+    int epoll_op;
```
dcristoloveanu, line 37: >epoll_op [](http://example.com/codeflow?start=8&length=8)

This seems to be passed to add.
I thought that we want to bind together in a handle one epoll and one thread.
So it would mean that the epoll is passed to the create me thinks.

Adding more sockets to the "completion port" would be allowed obviously.

But I think the key thing was to not have multiple threads process events from one epoll (because order is important).

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+```C
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, completion_port_create);
+```
+
+`completion_port_create` creates and initializes the completion port module
+
+`completion_port_create` shall create a completion port object.
+
+`completion_port_create` shall create the epoll module by calling `epoll_create`.
```
dcristoloveanu, line 61: >module [](http://example.com/codeflow?start=48&length=6)

instance

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+    SOCKET_HANDLE socket;
+    ON_COMPLETION_PORT_EVENT_COMPLETE event_callback;
+    void* event_callback_ctx;
+    EPOLL_BUFFER_DATA buffer_data;
+} COMPLETION_PORT_INGRESS_DATA;
+
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, completion_port_create);
+MOCKABLE_FUNCTION(, void, completion_port_inc_ref, COMPLETION_PORT_HANDLE, completion_port);
+MOCKABLE_FUNCTION(, void, completion_port_dec_ref, COMPLETION_PORT_HANDLE, completion_port);
+MOCKABLE_FUNCTION(, int, completion_port_add, COMPLETION_PORT_HANDLE, completion_port, const COMPLETION_PORT_INGRESS_DATA*, ingress_data);
```
dcristoloveanu, line 47: >const COMPLETION_PORT_INGRESS_DATA*, ingress_data [](http://example.com/codeflow?start=87&length=49)

Can we have the parameters unfolded in the API (it makes it much harder for testing to have it as a struct.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+- if send returns value is < 0 and not `EAGAIN` or `EWOULDBLOCK`, then `epoll_worker_func` shall do the following:
+
+  - if `errno` is `ECONNRESET`, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ABANDONED`.
+
+  - if `errno` is anything else, then `on_send_complete` shall be called with `COMPLETION_PORT_RESULT_ERROR`.
+
+- Then remove the `EPOLL_THREAD_DATA` from the list and free the object
+
+If the thread_access_cnt variable is not 0, `epoll_worker_func` shall issue another `epoll_wait`, otherwise it shall exit.
```
dcristoloveanu, line 187: >thread_access_cnt [](http://example.com/codeflow?start=7&length=17)

This only appears here. It is impossible to guess what this will be and it actually looks like this is part of the internal design of the module.
This should be fully removed IMO.

linux/devdoc/completion_port_linux_requirements.md
jebrando: This is the epoll_operation value.  I will change the name so it makes more sense.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+- If the recv size equal 0, then `epoll_worker_func` shall call `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_OK`
+
+- If the recv size > 0, then if recv size < buffer length is less than `epoll_worker_func` shall issue another recv, otherwise we shall call `event_callback` callback with the `event_callback_context` with `COMPLETION_PORT_RESULT_OK`
+
+- Then remove the `EPOLL_THREAD_DATA` from the list and free the object
+
+If the events value contains `EPOLLOUT` (send), `epoll_worker_func` shall the following:
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
```
dcristoloveanu, line 175: >receive [](http://example.com/codeflow?start=28&length=7)

obtain?

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+- If the recv size < 0, then:
+
+  - If `errno` is either `EAGAIN` or `EWOULDBLOCK`, then `epoll_worker_func` shall break and call `event_callback` callback with the `event_callback_context` with `COMPLETION_PORT_RESULT_OK`
+
+  - If `errno` is `ECONNRESET`, then `epoll_worker_func` shall call the `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_ABANDONED`
+
+- If the recv size equal 0, then `epoll_worker_func` shall call `event_callback` callback with the `event_callback_context` and `COMPLETION_PORT_RESULT_OK`
+
+- If the recv size > 0, then if recv size < buffer length is less than `epoll_worker_func` shall issue another recv, otherwise we shall call `event_callback` callback with the `event_callback_context` with `COMPLETION_PORT_RESULT_OK`
```
dcristoloveanu, line 169: > > 0, then if recv size < buffer  [](http://example.com/codeflow?start=18&length=33)

I think the fact that we have 2 ifs in the same spec points to "let's break it down into at least 2 sentences

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+- Then call the `event_callback` callback with the `event_callback_ctx` and `COMPLETION_PORT_RESULT_ABANDONED`.
+
+- Then remove the `EPOLL_THREAD_DATA` from the list and free the object
+
+If the events value contains `EPOLLIN` (recv), `epoll_worker_func` shall do the following:
+
+- `epoll_worker_func` shall receive the `EPOLL_THREAD_DATA` value from the ptr variable from the `epoll_event` data ptr.
+
+- Then `epoll_worker_func` shall call `recv` and do the following:
```
dcristoloveanu, line 159: > `recv` and  [](http://example.com/codeflow?start=37&length=12)

We should spec how many bytes we want to receive

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
...
+
+- If the recv size < 0, then:
```
dcristoloveanu, line 161: >If the recv size < 0, then: [](http://example.com/codeflow?start=2&length=27)

While reading the specs I had a thought of: should this module even do the recv and send? I think it is doing too much. I am more and more thinking it should just tell whoever called that the wait is completed. And then the upper layer could deal with receiving and the complexity of queueing things for send!

dcristoloveanu: I think we should talk about it! What say you?

dcristoloveanu (changes requested):
:clock1:

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,187 @@
+# completion_port_linux requirements
+
+## Overview
+
+The completion_port_linux module handles the threading and control of the epoll socket system
+
+## Exposed API
+
+```C
```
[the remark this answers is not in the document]
jebrando: I didn't think there was a difference, and after a short bing search I didn't see anywhere it would say that I should use a lower case or an upper case.

linux/devdoc/completion_port_linux_requirements.md
jebrando: This is determined by the calling function and how big a buffer the calling function sends in.

linux/devdoc/completion_port_linux_requirements.md
jebrando: AH, I like that idea.  Let's talk that.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,134 @@
...
+
+If `completion_port` is `NULL`, `completion_port_add` shall return a non-NULL value.
+
+If `socket` is `INVALID_SOCKET`, `completion_port_add` shall return a non-NULL value.
+
+If `event_callback` is `NULL`, `completion_port_add` shall return a non-NULL value.
+
+`completion_port_add` shall allocate a `EPOLL_THREAD_DATA` object to store thread data.
+
+`completion_port_add` shall add the `EPOLL_THREAD_DATA` object to a list for later removal
```
dcristoloveanu, line 98: >l [](http://example.com/codeflow?start=89&length=1)

Needs a . at the end

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,134 @@
...
+
+`completion_port_add` adds an ingress item to the post queue for epoll signaling
+
+If `completion_port` is `NULL`, `completion_port_add` shall return a non-NULL value.
+
+If `socket` is `INVALID_SOCKET`, `completion_port_add` shall return a non-NULL value.
+
+If `event_callback` is `NULL`, `completion_port_add` shall return a non-NULL value.
+
+`completion_port_add` shall allocate a `EPOLL_THREAD_DATA` object to store thread data.
```
dcristoloveanu, line 96: >ort_add` shall allocate a `EPOLL_THREAD_DATA` object to store thread data. [](http://example.com/codeflow?start=13&length=74)

Ehm, will this trigger a memory allocation?
It would be so awesome to have a queue backed by one of these TARRAYs, wouldn't it?

dcristoloveanu: Maybe for another task?
I remember we had a similar thing for the threadpool work items for linux

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,134 @@
...
+
+If `parameter` is `NULL`, `epoll_worker_func` shall do nothing.
+
+`epoll_worker_func` shall call `epoll_wait` to wait for an epoll event to become signaled with a timeout of 2 Seconds.
+
+`epoll_worker_func` shall loop through the num of descriptors that was returned.
+
+`epoll_worker_func` shall call the `event_callback` with the specified `COMPLETION_PORT_EPOLL_ACTION` that was returned
+
+Then `epoll_worker_func` shall remove the `EPOLL_THREAD_DATA` from the list and free the object
```
dcristoloveanu, line 134: >t [](http://example.com/codeflow?start=94&length=1)

Missing . at the end of the sentence. Here and everywhere.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,134 @@
...
+
+### epoll_worker_func
+
+```c
+static int epoll_worker_func(void* parameter)
+```
+
+If `parameter` is `NULL`, `epoll_worker_func` shall do nothing.
+
+`epoll_worker_func` shall call `epoll_wait` to wait for an epoll event to become signaled with a timeout of 2 Seconds.
```
dcristoloveanu, line 128: >h a timeout of 2 Seconds. [](http://example.com/codeflow?start=93&length=25)

So what happens on timeout? It's a little bit underspecced

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,134 @@
...
+- close the epoll object.
+
+- close the thread by calling `ThreadAPI_Join`
+
+- then the memory associated with `completion_port` shall be freed.
+
+### completion_port_add
+
+```C
+MOCKABLE_FUNCTION(, int, completion_port_add, COMPLETION_PORT_HANDLE, completion_port, int, epoll_op, SOCKET_HANDLE, socket,
```
dcristoloveanu, line 84: >completion_port_add [](http://example.com/codeflow?start=25&length=19)

Can completion_port_add be called while dec_ref is doing the destruction of the completion port?

linux/devdoc/completion_port_linux_requirements.md
jebrando: I can create a task for this.

linux/devdoc/completion_port_linux_requirements.md
jebrando: I think I'll need to add a pending call count for this.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -0,0 +1,144 @@
...
+
+```c
+static int epoll_worker_func(void* parameter)
+```
+
+If `parameter` is `NULL`, `epoll_worker_func` shall do nothing.
+
+`epoll_worker_func` shall call `epoll_wait` to wait for an epoll event to become signaled with a timeout of 2 Seconds.
+
+On a `epoll_wait` timeout `epoll_worker_func` shall ensure it should not exit and issue another `epoll_wait`
```
dcristoloveanu, line 138: >t` [](http://example.com/codeflow?start=106&length=2)

Missing . at the end.

dcristoloveanu (approved):
:shipit:

PR #200, opened 2023-01-10
feature: adding completion port implementation

jebrando:


linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -66,19 +66,21 @@ MOCKABLE_FUNCTION(, void, completion_port_dec_ref, COMPLETION_PORT_HANDLE, compl
...
+**COMPLETION_PORT_LINUX_11_008: [** If `completion_port` is `NULL`, `completion_port_dec_ref` shall return. **]**
 
-If the reference count reaches 0, `completion_port_dec_ref` shall do the following:
+**COMPLETION_PORT_LINUX_11_009: [** `completion_port_dec_ref` shall decrement the reference count for `completion_port`. **]**
 
-- wait for the ongoing call count to reach zero.
+**COMPLETION_PORT_LINUX_11_010: [** If the reference count reaches 0, `completion_port_dec_ref` shall do the following: **]**
 
-- increment the flag signaling that the threads can complete.
+**COMPLETION_PORT_LINUX_11_011: [** - wait for the ongoing call count to reach zero. **]**
```
mattdurak, line 75: >-  [](http://example.com/codeflow?start=36&length=2)

nit: can we make the markdown look nice and reformat all of these as:

` - **SRS[** text **]**`

Instead of having the "-" inside

If you use the SRS extension ALT-F9 over multiple lines (instead of ALT-F8 one at a time) it will preserve those markdown lists

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -89,56 +91,60 @@ MOCKABLE_FUNCTION(, int, completion_port_add, COMPLETION_PORT_HANDLE, completion
 
 `completion_port_add` adds an ingress item to the post queue for epoll signaling.
 
-If `completion_port` is `NULL`, `completion_port_add` shall return a non-NULL value.
+**COMPLETION_PORT_LINUX_11_016: [** If `completion_port` is `NULL`, `completion_port_add` shall return a non-NULL value. **]**
```
mattdurak, line 94: >NULL [](http://example.com/codeflow?start=109&length=4)

non-zero here and everywhere

linux/linux_reals/real_completion_port_linux.h
```diff
@@ -0,0 +1,33 @@
...
+        completion_port_dec_ref,                            \
+        completion_port_add,                                \
+        completion_port_remove                              \
+    )
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    COMPLETION_PORT_HANDLE completion_port_create(void);
```
mattdurak, line 23: >completion_port_create [](http://example.com/codeflow?start=27&length=22)

these should all have the "real_" prefix

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+        if (current_val == 0)
+        {
+            break;
+        }
+        else
+        {
+            // Do Nothing wait for address
+        }
+        (void)wait_on_address(&completion_port->recv_data_access, current_val, UINT32_MAX);
+    } while (true);
```
mattdurak, line 68: This kind of looks like some sort of "acquire lock" or "enter critical section" function, and at the end those 2 lines seem like a "release lock" or "leave critical section".

At the very least I'd consider making a pair of static functions named as such (if not a full blown module), especially since it's repeated below.

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+                }
+                #ifdef USE_VALGRIND
+                    ANNOTATE_HAPPENS_AFTER(epoll_data);
+                #endif
+                // Codes_COMPLETION_PORT_LINUX_11_035: [ epoll_worker_func shall call the event_callback with the specified COMPLETION_PORT_EPOLL_ACTION that was returned. ]
+                epoll_data->event_callback(epoll_data->event_callback_ctx, epoll_action);
+                // Codes_COMPLETION_PORT_LINUX_11_036: [ Then epoll_worker_func shall remove the EPOLL_THREAD_DATA from the list and free the object. ]
+                if (remove_thread_data_from_list(completion_port, epoll_data) != 0)
+                {
+                    LogWarning("failure removing receive data to list %p, action: %" PRI_MU_ENUM "", epoll_data, MU_ENUM_VALUE(COMPLETION_PORT_EPOLL_ACTION, epoll_action));
```
mattdurak, line 171: >to [](http://example.com/codeflow?start=62&length=2)

nit: from

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+        // Codes_COMPLETION_PORT_LINUX_11_009: [ completion_port_dec_ref shall decrement the reference count for completion_port. ]
+        if (DEC_REF(COMPLETION_PORT, completion_port) == 0)
+        {
+            // Codes_COMPLETION_PORT_LINUX_11_010: [ If the reference count reaches 0, completion_port_dec_ref shall do the following: ]
+            int32_t value;
+            // Codes_COMPLETION_PORT_LINUX_11_011: [ - wait for the ongoing call count to reach zero. ]
+            while ((value = interlocked_add(&completion_port->pending_calls, 0)) != 0)
+            {
+                (void)wait_on_address(&completion_port->pending_calls, value, UINT32_MAX);
+            }
```
mattdurak, line 264: from what I see this is incremented and decremented in `completion_port_add`.

But there is no state change that happens before this check (like going to a "CLOSING" state).

Which means that someone could call `completion_port_add` after line 264 and it would work.

But this is a dec_ref so basically destroy, which means we are making this check during destroy.

As Andrei would say, destroy must be single threaded so no other thread should be accessing this object while it is being destroyed, so it seems like we should never execute line 263 (wait_on_address) because we can never have pending calls.

I think this check makes sense if this module has open/close. And if it does then it needs to do what we always did before sm - first set a state to close then wait for all calls to complete (and each call needs to first increment the pending count then check the state).

But as it is, I don't think it makes sense to have this pending call count at all. Otherwise I'm curious to see the test that would actually hit this `wait_on_address` line without artificially mocking `interlocked_add`.

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+                if (current_val == 0)
+                {
+                    break;
+                }
+                else
+                {
+                    // Do Nothing
+                }
+                (void)wait_on_address(&completion_port->recv_data_access, current_val, UINT32_MAX);
+            } while (true);
```
mattdurak, line 295: this seems to be "acquiring" the "lock" during destroy after waiting on all the threads to exit. Seems like it shouldn't be necessary (how is another thread accessing this?)

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+        LogError("Invalid arguments: COMPLETION_PORT_HANDLE completion_port=%p, epoll_op=%d, SOCKET_HANDLE socket=%" PRI_SOCKET ", ON_COMPLETION_PORT_EVENT_COMPLETE event_callback=%p, void* event_callback_ctx=%p",
+            completion_port, epoll_op, socket, event_callback, event_callback_ctx);
+        result = MU_FAILURE;
+    }
+    else
+    {
+        // Codes_COMPLETION_PORT_LINUX_11_019: [ completion_port_add shall ensure the thread completion flag is not set. ]
+        if (interlocked_add(&completion_port->worker_thread_continue, 0) != 0)
+        {
+            LogError("Invalid call sequence completion port module is closing");
```
mattdurak, line 335: > module is closing"); [](http://example.com/codeflow?start=59&length=21)

it looks like it's actually being destroyed if this happens...

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+                    }
+                    // Remove
+                    remove_thread_data_from_list(completion_port, epoll_thread_data);
+                }
+                free(epoll_thread_data);
+            }
+            // Codes_COMPLETION_PORT_LINUX_11_027: [ If any error occurs, completion_port_add shall fail and return a non-zero value. ]
+            result = MU_FAILURE;
+            (void)interlocked_decrement(&completion_port->pending_calls);
+            wake_by_address_single(&completion_port->pending_calls);
```
mattdurak, line 409: if you move "all_ok:" up here then you don't need to repeat these 2 lines 3 times.

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+    if (
+        // Codes_COMPLETION_PORT_LINUX_11_028: [ If completion_port is NULL, completion_port_remove shall return. ]
+        completion_port == NULL || 
+        // Codes_COMPLETION_PORT_LINUX_11_029: [ If socket is INVALID_SOCKET, completion_port_remove shall return. ]
+        socket == INVALID_SOCKET)
+    {
+        LogError("Invalid arguments: COMPLETION_PORT_HANDLE completion_port=%p, SOCKET_HANDLE socket=%" PRI_SOCKET "", completion_port, socket);
+    }
+    else
+    {
```
mattdurak, line 427: No `pending_calls`? Of course with my other comment I'm wondering if it's needed at all

mattdurak (approved):
:shipit:

linux/src/completion_port_linux.c
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    epoll_action = COMPLETION_PORT_EPOLL_EPOLLIN;
+                }
+                else if (events[index].events & EPOLLOUT)
+                {
+                    epoll_action = COMPLETION_PORT_EPOLL_EPOLLOUT;
+                }
+                else
+                {
+                    continue;
```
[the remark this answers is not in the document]
jebrando: We might get something, other than this, but it shouldn't be often, so a log might be OK.

linux/src/completion_port_linux.c
jebrando: I believe I need to add pending calls here.

linux/src/completion_port_linux.c
jebrando: I need to move the worker_thread_continue increment up before line 260.  This would ensure that the add and remove won't race.

If I had two threads, one calling add and another calling dec_ref then you would need the pending call count (Dan brought this one up, so I can't take credit).  I will rearrange this a little submit.

linux/devdoc/completion_port_linux_requirements.md
```diff
@@ -66,19 +66,21 @@ MOCKABLE_FUNCTION(, void, completion_port_dec_ref, COMPLETION_PORT_HANDLE, compl
...
+**SRS_COMPLETION_PORT_LINUX_11_008: [** If `completion_port` is `NULL`, `completion_port_dec_ref` shall return. **]**
 
-If the reference count reaches 0, `completion_port_dec_ref` shall do the following:
+**SRS_COMPLETION_PORT_LINUX_11_009: [** `completion_port_dec_ref` shall decrement the reference count for `completion_port`. **]**
 
-- wait for the ongoing call count to reach zero.
+**SRS_COMPLETION_PORT_LINUX_11_010: [** If the reference count reaches 0, `completion_port_dec_ref` shall do the following: **]**
 
-- increment the flag signaling that the threads can complete.
+**SRS_COMPLETION_PORT_LINUX_11_011: [** - wait for the ongoing call count to reach zero. **]**
```
mattdurak, line 75: not a big deal but my comment before was more about changing from this:

```
**SRS_COMPLETION_PORT_LINUX_11_011: [** - wait for the ongoing call count to reach zero. **]**
```

to this:

```
- **SRS_COMPLETION_PORT_LINUX_11_011: [** wait for the ongoing call count to reach zero. **]**
```

Because the markdown lists work then and the "-" doesn't need to go in the code.

jebrando: Oh, gotcha.  I will change it.

linux/src/completion_port_linux.c
mattdurak: I guess my question is, how do you have this?

Reference count =1

T1: completion_port_add
T2: completion_port_dec_ref

I would argue that either T2 cannot call dec ref because it knows about the add on T1 (it holds the 1 and only reference afterall)

OR

T1 cannot call add because it should have had an inc_ref at some time (ref count should have been 2 so T2 would just decrement).

In other words, you can't call dec_ref until you know that you are no longer using (your reference to) the object and conversely you can't call anything on the object unless you know you have a reference.

PR #202, opened 2023-01-11
feature: adding completion port to async socket

jebrando:


linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
+# platform_linux
+
+## Overview
+
+`platform_linux` provides initializes the completion_port system.
```
jebrando, line 5: >system [](http://example.com/codeflow?start=58&length=6)

object.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -77,52 +77,16 @@ MOCKABLE_FUNCTION(, ASYNC_SOCKET_HANDLE, async_socket_create, EXECUTION_ENGINE_H
...
-**SRS_ASYNC_SOCKET_LINUX_11_002: [** If `execution_engine` is NULL, `async_socket_create` shall fail and return `NULL`. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_002: [** `execution_engine` shall be allowed to be `NULL`. **]**
 
 **SRS_ASYNC_SOCKET_LINUX_11_003: [** If `socket_handle` is `INVALID_SOCKET`, `async_socket_create` shall fail and return `NULL`. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_004: [** `async_socket_create` shall increment the reference count on `execution_engine`. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_005: [** `async_socket_create` shall retreive an `COMPLETION_PORT_HANDLE` object by calling `platform_get_completion_port`. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_005: [** `async_socket_create` shall initialize the global thread. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_101: [** `async_socket_create` shall increment the reference count of the `COMPLETION_PORT_HANDLE` object by calling completion_port_inc_ref. **]**
```
anporumb, line 86: These 2 requirements are contradictory with themselves.

If these `COMPLETION_PORT_HANDLE` are in any way ref-counted, and the factory/getter for them seems to be "platform_get_completion_port" then there's no need to inc_ref them, the factory will return them already inc_ref'd.

Wondering how the tests didn't leak...

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -77,52 +77,16 @@ MOCKABLE_FUNCTION(, ASYNC_SOCKET_HANDLE, async_socket_create, EXECUTION_ENGINE_H
 
 **SRS_ASYNC_SOCKET_LINUX_11_001: [** `async_socket_create` shall allocate a new async socket and on success shall return a non-`NULL` handle. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_002: [** If `execution_engine` is NULL, `async_socket_create` shall fail and return `NULL`. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_002: [** `execution_engine` shall be allowed to be `NULL`. **]**
 
 **SRS_ASYNC_SOCKET_LINUX_11_003: [** If `socket_handle` is `INVALID_SOCKET`, `async_socket_create` shall fail and return `NULL`. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_004: [** `async_socket_create` shall increment the reference count on `execution_engine`. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_005: [** `async_socket_create` shall retreive an `COMPLETION_PORT_HANDLE` object by calling `platform_get_completion_port`. **]**
```
anporumb, line 84: retrieve

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_020: [** While `async_socket` is `OPENING` or `CLOSING`, `async_socket_destroy` shall wait for the open to complete either successfully or with error. **]**

open/close?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:100 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
+**SRS_ASYNC_SOCKET_LINUX_11_079: [** If context is `NULL`, `event_complete_callback` shall do nothing. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_080: [** Onced signaled `thread_worker_func` shall loop through the signaled epolls. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_080: [** If `COMPLETION_PORT_EPOLL_ACTION` is `COMPLETION_PORT_EPOLL_EPOLLRDHUP` or `COMPLETION_PORT_EPOLL_ABANDONED`, `event_complete_callback` shall do the following: **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_081: [** If the events value contains `EPOLLRDHUP` (hang up), `thread_worker_func` shall the following: **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_081: [** `event_complete_callback` shall call either the send or recv complete callback with an `ABANDONED` flag. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_082: [** `thread_worker_func` shall receive the `ASYNC_SOCKET_RECV_CONTEXT` value from the ptr variable from the `epoll_event` data ptr. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_084: [** Then `event_complete_callback` shall and free the `io_context` memory. **]**
```
anporumb, line 254: >shall [](http://example.com/codeflow?start=70&length=5)

shall what?

anporumb, line 254: >io_context [](http://example.com/codeflow?start=90&length=10)

The only reference to this `io_context` is right in this function (`event_complete_callback`). Would be nice to know what's the `io_context` supposed to contain/how it is used /etc

anporumb:
>     ASYNC_SOCKET_OPEN_ERROR

would an OPEN_ABANDONED be conceivable here?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:26 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
>     ASYNC_SOCKET_SEND_SYNC_ABANDONED

Why is `ABANDONED` a possible result of a `SYNC` action?

I was under the impression that `ABANDONED` always is a callback result, that is, it comes on the ASYNC path.

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:33 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_052: [** `async_socket_send_async` shall attempt to send the data by calling `send` with the `MSG_NOSIGNAL` flag to ensure an exception is not generated. **]**

I think the better term here (reading man send) is "signal". 

Exceptions are IMHO something that happens in C++ :)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:184 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> ## Design

Missing stating what's the threading model here.

This is rather important. Waiting.

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:7 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -231,9 +191,7 @@ MOCKABLE_FUNCTION(, ASYNC_SOCKET_SEND_SYNC_RESULT, async_socket_send_async, ASYN
 
   - **SRS_ASYNC_SOCKET_LINUX_11_056: [** `async_socket_send_async` shall create a context for the send where the `payload`, `on_send_complete` and `on_send_complete_context` shall be stored. **]**
```
anporumb, line 192: >shall create a context for the send where the `payload`, `on_send_complete` and `on_send_complete_context` shall be stored. [](http://example.com/codeflow?start=67&length=123)

So my understanding is that this "context" will be somehow queued to the completion port's notifications.

What happens when the completion port says "I am ready for more sends" and AT THE SAME TIME `async_socket_send_async` is called from another thread? 

Which wins?

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -231,9 +191,7 @@ MOCKABLE_FUNCTION(, ASYNC_SOCKET_SEND_SYNC_RESULT, async_socket_send_async, ASYN
...
 
-  - **SRS_ASYNC_SOCKET_LINUX_11_057: [** The the context shall then be added to the epoll system by calling `epoll_ctl` with `EPOLL_CTL_MOD`. **]**
-
-  - **SRS_ASYNC_SOCKET_LINUX_11_058: [** If the `epoll_ctl` call fails with `ENOENT`, `async_socket_send_async` shall call `epoll_ctl` again with `EPOLL_CTL_ADD`. **]**
+  - **SRS_ASYNC_SOCKET_LINUX_11_057: [** The context shall then be added to the completion port system by calling `completion_port_add` with `EPOLL_CTL_MOD`. **]**
 
 - **SRS_ASYNC_SOCKET_LINUX_11_059: [** If the `errno` value is `ECONNRESET`, `ENOTCONN`, or `EPIPE` shall fail and return `ASYNC_SOCKET_SEND_SYNC_ABANDONED`. **]**
```
anporumb, line 196: >ASYNC_SOCKET_SEND_SYNC_ABANDONED [](http://example.com/codeflow?start=123&length=32)

I don't think we should ever return ABADONED from any synchronous path.

linux/devdoc/async_socket_linux_requirements.md
anporumb, line 192: >create [](http://example.com/codeflow?start=73&length=6)

so let's assume there are 3 `buffer_count`. Let's assume the 1 buffer will be send synchronously (great joy!).

Let's assume the second buffer gets a "EWOULDBLOCK" so it is queued for notifications.

Let's assume the 3 buffer fails in this "create" (any create can fail).

So this leave us with:

1 packet "on the wire"
1 payload queued for sending whenever the completion port so pleases to tell us
1 payload failed (because malloc fails for it).

As per 11_063 the function will return "ERROR".

What's the mitigation at user level? This is a very delicate problem which frankly I don't see how it can be tackled with the current function prototype.

anporumb: That's a reason to wait, because curiosity is eating me.

anporumb:
> MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_socket, ASYNC_SOCKET_BUFFER*, payload, uint32_t, buffer_count, ON_SOCKET_ASYNC_RECEIVE_COMPLETE, on_receive_complete, void*, on_receive_complete_context);

404 - what is this type?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:209 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_073: [** Otherwise `async_socket_receive_async` shall create a context for the send where the `payload`, `on_receive_complete` and `on_receive_complete_context` shall be stored. **]**

aren't we in a "receive" context at this point?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:232 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_073: [** Otherwise `async_socket_receive_async` shall create a context for the send where the `payload`, `on_receive_complete` and `on_receive_complete_context` shall be stored. **]**

Is this intended to be singular, or plural? Is it one context / ASYNC_SOCKET_BUFFER or it is 1 context / ALL ASYNC_SOCKET_BUFFERS?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:232 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_073: [** Otherwise `async_socket_receive_async` shall create a context for the send where the `payload`, `on_receive_complete` and `on_receive_complete_context` shall be stored. **]**

11_907 seems to indicate that the context encompasses all ASYNC_SOCKET_BUFFER

---
In reply to: [1380904966](https://github.com/Azure/c-pal/pull/202#issuecomment-1380904966) [](http://example.com/codeflow?ancestors=1380904966)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:232 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_073: [** Otherwise `async_socket_receive_async` shall create a context for the send where the `payload`, `on_receive_complete` and `on_receive_complete_context` shall be stored. **]**

The fate of this `context` and what happens to it after it is created is left as an imagination exercise to the user.

I fail to see how it related to the rest of the objects created in here.

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:232 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
 **SRS_ASYNC_SOCKET_LINUX_11_077: [** On success, `async_socket_receive_async` shall return 0. **]**
 
 **SRS_ASYNC_SOCKET_LINUX_11_078: [** If any error occurs, `async_socket_receive_async` shall fail and return a non-zero value. **]**
 
-### thread_worker_func
+### event_complete_callback
 
 ```c
-static int thread_worker_func(void* parameter)
+static void event_complete_callback(void* context, COMPLETION_PORT_EPOLL_ACTION action)
```
anporumb, line 243: >event_complete_callback [](http://example.com/codeflow?start=12&length=23)

404 not found where this callback is passed to any other APIs.

anporumb: I will have to assume that this is whatever the epoll is actually calling.

anporumb, line 243: >event_complete_callback [](http://example.com/codeflow?start=12&length=23)

I do have to ask annoying questions here... but...

for the same epoll, can these event_complete_callback originate on different threads?

In the case when a connection is severed, will be events queued in the epoll be served:
1) in the order that they have been queued
2) in the reverse order they have been queued.
3) in a random order.

This might be important, and should be stated if these callbacks come from different threads to the user land or 1 unique thread and the order (if any). Users might care.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
+- **SRS_ASYNC_SOCKET_LINUX_11_081: [** `event_complete_callback` shall call either the send or recv complete callback with an `ABANDONED` flag. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_082: [** `thread_worker_func` shall receive the `ASYNC_SOCKET_RECV_CONTEXT` value from the ptr variable from the `epoll_event` data ptr. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_084: [** Then `event_complete_callback` shall and free the `io_context` memory. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_083: [** The `ASYNC_SOCKET_RECV_CONTEXT` object shall be removed from list of stored pointers. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_082: [** If `COMPLETION_PORT_EPOLL_ACTION` is `COMPLETION_PORT_EPOLL_EPOLLRDHUP`, `event_complete_callback` shall do the following: **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_084: [** Then call the `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_ABANDONED`. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_083: [** `event_complete_callback` shall call `recv` and do the following: **]**
```
anporumb, line 258: >SRS_ASYNC_SOCKET_LINUX_11_083 [](http://example.com/codeflow?start=4&length=29)

SRS_ASYNC_SOCKET_LINUX_11_083 has the same alignment as SRS_ASYNC_SOCKET_LINUX_11_081 and I don't think that is intended.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
+**SRS_ASYNC_SOCKET_LINUX_11_080: [** If `COMPLETION_PORT_EPOLL_ACTION` is `COMPLETION_PORT_EPOLL_EPOLLRDHUP` or `COMPLETION_PORT_EPOLL_ABANDONED`, `event_complete_callback` shall do the following: **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_081: [** If the events value contains `EPOLLRDHUP` (hang up), `thread_worker_func` shall the following: **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_081: [** `event_complete_callback` shall call either the send or recv complete callback with an `ABANDONED` flag. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_082: [** `thread_worker_func` shall receive the `ASYNC_SOCKET_RECV_CONTEXT` value from the ptr variable from the `epoll_event` data ptr. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_084: [** Then `event_complete_callback` shall and free the `io_context` memory. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_083: [** The `ASYNC_SOCKET_RECV_CONTEXT` object shall be removed from list of stored pointers. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_082: [** If `COMPLETION_PORT_EPOLL_ACTION` is `COMPLETION_PORT_EPOLL_EPOLLRDHUP`, `event_complete_callback` shall do the following: **]**
```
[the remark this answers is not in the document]
anporumb: I think this is sort of "OK/available" return code actually.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
 
-- **SRS_ASYNC_SOCKET_LINUX_11_084: [** Then call the `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_ABANDONED`. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_083: [** `event_complete_callback` shall call `recv` and do the following: **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_085: [** If the events value contains `EPOLLIN` (recv), `thread_worker_func` shall the following: **]**
+  - **SRS_ASYNC_SOCKET_LINUX_11_088: [** If the recv size < 0, then: **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_086: [** `thread_worker_func` shall receive the `ASYNC_SOCKET_RECV_CONTEXT` value from the ptr variable from the `epoll_event` data ptr. **]**
+    - **SRS_ASYNC_SOCKET_LINUX_11_089: [** If `errno` is `EAGAIN` or `EWOULDBLOCK`, then no data is available and `event_complete_callback` will break out of the function. **]**
```
anporumb, line 262: >will break out of the function [](http://example.com/codeflow?start=140&length=30)

"and angrily log that the epoll is lying about availability of the function call" :)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
 
-- **SRS_ASYNC_SOCKET_LINUX_11_087: [** Then `thread_worker_func` shall call `recv` and do the following: **]**
+    - **SRS_ASYNC_SOCKET_LINUX_11_090: [** If `errno` is `ECONNRESET`, then `event_complete_callback` shall call the `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_ABANDONED`. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_088: [** If the recv size < 0, then: **]**
+    - **SRS_ASYNC_SOCKET_LINUX_11_095: [** If `errno` is any other error, then `event_complete_callback` shall call the `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_ERROR`. **]**
 
-  - **SRS_ASYNC_SOCKET_LINUX_11_089: [** If `errno` is `EAGAIN` or `EWOULDBLOCK`, then unlikely errors will continue **]**
+  - **SRS_ASYNC_SOCKET_LINUX_11_091: [** If the recv size equal 0, then `event_complete_callback` shall call `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_OK`. **]**
```
anporumb, line 268: >If the recv size equal 0 [](http://example.com/codeflow?start=41&length=24)

errr... documentation for revc says:

`When a stream socket peer has performed an orderly shutdown, the return value will be 0 (the traditional "end-
       of-file" return).`

so why would we tell to the user some "OK" when the connection is not there anymore? There are NO bytes received...

anporumb, line 268: >size equal [](http://example.com/codeflow?start=53&length=10)

"equals" or "is equal to"

linux/devdoc/async_socket_linux_requirements.md
anporumb, line 258: >recv [](http://example.com/codeflow?start=77&length=4)

it should state the parameters which recv is called with.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
 ```
 
-`thread_worker_func` handles the epoll thread.
+`event_complete_callback` handles all the completion port callback info.
 
-**SRS_ASYNC_SOCKET_LINUX_11_079: [** `thread_worker_func` shall call `epoll_wait` waiting for the epoll to become signaled. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_079: [** If context is `NULL`, `event_complete_callback` shall do nothing. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_080: [** Onced signaled `thread_worker_func` shall loop through the signaled epolls. **]**
+**SRS_ASYNC_SOCKET_LINUX_11_080: [** If `COMPLETION_PORT_EPOLL_ACTION` is `COMPLETION_PORT_EPOLL_EPOLLRDHUP` or `COMPLETION_PORT_EPOLL_ABANDONED`, `event_complete_callback` shall do the following: **]**
```
anporumb, line 250: >COMPLETION_PORT_EPOLL_EPOLLRDHUP [](http://example.com/codeflow?start=75&length=32)

If this value is ever encountered... what does that say about the actual state of the socket?

And... if it says "disconnected" then should we accept all the future send/receive asyncs? They will fail.

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -275,62 +233,58 @@ MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_
...
+- **SRS_ASYNC_SOCKET_LINUX_11_093: [** `event_complete_callback` shall then free the `io_context` memory. **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_092: [** If the recv size > 0, if we have another buffer to fill then we will attempt another read, otherwise we shall call `on_receive_complete` callback with the `on_receive_complete_context` and `ASYNC_SOCKET_RECEIVE_OK` **]**
+**SRS_ASYNC_SOCKET_LINUX_11_094: [** If the events value contains `COMPLETION_PORT_EPOLL_EPOLLOUT`, `event_complete_callback` shall the following: **]**
 
-- **SRS_ASYNC_SOCKET_LINUX_11_093: [** The `ASYNC_SOCKET_RECV_CONTEXT` object shall be removed from list of stored pointers. **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_096: [** `event_complete_callback` shall call `send` on the data in the `ASYNC_SOCKET_SEND_CONTEXT` buffer. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_094: [** If the events value contains `EPOLLOUT` (send), `thread_worker_func` shall the following: **]**
+- **SRS_ASYNC_SOCKET_LINUX_11_097: [** If send returns value is < 0 `event_complete_callback` shall do the following: **]**
```
anporumb, line 278: >send [](http://example.com/codeflow?start=42&length=4)

strictly talking about "send" it looks like 

`RETURN VALUE
       On success, these calls return the number of bytes sent.  On error, -1 is returned, and errno is set appropri‐
       ately.`

So what happens if we try to "send" 1MB of data, and send returns 100KB... what happens to the rest of the data??

... this is a reason to wait because we will absolutely have buffers big enough that we cannot send in 1 shot, but would otherwise be "send"-able in multiple "send"s.

linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
...
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, platform_get_completion_port);
+```
+
+### platform_init
+
+```c
+MOCKABLE_FUNCTION(, int, platform_init);
+```
+
+`platform_init` initializes the completion port object if it's not initialized
```
anporumb, line 21: >the completion port object [](http://example.com/codeflow?start=28&length=26)

So there can only ever be just 1 completion port?

linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+**SRS_PLATFORM_LINUX_11_001: [** If the completion port object is NULL, `platform_init` shall call `completion_port_create`. **]**
+
+**SRS_PLATFORM_LINUX_11_002: [** If `completion_port_create` returns a valid completion port object, `platform_init` shall return zero. **]**
+
+**SRS_PLATFORM_LINUX_11_007: [** If the completion port object is non-NULL, `platform_init` shall return zero. **]**
```
anporumb, line 27: >completion port object is non-NULL [](http://example.com/codeflow?start=40&length=34)

how is this different from `valid completion port object`  in 

`SRS_PLATFORM_LINUX_11_002`.

linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+### platform_deinit
+
+```c
+MOCKABLE_FUNCTION(, void, platform_deinit);
+```
+
+`platform_deinit` .
+
+**SRS_PLATFORM_LINUX_11_004: [** If the completion port object is non-NULL, `platform_deinit` shall decrement the reference by calling `completion_port_dec_ref`. **]**
```
anporumb, line 39: >the reference [](http://example.com/codeflow?start=110&length=13)

whose reference.. (I think).

linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+**SRS_PLATFORM_LINUX_11_008: [** If the completion port object is non-NULL, `platform_deinit` shall do nothing. **]**
```
anporumb, line 41: >If the completion port object is non-NULL, `platform_deinit` shall do nothing. [](http://example.com/codeflow?start=33&length=78)

as opposed to line 39 which says `If the completion port object is non-NULL, `platform_deinit` shall decrement...`.

So it either decrements or does nothing?

linux/devdoc/platform_linux_requirements.md
```diff
@@ -0,0 +1,51 @@
...
+
+### platform_get_completion_port
+
+```c
+MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, platform_get_completion_port);
+```
+
+`platform_get_completion_port` returns the completion port object reference.
+
+**SRS_PLATFORM_LINUX_11_006: [** `platform_get_completion_port` shall return the completion object. **]**
```
anporumb, line 51: >shall return the completion object. [](http://example.com/codeflow?start=64&length=35)

One does not simply return ref counted objects 


... without increasing the reference count BEFORE returning them.

anporumb: Gotta wonder how the tests didn't catch that.

linux/inc/c_pal/platform_linux.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef PLATFORM_LINUX_H
+#define PLATFORM_LINUX_H
+
+#include "umock_c/umock_c_prod.h"
```
anporumb, line 7: move near `#ifdef __cplusplus`

linux/inc/c_pal/platform_linux.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "c_pal/completion_port_linux.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+    MOCKABLE_FUNCTION(, COMPLETION_PORT_HANDLE, platform_get_completion_port);
```
anporumb, line 15: >platform_get_completion_port [](http://example.com/codeflow?start=48&length=28)

what happened to platform_init/deinit which are listed in the .md?

linux/src/platform_linux.c
```diff
@@ -1,13 +1,53 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#include "c_logging/xlogging.h"
+
 #include "c_pal/platform.h"
+#include "c_pal/completion_port_linux.h"
+
+COMPLETION_PORT_HANDLE g_completion_port = NULL;
```
anporumb, line 9: >g_completion_port [](http://example.com/codeflow?start=23&length=17)

Is this intended to actually be static?

linux/src/platform_linux.c
```diff
@@ -1,13 +1,53 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#include "c_logging/xlogging.h"
+
 #include "c_pal/platform.h"
```
anporumb, line 6: >"c_pal/platform.h" [](http://example.com/codeflow?start=9&length=18)

Is this intended to be "platform_linux.h"? Seeing as that's where `platform_get_completion_port` is declared.

linux/devdoc/async_socket_linux_requirements.md
jebrando: Good point, if the connection is severed the events will be serviced in the order they are queued.  I'll add that to the doc as a note.

linux/src/platform_linux.c
```diff
@@ -1,13 +1,53 @@
...
 }
 
 void platform_deinit(void)
 {
+    // Codes_SRS_PLATFORM_LINUX_11_008: [ If the completion port object is non-NULL, platform_deinit shall do nothing. ]
+    if (g_completion_port != NULL)
+    {
+        // Codes_SRS_PLATFORM_LINUX_11_004: [ If the completion port object is non-NULL, platform_deinit shall decrement the reference by calling completion_port_dec_ref. ]
+        completion_port_dec_ref(g_completion_port);
+        g_completion_port = NULL;
```
anporumb, line 45: >g_completion_port = NULL; [](http://example.com/codeflow?start=8&length=25)

wrong always.

mattdurak:
>     ASYNC_SOCKET_SEND_SYNC_ABANDONED

Likely should be called `NOT_OPEN` in this case.

I think way back when we had `BECAUSE_CLOSE`, it was used when we encountered a non-open state in an operation and called the callback. It was also used when we had some operation which was "cancelled" because we are closing the service.

Then we just renamed `BECAUSE_CLOSE` to `ABANDONED` everywhere, though the meaning wasn't really correct.

And now that we've been cleaning up `ABANDONED` to mean - this operation started and then was somehow cancelled and may or may not have completed, we've been adding back a `NOT_OPEN` result as needed in places where it's appropriate.

In general, a sync path return is `NOT_OPEN` because no changes have been made by the module. The callbacks should generally return `ABANDONED` because some operation may have started (e.g. reached the server) but then the operation is cancelled (e.g. we break the network connection but don't know what happened on the server) so the result is indeterminate (e.g. a write may or may not have happened).

---
In reply to: [1380882279](https://github.com/Azure/c-pal/pull/202#issuecomment-1380882279) [](http://example.com/codeflow?ancestors=1380882279,1380882279)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:33 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
jebrando: Yes, it shall because you are correct, if we get this then epoll was lying.

mattdurak:
>     ASYNC_SOCKET_OPEN_ERROR

I would hope we can't get into an indeterminate state on open complete. Either we successfully open or we fail and we are back in a "closed" state. What would OPEN_ABANDONED mean, could there be some connected socket sticking around?

---
In reply to: [1380880779](https://github.com/Azure/c-pal/pull/202#issuecomment-1380880779) [](http://example.com/codeflow?ancestors=1380880779,1380880779)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:26 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> **SRS_ASYNC_SOCKET_LINUX_11_030: [** If `async_socket` has already closed the underlying socket handle then `async_socket_open_async` shall fail and return a non-zero value. **]**

what happens in the following sequence of calls:

create/open/close/open/close?

Looks like async_socket_close "shall call `close` on the underlying socket.". 

... so this module only has 1x open and 1x close?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:128 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
jebrando: Once we get this it does say that the socket is disconnected, but other sockets may still be connected.

mattdurak:
> **SRS_ASYNC_SOCKET_LINUX_11_030: [** If `async_socket` has already closed the underlying socket handle then `async_socket_open_async` shall fail and return a non-zero value. **]**

I think this is how it works for Windows too and it needs to work that way because the socket is created elsewhere (e.g. on the server it's the listener accepting a connection). You can't re-open a socket when you didn't initiate the connection.

---
In reply to: [1380937487](https://github.com/Azure/c-pal/pull/202#issuecomment-1380937487) [](http://example.com/codeflow?ancestors=1380937487,1380937487)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:128 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -185,12 +149,8 @@ MOCKABLE_FUNCTION(, void, async_socket_close, ASYNC_SOCKET_HANDLE, async_socket)
 
 **SRS_ASYNC_SOCKET_LINUX_11_037: [** `async_socket_close` shall wait for all executing `async_socket_send_async` and `async_socket_receive_async` APIs. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_038: [** Then `async_socket_close` shall remove the underlying socket form the epoll by calling  `epoll_ctl` with `EPOLL_CTL_DEL`. **]**
-
 **SRS_ASYNC_SOCKET_LINUX_11_039: [** `async_socket_close` shall call `close` on the underlying socket. **]**
```
anporumb, line 152: >async_socket_close` shall call `close` on the underlying socket. [](http://example.com/codeflow?start=38&length=64)

So does then `async_socket_open_async` call open on the underlying socket?

anporumb:
> MOCKABLE_FUNCTION(, ASYNC_SOCKET_HANDLE, async_socket_create, EXECUTION_ENGINE_HANDLE, execution_engine, SOCKET_HANDLE, socket_handle);

Why does this need to be passed as an argument when it could be created and OWNED by ASYNC_SOCKET?

What's the rationale around that? Shared ownership? 

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:61 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb:
> MOCKABLE_FUNCTION(, ASYNC_SOCKET_SEND_SYNC_RESULT, async_socket_send_async, ASYNC_SOCKET_HANDLE, async_socket, const ASYNC_SOCKET_BUFFER*, payload, uint32_t, buffer_count, ON_ASYNC_SOCKET_SEND_COMPLETE, on_send_complete, void*, on_send_complete_context);

These seem unseparable (the payloads and their number) How about having just 1 data structure that's passed around?

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:161 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -185,12 +149,8 @@ MOCKABLE_FUNCTION(, void, async_socket_close, ASYNC_SOCKET_HANDLE, async_socket)
 
 **SRS_ASYNC_SOCKET_LINUX_11_037: [** `async_socket_close` shall wait for all executing `async_socket_send_async` and `async_socket_receive_async` APIs. **]**
```
anporumb, line 150: >all executing `async_socket_send_async` and `async_socket_receive_async` [](http://example.com/codeflow?start=73&length=72)

This "wait for all" how does it work? it looks like some sort of "counting" would be necessary. 

How is "all" determined?

linux/devdoc/async_socket_linux_requirements.md
```diff
@@ -185,12 +149,8 @@ MOCKABLE_FUNCTION(, void, async_socket_close, ASYNC_SOCKET_HANDLE, async_socket)
...
 
-**SRS_ASYNC_SOCKET_LINUX_11_038: [** Then `async_socket_close` shall remove the underlying socket form the epoll by calling  `epoll_ctl` with `EPOLL_CTL_DEL`. **]**
-
 **SRS_ASYNC_SOCKET_LINUX_11_039: [** `async_socket_close` shall call `close` on the underlying socket. **]**
 
-**SRS_ASYNC_SOCKET_LINUX_11_040: [** `async_socket_close` shall remove any memory that is stored in the epoll system **]**
-
 **SRS_ASYNC_SOCKET_LINUX_11_041: [** `async_socket_close` shall set the state to closed. **]**
```
anporumb, line 154: >closed [](http://example.com/codeflow?start=81&length=6)

Likely capitalized because the other states are written with capital letters.

anporumb:
> An `async_socket` object receives an execution engine as creation argument in order to be able to schedule all asynchronous work using the execution engine.

404 I have not found any indication that the execution_engine is actually used in this module for scheduling work or otherwise.

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:11 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -910,90 +656,49 @@ int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BU
...
+                    wake_by_address_single(&async_socket->pending_api_calls);
                 }
-                (void)interlocked_decrement(&async_socket->pending_api_calls);
-                wake_by_address_single(&async_socket->pending_api_calls);
             }
         }
     }
-
-all_ok:
     return result;
```
anporumb, line 703: >return result; [](http://example.com/codeflow?start=4&length=14)

I have no reviewed this file at all, but I will in a future CR iteration.

anporumb (changes requested):
:clock1:

anporumb:
>     build_test_folder(async_socket_linux_int)

given the extensive changes in the code, I wouldhave expected "some" changes in _int tests.

However, this PR does not contains any changes to int tests.

---
Refers to: linux/tests/CMakeLists.txt:20 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

jebrando:
> MOCKABLE_FUNCTION(, ASYNC_SOCKET_HANDLE, async_socket_create, EXECUTION_ENGINE_HANDLE, execution_engine, SOCKET_HANDLE, socket_handle);

Because the SOCKET_HANDLE gets created at another level today and it.  Moving this into socket handle is a very interesting concept.  I'll consider.

---
In reply to: [1380990751](https://github.com/Azure/c-pal/pull/202#issuecomment-1380990751) [](http://example.com/codeflow?ancestors=1380990751,1380990751)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:61 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

jebrando:
>     ASYNC_SOCKET_OPEN_ERROR

OPEN_ABANDONED would not mean anything because abandoned means that things are closed/closing.

---
In reply to: [1380937208](https://github.com/Azure/c-pal/pull/202#issuecomment-1380937208) [](http://example.com/codeflow?ancestors=1380937208,1380880779)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:26 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
jebrando: Ok, I was wondering about that.  In our base code we have it both ways.  I first started it with the increment happening in the platform call, then as I looked at example I changed it.  I can change it back.

linux/devdoc/async_socket_linux_requirements.md
jebrando: We have a pending call count that letsthis function know that all the executing send_async/receive_async calls.

linux/devdoc/async_socket_linux_requirements.md
jebrando: No it doesn't that gets done before by another level.

jebrando:
> MOCKABLE_FUNCTION(, ASYNC_SOCKET_SEND_SYNC_RESULT, async_socket_send_async, ASYNC_SOCKET_HANDLE, async_socket, const ASYNC_SOCKET_BUFFER*, payload, uint32_t, buffer_count, ON_ASYNC_SOCKET_SEND_COMPLETE, on_send_complete, void*, on_send_complete_context);

I tried that before and the previous PR comment was to separate them because 
1) its easier to test to have them separated
2)This would be the only place it would be used

---
In reply to: [1380992132](https://github.com/Azure/c-pal/pull/202#issuecomment-1380992132) [](http://example.com/codeflow?ancestors=1380992132,1380992132)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:161 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
jebrando: I'm not really sure I understand the question.  In async_socket_send_async it creates a new memory allocation with all the data used to send, so the previous call to send shouldn't affect this call.

linux/devdoc/async_socket_linux_requirements.md
jebrando: ABANDONDED means that the socket on the other end has closed

linux/devdoc/platform_linux_requirements.md
jebrando: Yes.  For now.

linux/devdoc/platform_linux_requirements.md
jebrando: Ones for the positive and this is for the negative.

linux/devdoc/platform_linux_requirements.md
jebrando: Yes that's exactly what it does.

linux/devdoc/platform_linux_requirements.md
jebrando: Because if you're writing the test you increment it after you get it.  I updated the async socket and I'll update this.

linux/inc/c_pal/platform_linux.h
jebrando: Those are in the platform.h in the interface folder.  This is for linux only

jebrando:
>     build_test_folder(async_socket_linux_int)

I wanted to make sure that the int test passed as is to make sure the test pass just like in windows.  In theory we should have one set of int tests for linux and windows.  

---
In reply to: [1380998539](https://github.com/Azure/c-pal/pull/202#issuecomment-1380998539) [](http://example.com/codeflow?ancestors=1380998539,1380998539)

---
Refers to: linux/tests/CMakeLists.txt:20 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

linux/devdoc/async_socket_linux_requirements.md
jebrando: you are correct, I will add it to the completion_port_add spec

linux/devdoc/async_socket_linux_requirements.md
jebrando: Ok, this API is kinda confusing to me and I didn't want to change it because of windows API would have to change too, but the send_async is only called with 1 buffer.  I think we should change this from being an array to be a singular ASYNC_SOCKET_BUFFER.

jebrando: I will run this by our fearless leader and see if I should change it (this will be done in a different PR btw).

linux/devdoc/async_socket_linux_requirements.md
jebrando: Yes, there needs to be a change here.

linux/devdoc/async_socket_linux_requirements.md
jebrando: The code does this, but I didn't write the spec.  I will update.

linux/src/platform_linux.c
jebrando: This has to happen because if you call platform_init g_completion_port is a valid address, then you call platform_deinit which frees the memory (without setting it to NULL).  If you call platform_init again it will not allocate the memory again because g_completion_port will still be a stale address.  So this is why it has to be set to NULL.

jebrando:
> MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_socket, ASYNC_SOCKET_BUFFER*, payload, uint32_t, buffer_count, ON_SOCKET_ASYNC_RECEIVE_COMPLETE, on_receive_complete, void*, on_receive_complete_context);

I'm not sure what type you are wondering about.  

---
In reply to: [1380903280](https://github.com/Azure/c-pal/pull/202#issuecomment-1380903280) [](http://example.com/codeflow?ancestors=1380903280,1380903280)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:209 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

jebrando:
> **SRS_ASYNC_SOCKET_LINUX_11_073: [** Otherwise `async_socket_receive_async` shall create a context for the send where the `payload`, `on_receive_complete` and `on_receive_complete_context` shall be stored. **]**

Yes you are correct.

---
In reply to: [1380904124](https://github.com/Azure/c-pal/pull/202#issuecomment-1380904124) [](http://example.com/codeflow?ancestors=1380904124,1380904124)

---
Refers to: linux/devdoc/async_socket_linux_requirements.md:232 in 8f097b0. [](commit_id = 8f097b0ef0769667fbfd6e41f95a8d9ef5bac052, deletion_comment = False)

anporumb (approved):
:shipit:

PR #205, opened 2023-01-19
Get UUID building and passing tests for Mariner 2.0

darobs:
... or any OS that only supports the OSSP uuid library implementation.

Mariner 2.0 only seems to have the OSSP uuid implementation. We implemented the libuuid version.
This will select the implementation by the location of uuid.h.


CMakeLists.txt
```diff
@@ -175,6 +175,24 @@ set(C_PAL_INC_FOLDER
...
+    elseif(HAVE_UUID_UUID_H)
+        # libuuid implementation
+        add_definitions(-DHAVE_UUID_UUID_H)
+        set(uuid_lib uuid)
+    else()
+        # OSSP uuid
+        add_definitions(-DHAVE_UUID_H)
+        set(uuid_lib ossp-uuid)
+    endif()
+endif()
```
dcristoloveanu, line 194: Is there a better way to check for ossp-uuid than the existence of uuid.h or uuid/uuid.h? Just thinking that it looks a little bit brittle. Maybe some env var?

dcristoloveanu: I did not research by myself, just asking so that maybe we have a second look

linux/devdoc/uuid_linux_requirements.md
```diff
@@ -38,8 +38,28 @@ MOCKABLE_FUNCTION(, int, uuid_produce, UUID_T, destination);
...
+
 **SRS_UUID_LINUX_02_001: [** If `destination` is `NULL` then `uuid_produce` shall fail and return a non-NULL value. **]**
 
 **SRS_UUID_LINUX_02_002: [** `uuid_produce` shall call `uuid_generate` to generate a `UUID`. **]**
 
-**SRS_UUID_LINUX_02_004: [** `uuid_produce` shall succeed and return 0. **]**
\ No newline at end of file
+**SRS_UUID_LINUX_02_004: [** `uuid_produce` shall succeed and return 0. **]**
+
+If the system only has OSSP uuid available:
```
dcristoloveanu, line 49: >stem only has OSSP uuid available: [](http://example.com/codeflow?start=9&length=34)

Reaching here I started to question what is this ossp uuid
libuuid is the since forever default as UUID lib on Linux.

libuuid is available with linux kernel in util-linux.
How come we can't use it and we have to bring in ossp uuid?

linux/devdoc/uuid_linux_requirements.md
```diff
@@ -38,8 +38,28 @@ MOCKABLE_FUNCTION(, int, uuid_produce, UUID_T, destination);
...
+
+**SRS_UUID_LINUX_45_005: [** `uuid_produce` shall call `uuid_export` with fmt = UUID_FMT_BIN to load the value into the destination. **]**
+
+**SRS_UUID_LINUX_45_008: [** If `uuid_export` does not write `size_of(UUID_T)` bytes into the destination, the call shall fail and return a non-zero value. **]**
+
+**SRS_UUID_LINUX_45_006: [** `uuid_produce` shall call `uuid_destroy` to deallocate the uuid. **]**
+
+**SRS_UUID_LINUX_45_007: [** If any uuid call returns a value other than `UUID_RC_OK`, the call shall fail and return a non-zero value. **]**
+
+**SRS_UUID_LINUX_45_002: [** If all uuid calls return success, `uuid_produce` shall succeed and return 0. **]**
```
dcristoloveanu, line 65: This is a separate binding (to ossp uuid), so it should go to a separate file if we really need to support it.

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu:
I shall wait to clear whether we need ossp uuid or not

linux/devdoc/uuid_linux_requirements.md
darobs: Did all this work, and then I found the right package to install. :( 
`tdnf install util-linux-devel` on Mariner 2.0

darobs:
Closing this because I fixed it without using the OSSP uuid library.

PR #206, opened 2023-01-19
Add install targets for CMake.

darobs:
Add install targets for CMake, so I can extract libraries and header files.

Fix a build error for sync_linux_ut.c


dcristoloveanu (approved):
:shipit:

PR #208, opened 2023-01-20
chore: convert timers to use an array for the instance

jebrando:


linux/src/threadpool_linux.c
```diff
@@ -291,6 +308,46 @@ static int reallocate_threadpool_array(THREADPOOL* threadpool)
...
+    // Loop through the list and find the first disabled timer
+    int32_t index;
+    for (index = 0; index < MAX_TIMER_INSTANCE_COUNT; index++)
+    {
+        if (interlocked_add(&threadpool->timer_instance[index].timer_status, 0) == TIMER_DISABLED)
+        {
+            result = &threadpool->timer_instance[index];
+            break;
+        }
+    }
```
mattdurak, line 339: This doesn't work though if two calls to get_next_timer_instance come at the same time.

The first call would find index 0 under the "lock", then the next thread could also find index 0.

Seems like you need to change the timer_status under this "lock".

linux/src/threadpool_linux.c
```diff
@@ -291,6 +308,46 @@ static int reallocate_threadpool_array(THREADPOOL* threadpool)
...
+        if (current_val == 0)
+        {
+            break;
+        }
+        else
+        {
+            // Do Nothing wait for address
+        }
+        (void)wait_on_address(&threadpool->timer_instance_access, current_val, UINT32_MAX);
+    } while (true);
```
mattdurak, line 328: Actually we should be able to do this without a lock.

```
for (...)
{
    if (interlocked_compare_exchange(&threadpool->timer_instance[index].timer_status, TIMER_ENABLED, TIMER_DISABLED) == TIMER_DISABLED)
...
}
```

First one to switch a state from disabled to enabled will get exclusive access of that slot.

Simplifies this function quite a bit.

linux/src/threadpool_linux.c
```diff
@@ -545,15 +608,16 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
...
         if (timer_instance == NULL)
         {
-            LogError("Failure allocating Timer Instance");
+            LogError("Failure getting Timer Instance all timers are being used");
         }
         else
         {
             timer_instance->work_function = work_function;
             timer_instance->work_function_ctx = work_function_ctx;
+            (void)interlocked_exchange(&timer_instance->timer_status, TIMER_ENABLED);
```
mattdurak, line 620: >oid)interlocked_exchange(&timer_instance->timer_status, TIMER_ENABLED); [](http://example.com/codeflow?start=14&length=71)

needs to happen atomically when we selected the slot

linux/src/threadpool_linux.c
```diff
@@ -103,7 +117,10 @@ static void on_timer_callback(sigval_t timer_data)
     }
     else
     {
-        timer_instance->work_function(timer_instance->work_function_ctx);
+        if (interlocked_add(&timer_instance->timer_status, 0) == TIMER_ENABLED)
+        {
+            timer_instance->work_function(timer_instance->work_function_ctx);
+        }
     }
 }
```
mattdurak, line 125: Do we have any sort of unique identifier to match in the `timer_data` with the timer start?

Currently I think your solution avoids a segfault, but it's not correct in all cases.

Imagine you have 63 timers set already.

Now:
1. start a timer (A) with a 1 second period
2. some time later stop the timer (A), which will set the state to DISABLED
3. start a timer (B) with a 1 hour initial period

It's entirely possible that timer A fires after you delete it (hence your fix).

But now the handler for timer B will see the timer intended for A.

Clearly that's broken if we rely on this for something like a watchdog (we could re-use a timer slot and immediately fire the watchdog).

I think you could even catch this in a chaos test (fill up the 63 slots then have a thread that alternates between starting a very short timer and starting a timer with a long enough time that it will never fire and assert if it does).

linux/src/threadpool_linux.c
jebrando: Good point, I'll change this.

linux/src/threadpool_linux.c
mattdurak: Possibly could be solved by having some extra indirection/lookup.

If instead of this:

```
TIMER_INSTANCE* timer_instance = timer_data.sival_ptr;
```

we had some intermediate:

```
TIMER_IDENTIFIER timer_identifier = timer_data.sival_ptr;
TIMER_INSTANCE* timer_instance = some_lookup_function(timer_identifier);
```

But I'm looking at sigval now and I see we only have the one `void*` or `int` and we can't have another context.

I think we really need 2 fields in the callback to do what I'm thinking of, because my "some_lookup_function" likely needs to be "TIMER_INSTANCE* some_lookup_function(THREADPOOL* threadpool, TIMER_IDENTIFIER timer_identifier)" or you would have to have a global array in that lookup function.

linux/src/threadpool_linux.c
```diff
@@ -598,7 +641,6 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
                     LogError("Failure calling timer_delete. Error: %d: (%s)", errno, err_msg);
                 }
             }
-            free(timer_instance);
```
mattdurak, line 601: >  free(timer_instance); [](http://example.com/codeflow?start=10&length=23)

we can leak timers in the ENABLED state on the error path here now

jebrando: No the timers never need to be freed since it's just an array and not malloced.

linux/src/threadpool_linux.c
jebrando: The best way to solve this would be for the calling function in the callback to have the unique ID.  This would make sure that the callback is always set for you, but it does put undue work on the calling function.  I added a chaos test for this but I haven't gotten it to fail yet (I probably need to tweak it a little).  I don't know if this is an actual problem or an in-theory problem.  I'll continue to look for a reliable solution

mattdurak: I'm not sure how you would check for a unique ID. You have to store it somewhere, so if you store it in the fixed array, then it's still the same problem.

If the caller stores it in their context I don't see how that helps.

Timer A will have context A (with unique ID 1)
Timer B will have context B (with unique ID 2)

But say timer slot 0 is re-used.
First it gets context A (id 1)
If we re-use slot 0 with context B and it fires due to the original registration, then it will just call the callback for B with context B and unique ID 2.

So even if the timer is due to A, it will be handled by B and look exactly like it "should" be for for B.

Maybe a simpler work-around for this tiny race is to have some timestamp or something so we block before re-using a timer slot (e..g don't re-use the slot until 1-5 seconds or something has elapsed or something).

jebrando: I think I might have a solution.  Let me know what you think.

linux/src/threadpool_linux.c
```diff
@@ -103,7 +130,40 @@ static void on_timer_callback(sigval_t timer_data)
     }
     else
     {
-        timer_instance->work_function(timer_instance->work_function_ctx);
+        if (interlocked_add(&timer_instance->timer_status, 0) == TIMER_ENABLED)
+        {
+            // Make sure this timer is
```
mattdurak, line 135: >er is  [](http://example.com/codeflow?start=33&length=6)

is...?

linux/src/threadpool_linux.c
```diff
@@ -103,7 +130,40 @@ static void on_timer_callback(sigval_t timer_data)
...
+        {
+            // Make sure this timer is 
+            if (interlocked_add(&timer_instance->timer_trigger, 0) == 0)
+            {
+                // Increment the timer trigger so we know it won't go again
+                interlocked_increment(&timer_instance->timer_trigger);
+                struct timespec ts;
+                if (clock_gettime(CLOCK_REALTIME, &ts) == -1)
+                {
+                    LogInfo("Failure getting time from clock");
```
mattdurak, line 143: >LogInfo [](http://example.com/codeflow?start=20&length=7)

LogError, seems like we could be dropping a callback call here if this fails

linux/src/threadpool_linux.c
```diff
@@ -103,7 +130,40 @@ static void on_timer_callback(sigval_t timer_data)
...
+                }
+                else
+                {
+                    struct timespec result;
+                    timespec_diff_macro(ts, timer_instance->start_time, result);
+
+                    if (result.tv_sec - timer_instance->its.it_interval.tv_sec < 0)
+                    {
+                        // The timer triggered too soo, this is to combat a race condition in POSIX
```
mattdurak, line 152: >soo [](http://example.com/codeflow?start=51&length=3)

soon

linux/src/threadpool_linux.c
```diff
@@ -593,12 +680,9 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
 
                 if (timer_delete(time_id) != 0)
                 {
-                    char err_msg[128];
-                    (void)strerror_r(errno, err_msg, 128);
-                    LogError("Failure calling timer_delete. Error: %d: (%s)", errno, err_msg);
+                    LogErrorNo("Failure calling timer_delete.");
                 }
             }
-            free(timer_instance);
```
mattdurak, line 601: >  free(timer_instance); [](http://example.com/codeflow?start=10&length=23)

still need to reset to DISABLED here

linux/src/threadpool_linux.c
```diff
@@ -103,7 +130,40 @@ static void on_timer_callback(sigval_t timer_data)
...
+                {
+                    LogInfo("Failure getting time from clock");
+                }
+                else
+                {
+                    struct timespec result;
+                    timespec_diff_macro(ts, timer_instance->start_time, result);
+
+                    if (result.tv_sec - timer_instance->its.it_interval.tv_sec < 0)
+                    {
```
mattdurak, line 151: So the test is: did line 141 (clock_gettime) happen before line 640 when we registered another timer.

Seems like this might work, but are we basing the resolution on seconds? Could the tv_sec difference be 0 but the tv_usec is negative? So we only have a resolution of seconds here?

Besides that, I'm wondering if there is still a race, since we are now checking 3 different things  to synchronize on (timer_status, timer_trigger, start_time).

Couldn't an "old" call to "on_timer_callback" from the system get suspended around line 141 (before calling gettime here) then resume "later" so that it appears valid (enough time has passed)? Possibly could be fixed by getting the time before checking our synchronization variables.

My suspicion is we should consider putting more info into "timer_status", maybe just have multiple states like UNUSED, STARTING, STARTED, FIRING? That might make it easier to synchronize here and in the timer start function.

I also think we need some kind of volatile access on start_time or we could read it in this function before/during setting it for a new timer. So we still read the old start time or we see an invalid time assuming it's large enough that the write isn't atomic.

I need to think about this a bit more (tomorrow)

jebrando: After some time I figured out the what happens with the unwanted callbacks.

Say you have a callback that fires at 01:00:23:05
And you call the timer_delete at 01:00:23:01

The race condition in the timer system is such that the delete doesn't get to cancel the timer triggering.  So you won't get a mystery callback after that initial callback firing.  This is why it only happens once in a very long while because the timer and the delete have to hit within a small window.  I don't know the size of the window for the delete call and the removing of the callback and the doc doesn't even attempt to give you a clue.  I figure that the difference in the timeout that you set and the callback should not be closer than a second (that's why I remove the microseconds from the equation).  Since this will only happen once I made it so you don't have to go through this code after the first iteration.  

I also made some changes due to helgrind giving me clues as to where I should add interlocks.

mattdurak: I think my first actionable comment here is can we:

1. Combine `timer_status` and `timer_trigger` to one variable if possible
2. Get the current time earlier in `on_timer_callback`
3. Guard  writing to and reading `start_time` in some critical section (either just use the same interlocked with a "busy" state like a spin lock or something else)

linux/src/threadpool_linux.c
```diff
@@ -593,12 +689,10 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
 
                 if (timer_delete(time_id) != 0)
                 {
-                    char err_msg[128];
-                    (void)strerror_r(errno, err_msg, 128);
-                    LogError("Failure calling timer_delete. Error: %d: (%s)", errno, err_msg);
+                    LogErrorNo("Failure calling timer_delete.");
                 }
+                (void)interlocked_exchange(&timer_instance->timer_status,  TIMER_DISABLED);
```
mattdurak, line 694: >(void)interlocked_exchange(&timer_instance->timer_status,  TIMER_DISABLED); [](http://example.com/codeflow?start=16&length=75)

belongs after line 695 I think (in case timer_create fails)

linux/src/threadpool_linux.c
```diff
@@ -545,15 +639,20 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
...
+            LogError("Failure getting Timer Instance all timers are being used");
+        }
+        else if (clock_gettime(CLOCK_REALTIME, &timer_instance->start_time) == -1)
+        {
+            LogError("Failure clock_gettime");
         }
         else
         {
             timer_instance->work_function = work_function;
             timer_instance->work_function_ctx = work_function_ctx;
```
mattdurak, line 654: Ok so thinking about any more races here between a thread (T1) calling timer start, and another thread (T2) which is an extra timer callback from a deleted timer.

T1 == threadpool_timer_start
T1:1 set state ENABLED
T1:2 set start_time
T1:3 set timer_trigger 0
T1:4 timer_create

T2 == on_timer_callback
T2:1 check state ENABLED
T2:2 check timer_trigger 0
T2:3 set timer_trigger 1
T2:4 get time
T2:5 compare times
T2:6 call callback

We have a race if T1:1, T1:2 happens, then T2:1, T2:2 happens (which will not see timer_trigger as 0 so it will just call the callback from the else branch)

Otherwise I think the order is correct.

Why don't we get rid of `timer_trigger`. Instead we can have the states ENABLING, ENABLED, DISABLED. T1:1 sets to ENABLING first, then it can change to ENABLED before T1:4.
That way T2 will not see it ENABLED until we are done with all that setup.

This would solve the other comment I was going to make which is I don't know if setting `start_time` with `clock_gettime` is atomic. If we essentially "lock" changing `start_time` under a critical section (state==ENABLING) then we won't ever see partial updates from the other thread.

mattdurak (approved):
:shipit:

PR #209, opened 2023-01-24
Add installation step to c-pal interface and pal_linux.

darobs:
Add installation step to c-pal interface and pal_linux.  
Added public headers to c_pal project so they would install.


mattdurak (approved):
:shipit:

PR #212, opened 2023-01-31
chore: update sm requirement to handle close while opening callback

jebrando:


mattdurak:
> `sm_close_begin_with_cb` will invoke a callback function between prohibiting all the calls and waiting for the number of executing calls to be `0`. The callback function is useful in various cases, such as, when we need to cancel the ongoing operations, before we wait for the outstanding calls to finish.

we should add onto this with the motivation for the new callback you are adding

---
Refers to: common/devdoc/sm_requirements.md:61 in b51b725. [](commit_id = b51b725b3da81b33058244ecfeb4ddbf0b1d190a, deletion_comment = False)

common/devdoc/sm_requirements.md
```diff
@@ -55,7 +58,7 @@ To achieve the above goals, `sm` maintains the following state:
...
-`sm_close_begin_with_cb` will invoke a callback function between prohibiting all the calls and waiting for the number of executing calls to be `0`. The callback function is useful in various cases, such as, when we need to cancel the ongoing operations, before we wait for the outstanding calls to finish.
+`sm_close_begin_with_cb` will invoke a callback function between prohibiting all the calls and waiting for the number of executing calls to be `0`. The callback function is useful in various cases, such as, when we need to cancel the ongoing operations, before we wait for the outstanding calls to finish.  `sm_close_begin_with_cb` will also invoke an addition callback function (`ON_SM_CLOSING_WILE_OPENING_CALLBACK`) which will let the calling function know if sm is currently in the `opening` state.  This is useful in the scenario where the user will need to do additional steps to close if they are in the opening state.
```
mattdurak, line 61: >addition [](http://example.com/codeflow?start=352&length=8)

additional

mattdurak, line 61: >ing state. [](http://example.com/codeflow?start=615&length=10)

I'd maybe just add something like, "It is expected that the user will perform actions to complete the pending open in this callback before returning so that the close can continue."

Just because after the callback returns sm will re-check the state to attempt again to close.

common/devdoc/sm_requirements.md
```diff
@@ -197,13 +209,18 @@ static SM_RESULT sm_close_begin_internal(SM_HANDLE sm, ON_SM_CLOSING_COMPLETE_CA
 
 **SRS_SM_02_051: [** If the state is `SM_OPENED_DRAINING_TO_BARRIER` then `sm_close_begin_internal` shall re-evaluate the state. **]**
 
+If the state is `SM_OPENING` and the `close_while_opening_callback` is non-NULL then ...
+
+... `sm_close_begin_internal` shall call the `close_while_opening_callback` and then re-evaluate the state.
```
mattdurak, line 214: >m_close_begin_internal` shall call the `close_while_opening_callback` and then re-evaluate the state. [](http://example.com/codeflow?start=6&length=101)

I just realized that this means if the callback doesn't actually return the state of sm to something other than OPENING, it will keep getting called (possibly forever).

I wonder if we should either only call the callback once and just fail if still in opening the next time, or we should have a return value on the callback so the callback handler can fail if it's not possible to get out of the OPENING state.

I'm leaning toward the former because it seems like the callback should either be able to resolve this or not the first time, but maybe there's a case for "retrying" that callback (I just don't see it).

I just feel that it might help to have some way to avoid getting stuck.

mattdurak (approved):
:shipit:

PR #215, opened 2023-02-01
chore: adding implementation for sm_close while opening

jebrando:


common/tests/sm_int/sm_int.c
```diff
@@ -1635,11 +1637,57 @@ TEST_FUNCTION(sm_close_begin_with_cb_triggers_cancel)
...
+    ASSERT_IS_NOT_NULL(wait_for_cancel.sm);
+
+    ASSERT_ARE_EQUAL(SM_RESULT, sm_open_begin(wait_for_cancel.sm), SM_EXEC_GRANTED);
+
+    // This call should be refused because we are going to fail the sm_open_end
+    ASSERT_ARE_EQUAL(SM_RESULT, sm_close_begin_with_cb(wait_for_cancel.sm, cancellingCallback, (void*)&wait_for_cancel.cancel_wait, close_while_in_opening, &wait_for_cancel), SM_EXEC_REFUSED);
+
+    ASSERT_ARE_EQUAL(int, 1, wait_for_cancel.close_cb_called);
+
+    sm_destroy(wait_for_cancel.sm);
```
mattdurak, line 1670: missing arrange/act/assert (here and below)

common/tests/sm_int/sm_int.c
```diff
@@ -1635,11 +1637,57 @@ TEST_FUNCTION(sm_close_begin_with_cb_triggers_cancel)
     ThreadAPI_Sleep(500);
 
     /// assert
-    ASSERT_ARE_EQUAL(SM_RESULT, sm_close_begin_with_cb(context.sm, cancellingCallback, (void*)(&context.cancel_wait)), SM_EXEC_GRANTED);
+    ASSERT_ARE_EQUAL(SM_RESULT, sm_close_begin_with_cb(context.sm, cancellingCallback, (void*)(&context.cancel_wait), NULL, NULL), SM_EXEC_GRANTED);
```
mattdurak, line 1640: >  ASSERT_ARE_EQUAL(SM_RESULT, sm_close_begin_with_cb(context.sm, cancellingCallback, (void*)(&context.cancel_wait), NULL, NULL), SM_EXEC_GRANTED); [](http://example.com/codeflow?start=2&length=146)

here and below, should be ASSERT_ARE_EQUAL(type, EXPECTED, ACTUAL) so that the error message makes sense (so the args are flipped right now)

common/tests/sm_ut/sm_ut.c
```diff
@@ -700,6 +704,29 @@ TEST_FUNCTION(sm_close_begin_with_cb_unhappy_path)
     sm_destroy(sm);
 }
 
+// Tests_SRS_SM_11_002: [ If the state is SM_OPENING, the close_while_opening_callback is non-NULL and this is the first evaluation of the close then ... ]
```
mattdurak, line 707: > and this is the first evaluation of the close then ... ] [](http://example.com/codeflow?start=98&length=57)

shouldn't we have another test showing the callback is only called once?

mattdurak (approved):
:shipit:

common/tests/sm_ut/sm_ut.c
jebrando: This test actually goes through the loop twice since we don't call sm_open_end in the callback, so we are ensuring that the callback is only called once.

common/inc/c_pal/sm.h
```diff
@@ -37,7 +38,7 @@ MOCKABLE_FUNCTION(, SM_RESULT, sm_open_begin, SM_HANDLE, sm);
 MOCKABLE_FUNCTION(, void, sm_open_end, SM_HANDLE, sm, bool, success);
 
 MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin, SM_HANDLE, sm);
-MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin_with_cb, SM_HANDLE, sm, ON_SM_CLOSING_COMPLETE_CALLBACK, callback, void*, callback_context);
+MOCKABLE_FUNCTION(, SM_RESULT, sm_close_begin_with_cb, SM_HANDLE, sm, ON_SM_CLOSING_COMPLETE_CALLBACK, callback, void*, callback_context, ON_SM_CLOSING_WILE_OPENING_CALLBACK, close_while_opening_callback, void*, close_while_opening_context);
```
anporumb, line 41: >WILE [](http://example.com/codeflow?start=152&length=4)

WILE?

common/src/sm.c
```diff
@@ -267,6 +268,17 @@ static SM_RESULT sm_close_begin_internal(SM_HANDLE sm, ON_SM_CLOSING_COMPLETE_CA
...
             }
+            else if (
+                // Codes_SRS_SM_11_002: [ If the state is SM_OPENING, the close_while_opening_callback is non-NULL and this is the first evaluation of the close then ... ]
+                ((state & SM_STATE_MASK) == SM_OPENING) && 
+                (close_while_opening_callback != NULL) &&
+                (initial_iteration)
+                )
+            {
+                // Codes_SRS_SM_11_003: [ ... sm_close_begin_internal shall call the close_while_opening_callback and then re-evaluate the state. ]
+                close_while_opening_callback(close_while_opening_context);
```
anporumb, line 279: >close_while_opening_callback [](http://example.com/codeflow?start=16&length=28)

I suspect this callback can only be called in its own little state.

PR #233, opened 2023-04-11
Threadpool THANDLE-ifyed

jebrando:


interfaces/devdoc/threadpool_requirements.md
```diff
@@ -22,19 +22,12 @@ The lifetime of the execution engine should supersede the lifetime of the `threa
...
-MU_DEFINE_ENUM(THREADPOOL_OPEN_RESULT, THREADPOOL_OPEN_RESULT_VALUES)
-
-typedef void (*ON_THREADPOOL_OPEN_COMPLETE)(void* context, THREADPOOL_OPEN_RESULT open_result);
 typedef void (*THREADPOOL_WORK_FUNCTION)(void* context);
 
 MOCKABLE_FUNCTION(, THREADPOOL_HANDLE, threadpool_create, EXECUTION_ENGINE_HANDLE, execution_engine);
 MOCKABLE_FUNCTION(, void, threadpool_destroy, THREADPOOL_HANDLE, threadpool);
 
-MOCKABLE_FUNCTION(, int, threadpool_open_async, THREADPOOL_HANDLE, threadpool, ON_THREADPOOL_OPEN_COMPLETE, on_open_complete, void*, on_open_complete_context);
+MOCKABLE_FUNCTION(, int, threadpool_open, THREADPOOL_HANDLE, threadpool);
```
mattdurak, line 30: >threadpool_open [](http://example.com/codeflow?start=25&length=15)

:)

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -78,29 +71,23 @@ MOCKABLE_FUNCTION(, void, threadpool_destroy, THREADPOOL_HANDLE, threadpool);
...
-**NON_THREADPOOL_01_009: [** If `on_open_complete` is `NULL`, `threadpool_open_async` shall fail and return a non-zero value. **]**
-
-**NON_THREADPOOL_01_010: [** `on_open_complete_context` shall be allowed to be `NULL`. **]**
+`threadpool_open` opens the threadpool object so that subsequent calls to `threadpool_schedule_work` can be made.
 
-**NON_THREADPOOL_01_011: [** Otherwise, `threadpool_open_async` shall switch the state to OPENING and perform any actions needed to open the `threadpool` object. **]**
+**NON_THREADPOOL_01_008: [** If `threadpool` is `NULL`, `threadpool_open` shall fail and return a non-zero value. **]**
 
-**NON_THREADPOOL_01_012: [** On success, `threadpool_open_async` shall return 0. **]**
+**NON_THREADPOOL_01_009: [** If `on_open_complete` is `NULL`, `threadpool_open` shall fail and return a non-zero value. **]**
```
mattdurak, line 84: > If `on_open_complete` is `NULL [](http://example.com/codeflow?start=28&length=31)

remove

interfaces/inc/c_pal/threadpool.h
```diff
@@ -10,34 +10,31 @@
...
-MU_DEFINE_ENUM(THREADPOOL_OPEN_RESULT, THREADPOOL_OPEN_RESULT_VALUES)
-
-typedef void (*ON_THREADPOOL_OPEN_COMPLETE)(void* context, THREADPOOL_OPEN_RESULT open_result);
 typedef void (*THREADPOOL_WORK_FUNCTION)(void* context);
 
-MOCKABLE_FUNCTION(, THREADPOOL_HANDLE, threadpool_create, EXECUTION_ENGINE_HANDLE, execution_engine);
-MOCKABLE_FUNCTION(, void, threadpool_destroy, THREADPOOL_HANDLE, threadpool);
+THANDLE_TYPE_DECLARE(THREADPOOL);
+
+MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HANDLE, execution_engine);
```
mattdurak, line 30: >THANDLE(THREADPOOL), [](http://example.com/codeflow?start=20&length=20)

looks like the MD needs updates

mattdurak: (in the interfaces/devdoc folder)

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -124,31 +116,29 @@ MOCKABLE_FUNCTION(, void, threadpool_destroy, THREADPOOL_HANDLE, threadpool);
 
 **SRS_THREADPOOL_LINUX_07_016: [** `threadpool_destroy` shall free the memory allocated in `threadpool_create`. **]**
 
-### threadpool_open_async
+### threadpool_open
 
 ```C
-MOCKABLE_FUNCTION(, int, threadpool_open_async, THREADPOOL_HANDLE, threadpool, ON_THREADPOOL_OPEN_COMPLETE, on_open_complete, void*, on_open_complete_context);
+MOCKABLE_FUNCTION(, int, threadpool_open, THREADPOOL_HANDLE, threadpool);
```
mattdurak, line 122: >THREADPOOL_HANDLE [](http://example.com/codeflow?start=42&length=17)

THANDLE here and below

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -124,31 +116,29 @@ MOCKABLE_FUNCTION(, void, threadpool_destroy, THREADPOOL_HANDLE, threadpool);
...
 ```
 
-`threadpool_open_async` opens the threadpool asynchronously.
-
-**SRS_THREADPOOL_LINUX_07_017: [** If `threadpool` is `NULL`, `threadpool_open_async` shall fail and return a non-zero value. **]**
+`threadpool_open` opens the threadpool asynchronously.
```
mattdurak, line 125: >asynchronously [](http://example.com/codeflow?start=39&length=14)

remove

mattdurak:
> MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);

THANDLE here and below

---
Refers to: win32/devdoc/threadpool_win32.md:102 in 50a093b. [](commit_id = 50a093b2f5189e72e7e1dba2f3448671cd02536a, deletion_comment = False)

win32/src/threadpool_win32.c
```diff
@@ -99,14 +100,38 @@ static void internal_close(THREADPOOL_HANDLE threadpool)
...
+        (void)WaitOnAddress(&threadpool->state, &current_state, sizeof(current_state), INFINITE);
+    } while (1);
 
+    /* Codes_SRS_THREADPOOL_WIN32_42_028: [ threadpool_dispose shall decrement the reference count on the execution_engine. ]*/
+    execution_engine_dec_ref(threadpool->execution_engine);
+}
+
+THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
+{
+    THREADPOOL* result = NULL;
```
mattdurak, line 134: >THREADPOOL [](http://example.com/codeflow?start=4&length=10)

THANDLE(THREADPOOL)

interfaces/devdoc/threadpool_requirements.md
jebrando: Why do we have 3 md's for threadpool?

mattdurak: The PAL is organized such that there is an MD for the interface, with only high level requirements (likely could be tested in the int test), then an MD for each platform with actual requirements like shall call x-API (which are definitely meant to be tested in platform-specific UT's).

I think it makes sense to have the separate docs, though maybe repeating the header isn't needed in 3 places...

win32/src/threadpool_win32.c
jebrando: THANDLE(Threadpool) will not build because it make this const and you have to initialize it here.  This needs to be THREADPOOL*

win32/src/threadpool_win32.c
```diff
@@ -433,8 +420,7 @@ int threadpool_timer_start(THREADPOOL_HANDLE threadpool, uint32_t start_delay_ms
             else
             {
                 /* Codes_SRS_THREADPOOL_WIN32_42_006: [ threadpool_timer_start shall call CreateThreadpoolTimer to schedule execution the callback while passing to it the on_timer_callback function and the newly created context. ]*/
-                PTP_TIMER tp_timer = CreateThreadpoolTimer(on_timer_callback, timer_temp, &threadpool->tp_environment);
-
+                PTP_TIMER tp_timer = CreateThreadpoolTimer(on_timer_callback, timer_temp, NULL);
```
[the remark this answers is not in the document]
jebrando: I took this PR up through zprc and when you use the CreateThreadpoolTimer with the created tp_environment it throws an EV with Invalid Parameter on Cleanup.  I was reading some of the docs and it was mentioning that it could use the default Environment and when I remove the created Environment, it passes.  I've been looking for an explanation as to why use a created environment instead of the default environment, without much clarity.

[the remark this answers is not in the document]
mattdurak: Why use the non-default threadpool is:
- Set the min/max number of threads and actually use it as a pool you control
- Be able to close this threadpool, which will call the closecleanup group and it will block on pending timers

If you want to decouple a timer from the threadpool, why not create a separate threadpool with the default (NULL) execution engine)?

win32/src/threadpool_win32.c
mattdurak: but we are returning a THANDLE(THREADPOOL), so result should technically be that type. Typically we do this by assigning to NULL at declaration and using the THANDLE_GET_T macro as needed.

I mean, it will compile as-is, just looks a bit wrong to me because `THANDLE(THREADPOOL)` is actually a different type than `THREADPOOL*`

mattdurak (approved):
:shipit:

PR #237, opened 2023-05-11
chore: fix helgrind failure in async_socket_linux

jebrando:


linux/src/async_socket_linux.c
```diff
@@ -274,6 +282,12 @@ static void internal_close(ASYNC_SOCKET_HANDLE async_socket)
         completion_port_remove(async_socket->completion_port, async_socket->socket_handle);
     }
 
+    // Codes_SRS_ASYNC_SOCKET_LINUX_11_103: [ async_socket_close shall wait for the event_complete_callback to complete ]
+    while ((value = interlocked_add(&async_socket->event_complete_executing, 0)) != 0)
+    {
+        (void)wait_on_address(&async_socket->event_complete_executing, value, UINT32_MAX);
```
anporumb, line 288: >UINT32_MAX [](http://example.com/codeflow?start=78&length=10)

`while` (line 286) and `UINT32_MAX` here really prepare us for eternity.

How about having a timeout of 1 second and Log something every time `wait_on_address` doesn't get it signalled? That way it is more descriptive of what happened.

Sort of here and everywhere as I see the same pattern used in other places too.

linux/src/async_socket_linux.c
```diff
@@ -257,6 +261,10 @@ static void event_complete_callback(void* context, COMPLETION_PORT_EPOLL_ACTION
                 break;
             }
         }
+        if (interlocked_decrement(event_complete_executing) == 0)
+        {
+            wake_by_address_single(event_complete_executing);
```
anporumb, line 266: >wake_by_address_single [](http://example.com/codeflow?start=12&length=22)

The only time we're listening for this number of events to reach "0" is in `ASYNC_SOCKET_LINUX_STATE_CLOSING` state.

I think it is better to only wake up in that state.

linux/src/async_socket_linux.c
```diff
@@ -274,6 +282,12 @@ static void internal_close(ASYNC_SOCKET_HANDLE async_socket)
         completion_port_remove(async_socket->completion_port, async_socket->socket_handle);
```
anporumb, line 282: >completion_port_remove [](http://example.com/codeflow?start=8&length=22)

Looking in the code, it seems that `event_complete_callback` is given to the completion port.

So line 282 is detaching the socket from the completion port. 

Shouldn't THAT take care of no ongoing calls rather than this layer?

I would expect that at the call of `completion_port_remove` there's NO WAY in which not all the calls are exhauster or there's still ongoing threads that have started and not returned yet to the completion port.

linux/src/async_socket_linux.c
```diff
@@ -115,6 +116,9 @@ static void event_complete_callback(void* context, COMPLETION_PORT_EPOLL_ACTION
     else
     {
         ASYNC_SOCKET_IO_CONTEXT* io_context = (ASYNC_SOCKET_IO_CONTEXT*)context;
```
anporumb, line 118: >context [](http://example.com/codeflow?start=72&length=7)

What happens if you insert here a Sleep() and while the sleep is executing, you call "_close"?

Wouldn't that lead to the same helgrind error?

I will be waiting.

anporumb (changes requested):
:clock1:

linux/src/async_socket_linux.c
jebrando: detaching the socket from the completion port seems to be async and so the failure we're seeing is a race between line 292 and line 152

linux/src/async_socket_linux.c
jebrando: I see, but the issue would be we could get 1000s of logs on a situation that is not a problem and it may happens often.

linux/src/async_socket_linux.c
jebrando: I could test to see if the module is in the open state, and then I'll be sure you can execute it.  Let me test that first.

anporumb:
>     while ((value = interlocked_add(&async_socket->pending_api_calls, 0)) != 0)

Maybe it is not a good idea for me to look too much at this file.

So here's a question: why isn't `pending_api_calls` enough of a check and we ALSO need `event_complete_executing`?

---
Refers to: linux/src/async_socket_linux.c:287 in 080adbd. [](commit_id = 080adbdc2d369135440461d5ed9a5b7da2788da2, deletion_comment = False)

anporumb:
>                 (void)interlocked_increment(&async_socket->pending_api_calls);

This is sort of "here and everywhere".

---
In reply to: [1548056058](https://github.com/Azure/c-pal/pull/237#issuecomment-1548056058) [](http://example.com/codeflow?ancestors=1548056058)

---
Refers to: linux/src/async_socket_linux.c:531 in 080adbd. [](commit_id = 080adbdc2d369135440461d5ed9a5b7da2788da2, deletion_comment = False)

linux/src/async_socket_linux.c
anporumb: I've changed the state of this item from "WontFix" to "Active" because I truly believe that the completion port bears some responsibility here. And maybe the fix lies both with the completion port AND within this file.

anporumb:
>     return result;

Iteration 6 comment: I believe there's multiple issues with this file, some of them I've noticed in previous iterations, some in iteration 6 and I don't think (or I don't see how!) we are addressing them.

So I will still be "waiting".

---
Refers to: linux/src/async_socket_linux.c:738 in 080adbd. [](commit_id = 080adbdc2d369135440461d5ed9a5b7da2788da2, deletion_comment = False)

linux/src/async_socket_linux.c
jebrando: The reason the event_complete_executing is added to protect the event_complete_callback function.  This is to ensure we don't close while the event_complete_callback is executing.

jebrando:
>     while ((value = interlocked_add(&async_socket->pending_api_calls, 0)) != 0)

pending_api_call protects against closing while the API's (async_socket_send_async and async_socket_receive_async) are executing, while event_complete_executing protects against closing while in the event_complete_callback function.

---
In reply to: [1548050945](https://github.com/Azure/c-pal/pull/237#issuecomment-1548050945) [](http://example.com/codeflow?ancestors=1548050945,1548050945)

---
Refers to: linux/src/async_socket_linux.c:287 in 080adbd. [](commit_id = 080adbdc2d369135440461d5ed9a5b7da2788da2, deletion_comment = False)

PR #247, opened 2023-06-22
srw lock ll specs

dcristoloveanu:
We need to split the current srw_lock in 2 layers, thus creating a srw_lock_ll which should be a simple wrapper of a lock, without any memory allocation.

### DIFF
--- a/interfaces/devdoc/srw_lock_ll_requirements.md
+++ b/interfaces/devdoc/srw_lock_ll_requirements.md
@@ -4,6 +4,10 @@
 
 `srw_lock_ll` is the implementation of a slim reader writer lock.
 
+References:
+
+[https://learn.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks](https://learn.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks)
+
 ## Exposed API
 
 ```c
@@ -12,11 +16,9 @@
     SRW_LOCK_LL_TRY_ACQUIRE_COULD_NOT_ACQUIRE, \
     SRW_LOCK_LL_TRY_ACQUIRE_INVALID_ARGS
 
-typedef void* SRW_LOCK_LL;
-
 MU_DEFINE_ENUM(SRW_LOCK_LL_TRY_ACQUIRE_RESULT, SRW_LOCK_LL_TRY_ACQUIRE_RESULT_VALUES)
 
-MOCKABLE_FUNCTION(, void, srw_lock_ll_init);
+MOCKABLE_FUNCTION(, int, srw_lock_ll_init, SRW_LOCK_LL*, srw_lock_ll);
 MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
 
 /*writer APIs*/
@@ -30,6 +32,8 @@ MOCKABLE_FUNCTION(, SRW_LOCK_LL_TRY_ACQUIRE_RESULT, srw_lock_ll_try_acquire_shar
 MOCKABLE_FUNCTION(, void, srw_lock_ll_release_shared, SRW_LOCK_LL*, srw_lock_ll);
 ```
 
+Note that the type `SRW_LOCK_LL` is defined individually for each platform.
+
 ### srw_lock_ll_init
 ```c
 MOCKABLE_FUNCTION(, void, srw_lock_ll_init, SRW_LOCK_LL*, srw_lock_ll);
@@ -42,7 +46,7 @@ MOCKABLE_FUNCTION(, void, srw_lock_ll_init, SRW_LOCK_LL*, srw_lock_ll);
 MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
 ```
 
-`srw_lock_ll_deinit` frees deinitializes the slim reader writer lock.
+`srw_lock_ll_deinit` deinitializes the slim reader writer lock.
 
 ### srw_lock_ll_acquire_exclusive
 ```c
@@ -71,7 +75,7 @@ MOCKABLE_FUNCTION(, void, srw_lock_ll_release_exclusive, SRW_LOCK_LL*, srw_lock_
 MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_shared, SRW_LOCK_LL*, srw_lock_ll);
 ```
 
-`srw_lock_ll_acquire_shared` acquires the SRWLOCK in shared (read) mode.
+`srw_lock_ll_acquire_shared` acquires the slim reader writer lock in shared (read) mode.
 
 ### srw_lock_ll_try_acquire_shared
 ```c
--- a/interfaces/devdoc/srw_lock_ll_requirements.md
+++ b/interfaces/devdoc/srw_lock_ll_requirements.md
@@ -1,8 +1,8 @@
-# `srw_lock` requirements
+# `srw_lock_ll` requirements
 
 ## Overview
 
-`srw_lock` is a wrapper over a `SRWLOCK` with the additional benefit of having some statistics printed.
+`srw_lock_ll` is the implementation of a slim reader writer lock (on Windows it simply wraps`SRWLOCK`).
 
 ## Exposed API
 
@@ -20,7 +20,7 @@ MOCKABLE_FUNCTION(, void, srw_lock_ll_init);
 MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
 
 /*writer APIs*/
-MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_exclusive, SRW_LOCK_LL*, handle);
+MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
 MOCKABLE_FUNCTION(, SRW_LOCK_LL_TRY_ACQUIRE_RESULT, srw_lock_ll_try_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
 MOCKABLE_FUNCTION(, void, srw_lock_ll_release_exclusive, SRW_LOCK_LL*, srw_lock_ll);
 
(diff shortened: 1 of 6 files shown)


ACKPaul (approved):

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Exposed API
+
+```c
+#define SRW_LOCK_LL_TRY_ACQUIRE_RESULT_VALUES \
+    SRW_LOCK_LL_TRY_ACQUIRE_OK, \
+    SRW_LOCK_LL_TRY_ACQUIRE_COULD_NOT_ACQUIRE, \
+    SRW_LOCK_LL_TRY_ACQUIRE_INVALID_ARGS
+
+typedef void* SRW_LOCK_LL;
```
anporumb, line 15: >void* [](http://example.com/codeflow?start=8&length=5)

how about... instead of void... have an incomplete type here.

Such as `typedef struct HAGAUAGA* SWL_LOCK_LL". 

That'll make SRW_LOCK_LL pointers only compatible among themselves

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+MU_DEFINE_ENUM(SRW_LOCK_LL_TRY_ACQUIRE_RESULT, SRW_LOCK_LL_TRY_ACQUIRE_RESULT_VALUES)
+
+MOCKABLE_FUNCTION(, void, srw_lock_ll_init);
```
anporumb, line 19: >init [](http://example.com/codeflow?start=38&length=4)

poetry usually is with "create/destroy"unless it initializes global objects.

I don't see a problem having this as "_create" and return `void`, just saying.

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
+
+/*writer APIs*/
+MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
+MOCKABLE_FUNCTION(, SRW_LOCK_LL_TRY_ACQUIRE_RESULT, srw_lock_ll_try_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
```
anporumb, line 24: >SRW_LOCK_LL* [](http://example.com/codeflow?start=87&length=12)

all these parameters, typed as "SRW_LOCK_LL*" are sort of a "pointer to pointer" to void. Is that intended? (note the `*` in the argument)

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+`srw_lock_ll_init` initializes a slim reader writer lock.
+
+### srw_lock_ll_deinit
+```c
+MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
+```
+
+`srw_lock_ll_deinit` frees deinitializes the slim reader writer lock.
```
anporumb, line 45: >frees deinitializes [](http://example.com/codeflow?start=21&length=19)

either frees or de-initializes (or de-constructs) but not both :)

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+### srw_lock_ll_acquire_exclusive
+```c
+MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
+```
+
+`srw_lock_ll_acquire_exclusive` acquires the slim reader writer lock in exclusive (writer) mode.
```
anporumb, line 52: > exclusive (writer) mode [](http://example.com/codeflow?start=71&length=24)

would be nice to have some REferences to SWR_LOCK (the real one in windows documentation) that would explain what "exclusive" / "shared" mode mean.

interfaces/devdoc/srw_lock_ll_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+`srw_lock_ll_release_exclusive` releases the underlying slim reader writer lock from exclusive (write) mode.
+
+### srw_lock_ll_acquire_shared
+```c
+MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_shared, SRW_LOCK_LL*, srw_lock_ll);
+```
+
+`srw_lock_ll_acquire_shared` acquires the SRWLOCK in shared (read) mode.
```
anporumb, line 74: >SRWLOCK [](http://example.com/codeflow?start=42&length=7)

`SRW_LOCK_LL`?

interfaces/inc/c_pal/srw_lock_ll.h
```diff
@@ -0,0 +1,46 @@
...
+typedef struct SRW_LOCK_HANDLE_DATA_TAG* SRW_LOCK_HANDLE;
+
+#define SRW_LOCK_TRY_ACQUIRE_RESULT_VALUES \
+    SRW_LOCK_TRY_ACQUIRE_OK, \
+    SRW_LOCK_TRY_ACQUIRE_COULD_NOT_ACQUIRE, \
+    SRW_LOCK_TRY_ACQUIRE_INVALID_ARGS
+
+MU_DEFINE_ENUM(SRW_LOCK_TRY_ACQUIRE_RESULT, SRW_LOCK_TRY_ACQUIRE_RESULT_VALUES)
+
+MOCKABLE_FUNCTION(, SRW_LOCK_HANDLE, srw_lock_create, bool, do_statistics, const char*, lock_name);
```
anporumb, line 28: >bool, do_statistics, [](http://example.com/codeflow?start=54&length=20)

this parameter is not part of `srw_lock_ll_requirements.md`

anporumb: I am confused why this file exists in this PR.

interfaces/inc/c_pal/srw_lock_ll.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#ifndef SRW_LOCK_H
```
anporumb, line 3: >SRW_LOCK_H [](http://example.com/codeflow?start=8&length=10)

name of the file is SRW_LOCK_LL.h

interfaces/inc/c_pal/srw_lock_ll.h
anporumb: I think this is part of the interface... cause I see it mentioned in cmakelists.txt of "interfaces" folder.

Still puzzled./

win32/devdoc/srw_lock_ll_win32_requirements.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Exposed API
+
+```c
+#define SRW_LOCK_LL_TRY_ACQUIRE_RESULT_VALUES \
+    SRW_LOCK_LL_TRY_ACQUIRE_OK, \
+    SRW_LOCK_LL_TRY_ACQUIRE_COULD_NOT_ACQUIRE, \
+    SRW_LOCK_LL_TRY_ACQUIRE_INVALID_ARGS
+
+typedef void* SRW_LOCK_LL;
```
anporumb, line 15: >void* [](http://example.com/codeflow?start=8&length=5)

Suppose many of the comments from srw_lock_ll_requirements.md apply here too, so Iw ill try to only mention the new ones.

win32/devdoc/srw_lock_ll_win32_requirements.md
```diff
@@ -0,0 +1,127 @@
...
+If `srw_lock_ll` is `NULL`, `srw_lock_ll_init` shall return.
+
+Otherwise, `srw_lock_ll_init` shall call `InitializeSRWLock`.
+
+### srw_lock_ll_deinit
+```c
+MOCKABLE_FUNCTION(, void, srw_lock_ll_deinit, SRW_LOCK_LL*, srw_lock_ll);
+```
+
+`srw_lock_ll_deinit` frees deinitializes the slim reader writer lock.
```
anporumb, line 49: >frees deinitializes [](http://example.com/codeflow?start=21&length=19)

copy&paste detected.

win32/devdoc/srw_lock_ll_win32_requirements.md
anporumb, line 15: >void [](http://example.com/codeflow?start=8&length=4)

when we are in Windows land here, `SRWLOCK` should appear someplace in this file... as an explicit type.

win32/devdoc/srw_lock_ll_win32_requirements.md
```diff
@@ -0,0 +1,127 @@
...
+Otherwise, `srw_lock_ll_deinit` shall return.
+
+### srw_lock_ll_acquire_exclusive
+```c
+MOCKABLE_FUNCTION(, void, srw_lock_ll_acquire_exclusive, SRW_LOCK_LL*, srw_lock_ll);
+```
+
+`srw_lock_ll_acquire_exclusive` acquires the slim reader writer lock in exclusive (writer) mode.
+
+If `srw_lock_ll` is `NULL` then `srw_lock_ll_acquire_exclusive` shall return.
```
anporumb, line 62: >shall return [](http://example.com/codeflow?start=64&length=12)

I do have to question here the decision to "return". Upon returning, the calling thread will rightfully believe that it have ownership over some data, while in reality, it does not.

What would happen if it would eb specced as "shall not return"?

anporumb (approved):
:shipit:

interfaces/devdoc/srw_lock_ll_requirements.md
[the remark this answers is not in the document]
dcristoloveanu: Fixed.

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: lol, removed frees.

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: Changed to slim reader writer lock as it is like that in the other places also.

interfaces/inc/c_pal/srw_lock_ll.h
dcristoloveanu: It was supposed to have the APIs in the .md, but someone f-ed up and did not put them in here. Fixed.

win32/devdoc/srw_lock_ll_win32_requirements.md
dcristoloveanu: Fixed.

interfaces/inc/c_pal/srw_lock_ll.h
dcristoloveanu: Fixed.

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: Added reference. I thought of adding one to wikipedia too :-), but stuck with the microsoft one only.

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: I think init/deinit fits better with the current terminology we're using. When we would like to initialize the memory provided by the caller we usually use init/deinit. DList as example does the same.

If you feel very strongly I can rename to create/destroy but I think we never had issues with poetry ...

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: This is part of me trying to make too generic the type used. I was aiming for having our interface in the PAL be generic, but honestly in this case it is probably better to have everything generic except the type.

I was banking on the fact that SRWLOCK behind uses a PVOID to store its data. But after all, I changed my mind and simply did

typedef SRWLOCK SRW_LOCK_LL;

When it comes to Linux we'll deal with it.

interfaces/devdoc/srw_lock_ll_requirements.md
dcristoloveanu: Changed to typedef 

#ifdef WIN32
typedef SRWLOCK SRW_LOCK_LL;
#endif

win32/devdoc/srw_lock_ll_win32_requirements.md
dcristoloveanu: Myeah, I went back and forth with this with myself. This is why I needed the review actually and I am glad it got picked on.

I made SRL_LOCK_LL now 

#ifdef WIN32
typedef SRWLOCK SRW_LOCK_LL;
#endif

dcristoloveanu: "Many of the comments" applied to both

win32/devdoc/srw_lock_ll_win32_requirements.md
dcristoloveanu: After some tinkering I looked at what we use for Linux and decided will make this return an int as the API we call for pthreads can fail.

And the caller can decide whether to ignore or not this error but it will have the information.

PR #251, opened 2023-06-29
File util linux spec

samrach9:


interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+
+### CreateFileA
+
+```c
+MOCKABLE_FUNCTION(, HANDLE, CreateFileA, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
+```
+
+`CreateFileA` implements the creating file function in Linux. Uses the `open` function in C for Linux.
+
+**SRS_FILE_UTIL_LINUX_09_001: [** If the `full_file_name` input is either empty or `NULL`, `CreateFileA` shall return an `INVALID_HANDLE_VALUE`. **]**
```
parth21999, line 27: >INVALID_HANDLE_VALUE [](http://example.com/codeflow?start=122&length=20)

I see this value is defined in the .h file. Please include it in this file so we know which value you are referring to. Space is not that constrained :)

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_002: [** `CreateFileA` shall allocate memory for the file handle and on success return a non-`NULL` value.  **]**
```
parth21999, line 29: >`CreateFileA` shall allocate memory for the file handle and on success return a non-`NULL` value. [](http://example.com/codeflow?start=34&length=97)

These can be two separate requirements:

`CreateFileA` shall allocate memory for the file handle.
`CreateFileA` shall succeed and return a non-`NULL` value.

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_015: [** If `creation_disposition` is `CREATE_NEW`, `CreateFileA` will use `O_CREAT|O_EXCEL` on `open` and shall return a new file handle if the file doesn't already exist.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_016: [** If `creation_disposition` is `CREATE_NEW` and the file already exists, `CreateFileA` will fail and return `NULL`.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_017: [** If creation_disposition is `TRUNCATE_EXISTING`, `CreateFileA` will use `O_TRUNC` on `open` and shall return a file handle who's size has been truncated to zero bytes.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_007: [** If `open` returns a -1, `CreateFileA` shall free memory and return an `INVALID_HANDLE_VALUE`.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_008: [** If any error occurs, `CreateFileA` shall fail and return `INVALID_HANDLE_VALUE`.  **]**
```
parth21999, line 49: Usually we have one requirement to handle failures:

If there are any failures, `CreateFileA` shall fail and return `INVALID_HANDLE_VALUE`

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_003: [** If memory allocation for `result` fails, `CreateFileA` shall return a `NULL` value. **]**
+
+**SRS_FILE_UTIL_LINUX_09_004: [** If `dwDesiredAccess` is `GENERIC_READ`, `CreateFileA` will use `O_RDONLY` on `open` and shall return a file handle for read only.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_005: [** If `dwDesiredAcces` is `GENERIC_WRITE`, `CreateFileA` will use `O_WRONLY` on `open` and shall return a file handle for write only.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_006: [** If `dwDesiredAccess` is `GENERIC_ALL` or `GENERIC_READ&GENERIC_WRITE`, `CreateFileA` will use `O_RDWR` on `open` and shall return a file handle for read and write.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_014: [** If `creation_disposition` is `CREATE_ALWAYS` or `OPEN_ALWAYS`, `CreateFileA` will use `O_CREAT` on `open` and shall either create a new file handle if the specificied pathname exists and return it or return an existing file handle.  **]**
```
parth21999, line 39: >CreateFileA` will use `O_CREAT` on `open` [](http://example.com/codeflow?start=98&length=41)

I think a better way to phrase this is:

..., `CreateFileA` shall call `open` with `(insert param_name)` as `O_CREATE` and ...

Here and everywhere

interfaces/devdoc/file_util_linux.md
parth21999: It would be nice to know all values HANDLE can take.

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+MOCKABLE_FUNCTION(, HANDLE, CreateFileA, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
+```
+
+`CreateFileA` implements the creating file function in Linux. Uses the `open` function in C for Linux.
+
+**SRS_FILE_UTIL_LINUX_09_001: [** If the `full_file_name` input is either empty or `NULL`, `CreateFileA` shall return an `INVALID_HANDLE_VALUE`. **]**
+
+**SRS_FILE_UTIL_LINUX_09_002: [** `CreateFileA` shall allocate memory for the file handle and on success return a non-`NULL` value.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_003: [** If memory allocation for `result` fails, `CreateFileA` shall return a `NULL` value. **]**
```
parth21999, line 31: >`NULL` [](http://example.com/codeflow?start=104&length=6)

Reading these specs, it seems like there are two failure modes: `NULL` and `INVALID_HANDLE_VALUE`. 

The .h file defines `INVALID_HANDLE_VALUE` to be NULL. 

Is that intended? How can a user differentiate between these two error modes? Does the user need to differentiate between them?

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+**SRS_FILE_UTIL_LINUX_09_009: [** If `handle_input` is `NULL`, `CloseHandle` returns false.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_011: [** ``CloseHandle` passes value to `close`, sets success to 0, and returns true.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_012: [** If `close` returns a non-zero integer, `CloseHandle` returns false.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_013: [** Otherwise `CloseHandle` will return true.  **]**
+
+
+
```
parth21999, line 69: Nit: whitespace

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+```c
+MOCKABLE_FUNCTION(, bool, CloseHandle, HANDLE, handle_input);
+```
+
+`CloseHandle` implements the close handle function in Linux. Uses the `close` function in C for Linux.
+
+**SRS_FILE_UTIL_LINUX_09_009: [** If `handle_input` is `NULL`, `CloseHandle` returns false.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_011: [** ``CloseHandle` passes value to `close`, sets success to 0, and returns true.  **]**
```
parth21999, line 62: >`` [](http://example.com/codeflow?start=34&length=2)

two backticks

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_012: [** If `close` returns a non-zero integer, `CloseHandle` returns false.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_013: [** Otherwise `CloseHandle` will return true.  **]**
```
parth21999, line 66: The voice of the requirements should be like this:

If blah blah blah happens, `CloseHandle` **shall** do blah blah blah

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,69 @@
...
+**SRS_FILE_UTIL_LINUX_09_008: [** If any error occurs, `CreateFileA` shall fail and return `INVALID_HANDLE_VALUE`.  **]**
+
+
+### CloseHandle
+
+```c
+MOCKABLE_FUNCTION(, bool, CloseHandle, HANDLE, handle_input);
+```
+
+`CloseHandle` implements the close handle function in Linux. Uses the `close` function in C for Linux.
```
parth21999, line 58: >implements the close handle function in Linux. [](http://example.com/codeflow?start=14&length=46)

Maybe say "closes a file handle in Linux".

interfaces/devdoc/file_util_linux.md
parth21999, line 62: >passes value to `close` [](http://example.com/codeflow?start=49&length=23)

Value? Success?

parth21999, line 62: >CloseHandle` passes value to `close`, sets success to 0, and returns true [](http://example.com/codeflow?start=36&length=73)

This can be split into multiple requirements.

interfaces/devdoc/file_util_linux.md
samrach9: I included the define for the INVALID_HANDLE_VALUE. But, I can't define the other values HANDLE will take because there would be hundreds.

parth21999 (changes requested):
:clock1:

interfaces/devdoc/file_util_linux.md
samrach9: CreateFileA in windows returns INVALID_HANDE_VALUE for any failures in creating/opening a file handle. So, every time there is an error I will return INVALID_HANDLE_VALUE. I've went back and fixed it so instead of NULL, there is only one failure mode of INVALID_HANDLE_VALUE.

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef WIN32 
+#include "windows.h"
+#define ebs_create_file CreateFileA
+#define ebs_close_handle CloseHandle
+#else
```
mattdurak, line 22: Typically in the PAL project, we have the following conventions:

- There is a header in the interfaces/inc/c_pal which has the mockable interface
- The implementation for win32 is in win32/src/ (along with tests and specs in the win32 devdoc and tests folders)
- The implementation for Linux is in linux/src (and there are tests and devdoc's as well).

While I understand the reasoning for this to look very much like the win32 API (simple drop-in replacement is great!) I think we should keep the same convention, so I propose we:

- rename this file to "file_util.h"
- Get rid of the renames to CreateFileA and CloseHandle under "ifdef WIN32" here
- Move "file_util_linux.md" to the linux/devdoc folder
- Implementation for windows in win32/src/file_util.c can just be pass-through functions which call CreateFileA, CloseHandle, etc. That will allow the callers of this to have mockable functions for unit tests.

I don't think this is too much additional refactoring and then the module looks a lot more like our other PAL modules.

I guess having the "ifdef WIN32" part to either include windows.h or define the necessary symbols (GENERIC_READ, GENERIC_WRITE, etc.) is probably the easiest so we can leave that here.

mattdurak: I think that keeping this as a MOCKABLE_FUNCTION on both windows and linux will also make it a lot easier to unit test things that use this (like in bsdl) without having to do something difference in the BSDL tests for windows or linux.

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#define INVALID_FILE_SIZE               ((DWORD)0xFFFFFFFF)
+#define INVALID_SET_FILE_POINTER        ((DWORD)-1)
+#define INVALID_FILE_ATTRIBUTES         ((DWORD)-1)
+
+#define INVALID_HANDLE_VALUE            NULL
+
+typedef void *HANDLE;
+typedef void *LPSECURITY_ATTRIBUTES;
+
+MOCKABLE_FUNCTION(, HANDLE, ebs_create_file, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
```
mattdurak, line 56: >ebs_create_file [](http://example.com/codeflow?start=28&length=15)

I don't think we should leak the "ebs" name into c-pal (which doesn't "know" about block store).

I propose renaming this to match the module name, which is our teams convention anyway. So "file_util_create_file" and "file_util_close_file_handle".

This goes with my other comment of renaming this to "file_util" without "linux" and changing the include guards to "FILE_UTIL_H".

mattdurak, line 56: >lpFileName [](http://example.com/codeflow?start=58&length=10)

nit: while we're renaming things to match the conventions elsewhere in the c-pal project, we should consider converting these names from Hungarian notation to snake_case. So file_name, desired_access, share_mode, etc...

mattdurak, line 56: >unsigned long [](http://example.com/codeflow?start=70&length=13)

Typically we use the "uintNN_t" and "intNN_t" types to specify the size of the integers. I think this is especially important in PAL where "long" may be 32-bit (Windows, certain Linux systems) or 64-bit (most 64-bit Linux distros).

So we should replace "unsigned long" from Windows with "uint32_t" to have the exact same type in all platforms.

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#define FILE_FLAG_DELETE_ON_CLOSE       0x04000000
+#define FILE_FLAG_OPEN_NO_RECALL        0x00100000
+
+#define CREATE_NEW                      1
+#define CREATE_ALWAYS                   2
+#define OPEN_EXISTING                   3
+#define OPEN_ALWAYS                     4
+#define TRUNCATE_EXISTING               5
+
+#define INVALID_FILE_SIZE               ((DWORD)0xFFFFFFFF)
```
mattdurak, line 47: >DWORD [](http://example.com/codeflow?start=42&length=5)

I don't think this will build on Linux, DWORD comes from windows.h. It is really "uint32_t" to be portable.

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#define INVALID_SET_FILE_POINTER        ((DWORD)-1)
+#define INVALID_FILE_ATTRIBUTES         ((DWORD)-1)
+
```
mattdurak, line 50: I noticed you didn't put this in the MD and maybe it would just be cleaner to remove it from the header as well.

What if we had "windows_defines.h" in linux/inc/c_pal?

Then this file could just have:

```
#ifdef WIN32 
#include "windows.h"
#else
#include "windows_defines.h"
#endif
```

/consider

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#ifdef WIN32 
+#include "windows.h"
+#define ebs_create_file CreateFileA
+#define ebs_close_handle CloseHandle
+#else
+#define GENERIC_READ                    0x80000000
+#define GENERIC_WRITE                   0x40000000
+#define GENERIC_ALL                     0x80000000 | 0x40000000
+#define GENERIC_EXECUTE                 (0x20000000L)
+#define DELETE                          (0x00010000L)
```
mattdurak, line 27: Do we need GENERIC_EXECUTE or DELETE? I see the spec checks the READ/WRITE/ALL only...

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_001: [** If the `full_file_name` input is either empty or `NULL`, `ebs_create_file` shall return an `INVALID_HANDLE_VALUE`. **]**
+
+**SRS_FILE_UTIL_LINUX_09_002: [** `ebs_create_file` shall allocate memory for the file handle. ebs_create_file will succeed and return a non-`NULL` value.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_003: [** If memory allocation for `result` fails, `ebs_create_file` shall return an `INVALID_HANDLE_VALUE`. **]**
+
+**SRS_FILE_UTIL_LINUX_09_004: [** If `dwDesiredAccess` is `GENERIC_READ`, `ebs_create_file` will call `open` with `O_RDONLY` and shall return a file handle for read only.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_005: [** If `dwDesiredAcces` is `GENERIC_WRITE`, `ebs_create_file` will call `open` with `O_WRONLY` and shall return a file handle for write only.  **]**
```
mattdurak, line 36: >dwDesiredAcces [](http://example.com/codeflow?start=38&length=14)

dwDesiredAccess

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+typedef void *LPSECURITY_ATTRIBUTES;
+
+MOCKABLE_FUNCTION(, HANDLE, ebs_create_file, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
+MOCKABLE_FUNCTION(, bool, ebs_close_handle, HANDLE, handle_input);
+```
+
+### ebs_create_file
+
+```c
+MOCKABLE_FUNCTION(, HANDLE, ebs_create_file, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
```
mattdurak, line 23: >dwShareMode [](http://example.com/codeflow?start=117&length=11)

this is ignored for Linux? Perhaps we should call that out at least

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_003: [** If memory allocation for `result` fails, `ebs_create_file` shall return an `INVALID_HANDLE_VALUE`. **]**
+
+**SRS_FILE_UTIL_LINUX_09_004: [** If `dwDesiredAccess` is `GENERIC_READ`, `ebs_create_file` will call `open` with `O_RDONLY` and shall return a file handle for read only.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_005: [** If `dwDesiredAcces` is `GENERIC_WRITE`, `ebs_create_file` will call `open` with `O_WRONLY` and shall return a file handle for write only.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_006: [** If `dwDesiredAccess` is `GENERIC_ALL` or `GENERIC_READ&GENERIC_WRITE`, `ebs_create_file` will call `open` with `O_RDWR` and shall return a file handle for read and write.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_014: [** If `creation_disposition` is `CREATE_ALWAYS` or `OPEN_ALWAYS`, `ebs_create_file` will call `open` with `O_CREAT` and shall either create a new file handle if the specificied pathname exists and return it or return an existing file handle.  **]**
```
mattdurak, line 40: >creation_disposition [](http://example.com/codeflow?start=38&length=20)

Doesn't match parameter name (but maybe the parameter should be changed to this anyway)

interfaces/devdoc/file_util_linux.md
mattdurak: Same for lpSecurityAttributes, dwFlagsAndAttributes, and hTemplateFile

interfaces/inc/c_pal/file_util_linux.h
mattdurak: Along with this, typically the "interfaces/devdoc/*.md" file describes the interface in a generic way. Might be nice to just have that MD file which says something for each function, like: "ebs_create_file" is a PAL wrapper for the Win32 API CreateFileA (link to MSDN page).

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+
+**SRS_FILE_UTIL_LINUX_09_015: [** If `creation_disposition` is `CREATE_NEW`, `ebs_create_file` will call `open` with `O_CREAT|O_EXCL` and shall return a new file handle if the file doesn't already exist.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_016: [** If `creation_disposition` is `CREATE_NEW` and the file already exists, `ebs_create_file` will fail and return `INVALID_HANDLE_VALUE`.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_017: [** If creation_disposition is `TRUNCATE_EXISTING`, `ebs_create_file` will call `open` with `O_TRUNC` and shall return a file handle who's size has been truncated to zero bytes.  **]**
+
+**SRS_FILE_UTIL_LINUX_09_007: [** If `open` returns a -1, `ebs_create_file` shall free memory and return an `INVALID_HANDLE_VALUE`.  **]**
```
mattdurak, line 48: >SRS_FILE_UTIL_LINUX_09_007 [](http://example.com/codeflow?start=2&length=26)

It sounds like this is redundant with 09_008 below (or, there is nothing special about "open returns -1" as it is just "any failure"). Sometimes we have specific specs for failures if they are certain return values we handle as error, but in this case I would say that it is equivalent to "open failed".

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+
+```c
+MOCKABLE_FUNCTION(, HANDLE, ebs_create_file, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
+```
+
+`ebs_create_file` implements the creating file function in Linux. Uses the `open` function in C for Linux.
+
+**SRS_FILE_UTIL_LINUX_09_001: [** If the `full_file_name` input is either empty or `NULL`, `ebs_create_file` shall return an `INVALID_HANDLE_VALUE`. **]**
+
+**SRS_FILE_UTIL_LINUX_09_002: [** `ebs_create_file` shall allocate memory for the file handle. ebs_create_file will succeed and return a non-`NULL` value.  **]**
```
mattdurak, line 30: > shall allocate memory [](http://example.com/codeflow?start=51&length=22)

Do we need to allocate memory? "open" returns a file handle, which can be used for other file operations.

In close handle, we just call "close" on the handle.

Similarly, in win32, CreateFileA gives us a file handle that can be used with other API's.

Unless we need to store additional state with the handle, I would think we can just return the result of `open` as our HANDLE.

mattdurak: If we do need to malloc here, then close_handle is missing a spec that says we free it.

interfaces/inc/c_pal/file_util_linux.h
```diff
@@ -0,0 +1,67 @@
...
+#define CREATE_ALWAYS                   2
+#define OPEN_EXISTING                   3
+#define OPEN_ALWAYS                     4
+#define TRUNCATE_EXISTING               5
+
+#define INVALID_FILE_SIZE               ((DWORD)0xFFFFFFFF)
+#define INVALID_SET_FILE_POINTER        ((DWORD)-1)
+#define INVALID_FILE_ATTRIBUTES         ((DWORD)-1)
+
+#define INVALID_HANDLE_VALUE            NULL
```
mattdurak, line 51: >NULL [](http://example.com/codeflow?start=40&length=4)

Presumably we want this to be "-1" on Linux. Since "open" is going to return a file descriptor or "-1" on error, and file descriptor "0" is technically valid (it's a special case for stdin), then it seems better to have the invalid value be a truly invalid file descriptor.

But perhaps NULL (aka 0) will work because again, it's a special case value for stdin and won't be used for files.

interfaces/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ set(pal_interfaces_h_files
     inc/c_pal/arithmetic.h
     inc/c_pal/execution_engine.h
     inc/c_pal/async_socket.h
+    inc/c_pal/file_util_linux.h
```
mattdurak, line 7: >file_util_linux [](http://example.com/codeflow?start=14&length=15)

Typically we have a PR with at least a (mostly) empty .c file that includes the new headers also so we know that the code in the header will compile.

interfaces/devdoc/file_util_linux.md
```diff
@@ -0,0 +1,67 @@
...
+## Exposed API 
+
+```c
+// Most defines removed for space constraints
+#define INVALID_HANDLE_VALUE            NULL
+
+typedef void *HANDLE;
+typedef void *LPSECURITY_ATTRIBUTES;
+
+MOCKABLE_FUNCTION(, HANDLE, ebs_create_file, const char*, lpFileName, unsigned long, dwDesiredAccess, unsigned long, dwShareMode, LPSECURITY_ATTRIBUTES, lpSecurityAttributes, unsigned long, dwCreationDisposition, unsigned long, dwFlagsAndAttributes, HANDLE, hTemplateFile);
```
dcristoloveanu, line 16: >ebs_create_file [](http://example.com/codeflow?start=28&length=15)

Can we have some black magic where the user would write in the bsdl code CreateFile? and we redirect the calls to whatever APIs needed.

That brings the benfit that there are no changes needed in the user layer and also one can actually hit F1 in Visual studio and get help on the APIs right away.

interfaces/devdoc/file_util_linux.md
samrach9: We need to allocate memory to allow for threadpools later on. Even though I haven't written it in the spec yet, the malloc is necessary for that.

interfaces/devdoc/file_util_linux.md
samrach9: Yes! We renamed the functions to ebs_create_file and same thing for close handle so it can redirect into Windows and call win32 API.

interfaces/CMakeLists.txt
samrach9: I will do that next time!

interfaces/devdoc/file_util_linux.md
parth21999: Does this module return any of those hundreds of values? We should have the ones that the module does return listed here. If the only value that the module explicitly returns is `INVALID_HANDLE_VALUE` then we're good as is.

samrach9: INVALID_HANDLE_VALUE is the only value that is explicitly returned by this module.

interfaces/devdoc/file_util_requirements.md
```diff
@@ -0,0 +1,38 @@
...
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, lpFileName, unsigned long, desired_access, unsigned long, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, unsigned long, creation_disposition, unsigned long, flags_and_attributes, HANDLE, hTemplateFile);
+MOCKABLE_FUNCTION(, bool, file_util_close_handle, HANDLE, handle_input);
+```
+
+### file_util_create_file
+
+```c
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, full_file_name, unsigned long, desired_access, unsigned long, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, unsigned long, creation_disposition, unsigned long, flags_and_attributes, HANDLE, h_template_file);
+```
+//share_mode, lp_security_attributes, flags_and_attributes, and h_template_file parameters are ignored in Linux
```
mattdurak, line 25: nit: Consider putting this below the other sentence without the "//" comment (because it's an MD doc so we don't need c-style comments.

interfaces/devdoc/file_util_requirements.md
```diff
@@ -0,0 +1,38 @@
...
+
+`file_util_create_file` is a PAL wrapper for Linux for the CreateFileA function in Windows. Uses the `open` function in C for Linux. 
+    https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#return-value
```
mattdurak, line 28: Consider markdown links like `[CreateFileA]([https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#return-value)`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#return-value)`) here and below

linux/inc/c_pal/windows_defines.h
```diff
@@ -0,0 +1,52 @@
...
+#define INVALID_SET_FILE_POINTER        ((uint32_t)-1)
+#define INVALID_FILE_ATTRIBUTES         ((uint32_t)-1)
+
+#define INVALID_HANDLE_VALUE            ((HANDLE)(int64_t)-1)
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // WINDOWS_DEFINES_H
```
mattdurak, line 52: >_DEFINES_H [](http://example.com/codeflow?start=17&length=10)

nit: should have blank line at end of file

interfaces/inc/c_pal/file_util.h
```diff
@@ -0,0 +1,31 @@
...
+#endif
+
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, lp_file_name, uint32_t, desired_access, uint32_t, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, uint32_t, creation_disposition, uint32_t, flags_and_attributes, HANDLE, h_template_file);
+MOCKABLE_FUNCTION(, bool, file_util_close_handle, HANDLE, handle_input);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // FILE_UTIL_H
```
mattdurak, line 31: >// FILE_UTIL_H [](http://example.com/codeflow?start=7&length=14)

nit: add a blank line at end of file

interfaces/inc/c_pal/file_util.h
```diff
@@ -0,0 +1,31 @@
...
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef WIN32 
+#include "windows.h"
+#else
+#include "c_pal/windows_defines.h"
+#endif
```
mattdurak, line 22: This should be moved up before the include on line 13

linux/inc/c_pal/windows_defines.h
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#ifndef WINDOWS_DEFINES_H
+#define WINDOWS_DEFINES_H
+
+#ifdef __cplusplus
+#include <cstdint>
+#else
+#include <stdbool.h>
```
mattdurak, line 9: stdbool is not needed

linux/inc/c_pal/windows_defines.h
```diff
@@ -0,0 +1,52 @@
...
+#include <stdint.h>
+#endif
+
+#include "umock_c/umock_c_prod.h"
```
mattdurak, line 13: not needed

interfaces/inc/c_pal/file_util.h
```diff
@@ -0,0 +1,31 @@
...
+
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, lp_file_name, uint32_t, desired_access, uint32_t, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, uint32_t, creation_disposition, uint32_t, flags_and_attributes, HANDLE, h_template_file);
```
mattdurak, line 24: >lp_file_name [](http://example.com/codeflow?start=64&length=12)

nit: `file_name`, similarly can remove the other "lp" and "h" prefixes as those are more of the "hungarian notation" used by Windows API's which have a short "type" prefix in variable names.

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -0,0 +1,66 @@
...
+## Exposed API 
+
+```c
+// Most defines removed for space constraints
+#define INVALID_HANDLE_VALUE            NULL
+
+typedef void *HANDLE;
+typedef void *LPSECURITY_ATTRIBUTES;
+
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, lpFileName, unsigned long, desired_access, unsigned long, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, unsigned long, creation_disposition, unsigned long, flags_and_attributes, HANDLE, hTemplateFile);
```
mattdurak, line 16: >lpFileName [](http://example.com/codeflow?start=64&length=10)

should update this to match the header

interfaces/devdoc/file_util_requirements.md
```diff
@@ -0,0 +1,38 @@
...
+## Exposed API 
+
+```c
+// Most defines removed for space constraints
+#define INVALID_HANDLE_VALUE            NULL
+
+typedef void *HANDLE;
+typedef void *LPSECURITY_ATTRIBUTES;
+
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, lpFileName, unsigned long, desired_access, unsigned long, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, unsigned long, creation_disposition, unsigned long, flags_and_attributes, HANDLE, hTemplateFile);
```
mattdurak, line 16: >lpFileName [](http://example.com/codeflow?start=64&length=10)

Should also be updated to match the header

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -0,0 +1,66 @@
...
+MOCKABLE_FUNCTION(, bool, file_util_close_handle, HANDLE, handle_input);
+```
+
+### file_util_create_file
+
+```c
+MOCKABLE_FUNCTION(, HANDLE, file_util_create_file, const char*, full_file_name, unsigned long, desired_access, unsigned long, share_mode, LPSECURITY_ATTRIBUTES, lp_security_attributes, unsigned long, creation_disposition, unsigned long, flags_and_attributes, HANDLE, h_template_file);
+```
+//share_mode, lp_security_attributes, flags_and_attributes, and h_template_file parameters are ignored in Linux
```
mattdurak, line 25: >// [](http://example.com/codeflow?start=0&length=2)

nit: I'd also move this below the other sentence and get rid of the "//"

linux/devdoc/file_util_linux_requirements.md
mattdurak: All of the parameter types and names

mattdurak (approved):
:shipit:

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -0,0 +1,66 @@
+# file_util_linux
+
+## Overview
+
+Linux implementation of the file operation functions.
+
+## Exposed API 
+
+```c
+// Most defines removed for space constraints
```
parth21999, line 10: Since INVALID_HANDLE_VALUE is the only value returned by the module, you can remove this comment to avoid confusion.

parth21999 (approved):
:shipit:

PR #255, opened 2023-07-13
Add linux compiler checks

jebrando:


common/tests/refcount_ut/refcount_ut.c
```diff
@@ -292,6 +292,7 @@ BEGIN_TEST_SUITE(TEST_SUITE_NAME_FROM_CMAKE)
         POS_HANDLE p, clone_of_p;
         p = Pos_Create(2);
         clone_of_p = Pos_Clone(p);
+        (void)clone_of_p;
```
dcristoloveanu, line 295: >     (void)clone_of_p; [](http://example.com/codeflow?start=3&length=22)

I think that one of the 2 calls below should be to destroy clone_of_p

common/src/call_once.c
```diff
@@ -18,6 +18,12 @@ MU_DEFINE_ENUM_STRINGS(CALL_ONCE_RESULT, CALL_ONCE_RESULT_VALUES);
 /*a weak check to ensure that nobody has the idea to change call_once_t type to int8_t (or other type...  to save some memory for example)*/
 static int check_call_once_t_is_the_same_as_volatile_atomic_int32_t_because_we_are_going_to_pass_it_to_wait_on_address_or_wake_by_address_all[sizeof(volatile_atomic int32_t) == sizeof(call_once_t)];
 
+// Function is to avoid Warning -Werror=unused-variable in gcc
+static void unused_var_function(void)
+{
+    (void)check_call_once_t_is_the_same_as_volatile_atomic_int32_t_because_we_are_going_to_pass_it_to_wait_on_address_or_wake_by_address_all;
+}
```
dcristoloveanu, line 25: If we make the above const doesn't it achieve the same and no warning?

dcristoloveanu:
> 

If you make the above const or volatile doesn't that avoid the warning?

---
Refers to: linux/linux_reals/real_execution_engine_linux.c:16 in a5967d3. [](commit_id = a5967d39ebf8295fb99c88be92697750b016e886, deletion_comment = False)

linux/src/uuid_linux.c
```diff
@@ -13,6 +13,12 @@
 
 static int is_UUID_T_and_UUID_same_size[sizeof(UUID_T) == sizeof(uuid_t)]; /*just a sanity check*/
 
+// Function is to avoid Warning -Werror=unused-variable in gcc
+static void unused_var_function(void)
+{
+    (void)is_UUID_T_and_UUID_same_size;
+}
+
 int uuid_produce(UUID_T destination)
```
dcristoloveanu, line 22: Same comment about "if we make it const or volatile.

dcristoloveanu (approved):
:shipit:

linux/src/file_util_linux.c
```diff
@@ -68,6 +66,10 @@ HANDLE file_util_open_file(const char* full_file_name, uint32_t access, uint32_t
                 /*Codes_SRS_FILE_UTIL_LINUX_09_006: [ If desired_access is GENERIC_ALL or GENERIC_READ&GENERIC_WRITE, file_util_open_file shall call open with O_RDWR and shall return a file handle for read and write. ]*/
                 user_access = O_RDWR;
             }
+            else
+            {
+                user_access = 0;
+            }
```
dcristoloveanu, line 72: Hah, some real ones

common/src/call_once.c
jebrando: const doesn't quiet the warning, but I have another way to fix it.

common/tests/refcount_ut/refcount_ut.c
jebrando: valgrind says close_of_p should not be destroyed.

PR #256, opened 2023-07-13
error_handling_linux_spec

samrach9:
error handling functions for linux implementation. Spec, header, and windows error defines files


interfaces/inc/c_pal/error_handling.h
```diff
@@ -0,0 +1,31 @@
...
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef WIN32 
+#include "windows.h"
+#else
+#include "c_pal/windows_defines.h"
+#include "c_pal/windows_defines_errors.h"
```
jebrando, line 22: Not needed, since you don't have any windows defines in here you don't need these here either.

interfaces/inc/c_pal/error_handling.h
```diff
@@ -0,0 +1,31 @@
...
+
+MOCKABLE_FUNCTION(, void, error_handling_set_last_error, uint32_t, err_code);
+MOCKABLE_FUNCTION(, uint32_t, error_handling_get_last_error);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // ERROR_HANDLING_H
```
jebrando, line 31: Nitpick: always try to add a newline at the end of your files.  I believe it goes back to compilers failing if there wasn't one

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,41 @@
+# error handling
+
+## Overview
+
+Linux implentaton of GetLastError and SetLastError 
+
+## Exposed API
+
+```c
+uint64_t last_error_code;
```
jebrando, line 10: I'd remove this from here and the other spec file.  technically you could decided to keep the last_error_code in a lot of different containers so you don't have to advertise that here.

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### set_last_error
+
+```c
+MOCKABLE_FUNCTION(, void, error_handling_set_last_error, uint32_t, err_code);
+```
+`error_handling_set_last_error` implements the PAL wrapper for the SetLastError function from Windows.
+
+`error_handling_set_last_error` shall succeed and assign a non-`NULL` value to `last_error_code`.
```
jebrando, line 23: >succeed and [](http://example.com/codeflow?start=38&length=11)

Remove the words "succeed and".  Since this returns void there is no succeeding or failing, it just does some work.

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,41 @@
...
+
+`error_handling_set_last_error` shall call `interlocked_exchange_64` with `err_code` and `last_error_code`.
+
+
+If the `err_code` is `NULL`, `error_handling_set_last_error` shall not change `last_error_code`.
```
jebrando, line 28: Since err_code isn't a pointer it can not be NULL (0 is a valid value btw).  You can just remove this line all together.

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,41 @@
...
+
+### get_last_error
+
+```c
+MOCKABLE_FUNCTION(, uint32_t, error_handling_get_last_error);
+```
+`error_handling_get_last_error` implements the PAL wrapper for the GetLastError function from Windows.
+
+On success, `error_handling_get_last_error` shall return a non-`NULL` value.
```
jebrando, line 37: Since uint32_t is not a pointer, it can not be NULL.  Probably should say something like 

`error_handling_get_last_error` shall return the value last set through set_last_error or 0 (0 means everything is OK).

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,41 @@
...
+
+`error_handling_get_last_error` shall call `interlocked_add` with `last_error_code` and zero.
+
+If there are any errors, `error_handling_get_last_error` shall fail and return `NULL`.
```
jebrando, line 41: >If there are any errors, `error_handling_get_last_error` shall fail and return `NULL` [](http://example.com/codeflow?start=0&length=85)

This one is tricky, error_handling_get_last_error can't return NULL because of the pointer thing.  If interlock_add fails we probably should return something specific.  Maybe even crash.

jebrando:
> // ****** End of OBSOLETE functions.

Probably should remove the OBSOLETE functions.  We don't need them.

---
Refers to: linux/inc/c_pal/windows_defines_errors.h:29549 in 420232d. [](commit_id = 420232d195fd9fb6c848e2b12173a3ba22bd7904, deletion_comment = False)

linux/src/error_handling.c
```diff
@@ -0,0 +1,26 @@
...
+#include "c_pal/error_handling.h"
+
+uint32_t last_error_code;
+
+void error_handling_set_last_error(uint32_t err_code)
+{
+
+}
+
+uint32_t error_handling_get_last_error()
```
jebrando, line 23: >() [](http://example.com/codeflow?start=38&length=2)

Add a void here so it looks like this:

uint32_t error_handling_get_last_error(void)

linux/src/error_handling.c
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include <stddef.h>
+#include <inttypes.h>
+
+#include "c_pal/gballoc_hl.h" // IWYU pragma: keep
+#include "c_pal/gballoc_hl_redirect.h"
```
jebrando, line 7: Won't need this I don't think.

jebrando:
> // Copyright (C) Microsoft Corporation. All rights reserved.

Maybe we can talk about removing some of these errors that we never will use.  Let's talk.

---
Refers to: linux/inc/c_pal/windows_defines_errors.h:1 in 420232d. [](commit_id = 420232d195fd9fb6c848e2b12173a3ba22bd7904, deletion_comment = False)

linux/src/error_handling.c
```diff
@@ -0,0 +1,26 @@
...
+
+#include "c_logging/logger.h"
+
+#include "c_pal/windows_defines.h"
+#include "c_pal/windows_defines_errors.h"
+
+#include "c_pal/error_handling.h"
+
+uint32_t last_error_code;
```
jebrando, line 16: This has to be a static atomic uint64_t last_error_code.

jebrando (changes requested):
:clock1:

linux/devdoc/error_handling_requirements.md
samrach9: Could we just return nothing and then log an error? Like do return ;

interfaces/inc/c_pal/error_handling.h
jebrando: you don't need the windows.h either

interfaces/inc/c_pal/error_handling.h
```diff
@@ -0,0 +1,29 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef WIN32 
+#include "windows.h"
+#else
+
+MOCKABLE_FUNCTION(, void, error_handling_set_last_error, atomic_uint64_t, err_code);
+MOCKABLE_FUNCTION(, atomic_uint64_t, error_handling_get_last_error);
```
jebrando, line 23: >atomic_uint64_t [](http://example.com/codeflow?start=20&length=15)

This one should just be uint64_t

linux/src/error_handling.c
```diff
@@ -0,0 +1,25 @@
...
+#include "c_pal/gballoc_hl.h" // IWYU pragma: keep
+
+#include "c_logging/logger.h"
+
+#include "c_pal/windows_defines.h"
+#include "c_pal/windows_defines_errors.h"
+
+#include "c_pal/error_handling.h"
+
+static atomic_uint64_t last_error_code;
```
jebrando, line 15: >atomic_uint64_t [](http://example.com/codeflow?start=7&length=15)

static volatile_atomic int64_t

jebrando (approved):
:shipit:

linux/devdoc/error_handling_requirements.md
```diff
@@ -0,0 +1,36 @@
+# error handling
+
+## Overview
+
+Linux implentaton of GetLastError and SetLastError 
+
+## Exposed API
+
+```c
+MOCKABLE_FUNCTION(, void, error_handling_set_last_error,  volatile_atomic int64_t, err_code);
```
samrach9, line 10: >  [](http://example.com/codeflow?start=57&length=1)

Already took out this extra space.

PR #258, opened 2023-07-21
file util other functions spec and header

samrach9:


linux/devdoc/file_util_linux_requirements.md
```diff
@@ -64,3 +68,64 @@ MOCKABLE_FUNCTION(, bool, file_util_close_file, HANDLE, handle_input);
...
+
+`file_util_get_size_ex` shall call `stat` on `handle_input->full_file_name` and shall return the non-zero file size.
+
+### file_util_set_file_pointer_ex
+
+```c
+MOCKABLE_FUNCTION(, bool, file_util_set_file_pointer_ex, HANDLE, handle_input, LARGE_INTEGER, distance_to_move, PLARGE_INTEGER, new_file_pointer, uint32_t, move_method);
+```
+
+If there are any failures, `file_util_set_file_pointer_ex` shall fail and return false.
```
mattdurak, line 93: missing the description sentence, here and below

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -64,3 +68,64 @@ MOCKABLE_FUNCTION(, bool, file_util_close_file, HANDLE, handle_input);
...
+
+`file_util_set_end_of_file` shall call `fopen on `handle_input->full_file_name` to get the file stream.
+
+`file_util_set_end_of_file` shall call `ftell` on the input file stream and get the number of bytes to extend or truncate by.
+
+`file_util_set_end_of_file` shall call `truncate` on `handle_input->full_file_name` and `num_bytes`.
+
+`file_util_set_end_of_file` shall succeed and return true.
+
+### file_util_write_file_ex
```
mattdurak, line 121: >file_util_write_file_ex [](http://example.com/codeflow?start=4&length=23)

Don't we need an API which matches the usage of bsdl_file_ll.c in block store? As in an asynchronous IO? How will this function be used in block store?

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -64,3 +68,64 @@ MOCKABLE_FUNCTION(, bool, file_util_close_file, HANDLE, handle_input);
 
 **SRS_FILE_UTIL_LINUX_09_019: [** `file_util_close_file` shall call `close` on the given handle_input. **]**
 
+### file_util_get_file_size_ex
+
+```c
+MOCKABLE_FUNCTION(, bool, file_util_get_file_size, PLARGE_INTEGER*, file_size, HANDLE, handle_input);
+```
+
+`file_util_get_file_size_ex` implements the GetFileSizeEx function in Linux.
```
mattdurak, line 77: >GetFileSizeEx [](http://example.com/codeflow?start=44&length=13)

backticks

mattdurak, line 77: >in Linux. [](http://example.com/codeflow?start=67&length=9)

seems like above we mention what Linux API's we call to implement these, could be nice for these functions as well for reading at a glance.

linux/devdoc/file_util_linux_requirements.md
```diff
@@ -64,3 +68,64 @@ MOCKABLE_FUNCTION(, bool, file_util_close_file, HANDLE, handle_input);
...
+
+If the file handle does not have the `GENERIC_WRITE` or `GENERIC_READ` access, `file_util_get_file_size_ex` shall fail and return false.
+
+If there are any failures, `file_util_get_file_size_ex` shall fail and return false.
+
+`file_util_get_file_size_ex` shall succeed and return true.
```
mattdurak, line 83: nit: should go after the "shall call `stat`..." line

linux/devdoc/file_util_linux_requirements.md
samrach9: We decided to implement a CreateThreadpoolIO. Then the write file function would go within that. So that is something I will be working on implementing now.

mattdurak: Yeah that makes sense, I suppose this will be updated with some additional parameter which is an `OVERLAPPED*` on win32 and something equivalent on Linux

PR #270, opened 2023-09-25
Add async socket notify

avranju:


linux/src/async_socket_linux.c
```diff
@@ -711,3 +817,78 @@ int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BU
...
+all_ok:
+            if (interlocked_decrement(&async_socket->pending_api_calls) == 0)
+            {
+                wake_by_address_single(&async_socket->pending_api_calls);
+            }
+        }
+    }
+
+    return result;
+}
```
jebrando, line 894: Nitpick: newline here.

interfaces/inc/c_pal/async_socket.h
```diff
@@ -58,12 +75,22 @@ typedef struct ASYNC_SOCKET_BUFFER_TAG
...
 MOCKABLE_FUNCTION(, int, async_socket_receive_async, ASYNC_SOCKET_HANDLE, async_socket, ASYNC_SOCKET_BUFFER*, payload, uint32_t, buffer_count, ON_ASYNC_SOCKET_RECEIVE_COMPLETE, on_receive_complete, void*, on_receive_complete_context);
+MOCKABLE_FUNCTION(, int, async_socket_notify_io_async, ASYNC_SOCKET_HANDLE, async_socket, ASYNC_SOCKET_NOTIFY_IO_TYPE, io_type, ON_ASYNC_SOCKET_NOTIFY_IO_COMPLETE, on_notify_io_complete, void*, on_notify_io_complete_context);
+
+#ifndef WIN32
+#ifdef TEST_SUITE_NAME_FROM_CMAKE
+// If this is a CMake test build, then expose the following functions for testing purposes.
+ON_ASYNC_SOCKET_SEND get_async_socket_send_callback();
+ON_ASYNC_SOCKET_RECV get_async_socket_recv_callback();
+#endif
+#endif
```
jebrando, line 93: I was looking for a way to avoid having these function here and I would make an argument that this is not needed.  The functions are internal and so we can guarantee that the values sent to these function will never be NULL so you're testing an impossible situation.  I would remove these and the test that refer to them.

linux/src/async_socket_linux.c
```diff
@@ -711,3 +817,78 @@ int async_socket_receive_async(ASYNC_SOCKET_HANDLE async_socket, ASYNC_SOCKET_BU
...
+                result = MU_FAILURE;
+            }
+            else
+            {
+                io_context->io_type = ASYNC_SOCKET_IO_TYPE_NOTIFY;
+                io_context->on_notify_io_complete = on_notify_io_complete;
+                io_context->callback_context = on_notify_io_complete_context;
+                io_context->async_socket = async_socket;
+
+                int epoll_op = (io_type == ASYNC_SOCKET_NOTIFY_IO_TYPE_IN) ? EPOLLIN : EPOLLOUT;
```
jebrando, line 865: >int epoll_op = (io_type == ASYNC_SOCKET_NOTIFY_IO_TYPE_IN) ? EPOLLIN : EPOLLOUT; [](http://example.com/codeflow?start=16&length=80)

This may need to change due to our talk today, maybe these will be an or of both values.

jebrando (approved):
:shipit:

interfaces/inc/c_pal/async_socket.h
avranju: :) Yeah. I don't have strong feelings about this. I'll remove the functions, test and requirements.

linux/src/async_socket_linux.c
avranju: Added a couple of tasks last week after our discussion (25319277 & 25319272). If this really is a race as we suspect it might be then the fix needs to happen here and, in the send/receive flows.

linux/src/async_socket_linux.c
avranju: Nit has been picked.

PR #272, opened 2023-10-02
Add single_performance_counter to c-pal.

darobs:
`single_performance_counter` is a unit which samples a single performance counter. Behavior is similar to [ .NET `PerformanceCounter`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.performancecounter?view=dotnet-plat-ext-7.0) 

Currently, the implementation is Windows only, Linux OS implementation is stubbed to return `NULL` and `SINGLE_PERFORMANCE_COUNTER_SAMPLE_ERROR` as appropriate.


mattdurak (approved):
:shipit:

PR #273, opened 2023-10-03
Update dependencies

dcristoloveanu:


anporumb:
>     (void)flags_and_attributes;

not needed cause it is used in line 34.

---
Refers to: linux/src/file_util_linux.c:29 in b29d73a. [](commit_id = b29d73ab56ea6cf3a2b3bedb5f0943688f76c6aa, deletion_comment = False)

anporumb (approved):
:shipit:

PR #279, opened 2023-10-13
fix threadpool_linux open to initialize variables

jebrando:


linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -569,4 +569,53 @@ TEST_FUNCTION(cancel_timer_waits_for_ongoing_execution)
...
+    threadpool_close(threadpool);
+
+    (void)interlocked_exchange(&thread_counter, 0);
+
+    // Reopen the threadpool
+    ASSERT_ARE_EQUAL(int, 0, threadpool_open(threadpool));
+
+    // Create 1 thread pool
+    LogInfo("Scheduling work item");
+    ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work(threadpool, threadpool_task_wait_20_millisec, (void*)&thread_counter));
```
avranju, line 604: Noob question: Why would we be able to schedule work on a thread pool _after_ it has been closed?

avranju (approved):

linux/tests/threadpool_linux_int/threadpool_linux_int.c
avranju: Never mind. I see that the thread pool has been opened again before scheduling more work.

linux/src/threadpool_linux.c
```diff
@@ -199,7 +197,7 @@ static int threadpool_work_func(void* param)
                 }
             }
         /* Codes_SRS_THREADPOOL_LINUX_07_085: [ threadpool_work_func shall loop until threadpool_close or threadpool_destroy is called. ]*/
```
dcristoloveanu, line 199: Uhm, this is talking about close, but we changed the names to start/stop. i think we should not rename to start/stop frankly. Just makes more work....

dcristoloveanu:
>                                 LogError("Failure creating semi_init");

Uuuu, semi!

---
Refers to: linux/src/threadpool_linux.c:376 in 23581b7. [](commit_id = 23581b7a2485b1db44f0862c8716c1dfa81793c1, deletion_comment = False)

linux/src/threadpool_linux.c
```diff
@@ -429,6 +427,17 @@ int threadpool_open(THANDLE(THREADPOOL) threadpool)
         }
         else
         {
+            // Codes_SRS_THREADPOOL_LINUX_11_001: [ threadpool_open shall initialize internal threapool data items ]
+            for (int32_t index = 0; index < threadpool_ptr->task_array_size; index++)
```
dcristoloveanu, line 431: >int32_t [](http://example.com/codeflow?start=17&length=7)

Likely this should be uint32_t

linux/src/threadpool_linux.c
```diff
@@ -379,7 +377,7 @@ THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
                             }
                             else
                             {
-                                (void)interlocked_exchange(&result->state, 0);
+                                (void)interlocked_exchange(&result->state, THREADPOOL_STATE_STARTED);
```
dcristoloveanu, line 380: >sult->state, THREADPOOL_STATE_STARTED); [](http://example.com/codeflow?start=62&length=39)

Honestly I think that this state can only map to a state machine.
Now, the state machine can have states that align to the SM states or not.

The way I read the code it seems that this state machine now is a sub-state machine which maps the OPEN state of SM.

In other words, when we are in open, we can be in one of the 2 sub-states: stopped or started.
If that is the case the state should always be set to STARTED when we open with sm. The sub-state means nothing outside of sm OPEN.

If somehow this state (and state machine applies to other sm states, we kind of have a problem, because then we'd be having 2 state machines running in parallel.

So, I suspect this assignment is wrong (or something deeper exists as an issue here).
Shall we talk?

dcristoloveanu (changes requested):
:clock1:

linux/src/threadpool_linux.c
jebrando: Not really a states machine, it's more of a starting and stopping of the thread.  I'll revert this one and then we can talk about it.

dcristoloveanu:
>         } while (interlocked_add(&threadpool->state, 0) != 1);

Based on in person discussion this seems to be needing some poetry.

---
Refers to: linux/src/threadpool_linux.c:194 in 02346a6. [](commit_id = 02346a6b70ab0ae4a1376fc5415225b2e90ed01d, deletion_comment = False)

dcristoloveanu (approved):
:shipit:

jasmineymlo (approved):

anporumb (approved):
:shipit:

PR #294, opened 2023-12-29
adding header and spec for sm handling opening complete

jebrando:
Adding spec to sm to enable the handling of open_complete callback during the opening state.

### DIFF
--- a/common/devdoc/sm_requirements.md
+++ b/common/devdoc/sm_requirements.md
@@ -99,12 +99,14 @@ MU_DEFINE_ENUM(SM_RESULT, SM_RESULT_VALUES);
 
 typedef void(*ON_SM_CLOSING_COMPLETE_CALLBACK)(void* context);
 typedef void(*ON_SM_CLOSING_WHILE_OPENING_CALLBACK)(void* context);
+typedef void(*ON_ERROR_CALLBACK)(void* context, bool is_opening);
 
 MOCKABLE_FUNCTION(, SM_HANDLE, sm_create, const char*, name);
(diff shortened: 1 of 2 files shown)


common/devdoc/sm_requirements.md
```diff
@@ -179,10 +182,26 @@ MOCKABLE_FUNCTION(, void, sm_open_end, SM_HANDLE, sm, bool, success);
...
 
 **SRS_SM_02_075: [** If `success` is `false` then `sm_open_end` shall switch the state to `SM_CREATED`. **]**
 
+### sm_on_open_complete
+
+```c
+MOCKABLE_FUNCTION(, SM_RESULT, sm_on_open_complete, SM_HANDLE, sm);
+```
+
+`sm_on_open_complete` informs the user that calling on_open_complete is allowed.
```
mattdurak, line 197: Not sure I understand this. How does it "inform the user"?

I guess this is really "sm_is_in_opening" or something because it's checking the current state to make sure it's in OPENING.

But I don't understand how this helps because it's not synchronized with anything else. I assume the intention is to synchronize with `sm_on_error`. But it's entirely possible that this is called and returns GRANTED, then `sm_on_error` is called immediately after.

I think the only way for this to work is for it to behave like `sm_open_end` except it returns GRANTED or REFUSED. If you have the state change here, then it can synchronize with failures.

Similarly, you wouldn't need `sm_on_error` because that's just `sm_open_end` with success set to false.

jebrando: Ok, after a long spirited debate, we decided to go a different more difficult direction.

PR #299, opened 2024-01-29
adding function to determine if a uuid is NIL.

jebrando:


linux/src/uuid_linux.c
```diff
@@ -35,3 +36,45 @@ int uuid_produce(UUID_T destination)
...
+            uuid_value[6] == 0 &&
+            uuid_value[7] == 0 &&
+            uuid_value[8] == 0 &&
+            uuid_value[9] == 0 &&
+            uuid_value[10] == 0 &&
+            uuid_value[11] == 0 &&
+            uuid_value[12] == 0 &&
+            uuid_value[13] == 0 &&
+            uuid_value[14] == 0 &&
+            uuid_value[15] == 0
```
dcristoloveanu, line 67: I assume the compiler will optimize the heck out of these... :-)

win32/tests/uuid_win32_ut/uuid_win32_ut.c
```diff
@@ -318,7 +318,58 @@ TEST_FUNCTION(GUID_from_uuid_succeeds)
...
+    ///arrange
+    UUID_T valid_uuid = { 0 };
+
+    ///act
+    bool result = is_uuid_nil(valid_uuid);
+
+    ///assert
+    ASSERT_IS_TRUE(result);
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
 }
```
dcristoloveanu, line 373: Would be interesting to have a test where we only set one byte on a non-zero value (as in a loop of 16 and set the byte at *i* to be non-zero and check that we return false).

/consider

dcristoloveanu (approved):
:shipit:

linux/src/uuid_linux.c
```diff
@@ -35,3 +36,45 @@ int uuid_produce(UUID_T destination)
...
 }
+
+bool is_uuid_nil(const UUID_T uuid_value)
+{
+    bool result;
+    // Codes_SRS_UUID_LINUX_11_001: [ if uuid_value is NULL then is_uuid_nil shall fail and return false. ]
+    if (uuid_value == NULL)
+    {
+        LogError("invalid argument UUID_T uuid_value=%p", uuid_value);
+        result = false;
```
avranju, line 47: Doesn't this mean that a `NULL` `uuid_value` is not `nil`? From the point of view of the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), should we return `true` here?

win32/src/uuid_win32.c
```diff
@@ -128,3 +130,45 @@ int GUID_from_uuid(GUID* destination, const UUID_T source)
     }
     return result;
 }
+
+bool is_uuid_nil(const UUID_T uuid_value)
```
avranju, line 134: It doesn't seem like this Windows implementation is different from the Linux version. Should we put this in a common platform neutral source file instead (to avoid forgetting to update the _other_ version in case we make some change in the future)?

linux/src/uuid_linux.c
jebrando: I thought about this a lot, even thought about returning an enum.  I get what you're saying.  Do you think I should just return an enum?

win32/src/uuid_win32.c
jebrando: These implementations are different, the functions used to create the UUID are different.  What I did here is the same so the behavior is the same.

avranju (approved):
LGTM

PR #301, opened 2024-02-05
Use real_interlocked in real_threadpool

parth21999:
If a test is mocking `interlocked.h` and using `real_threadpool.h`, `THANDLE` functions cause `interlocked` calls to get recorded.

### DIFF
--- a/win32/reals/real_threadpool.c
+++ b/win32/reals/real_threadpool.c
@@ -3,6 +3,8 @@
 
 #include "real_gballoc_hl_renames.h"
 
+#include "real_interlocked_renames.h"
+
 #include "real_execution_engine_renames.h"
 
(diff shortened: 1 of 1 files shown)


win32/reals/real_threadpool.c
```diff
@@ -3,6 +3,8 @@
 
 #include "real_gballoc_hl_renames.h"
 
+#include "real_interlocked_renames.h"
```
anporumb, line 6: >"real_interlocked_renames.h" [](http://example.com/codeflow?start=9&length=28)

<I suspect>

The amount of work which will need to follow this in order to propagate the change in UTs and everywhere will be... rather large, euphemistically talking.

</I suspect>

anporumb (approved):
:shipit:

PR #303, opened 2024-02-14
Add a macro for volatile int32_t plus enum union

mattdurak:
One-liner for:

```c
    union \
    { \
        volatile_atomic int32_t variable_name; \
        enum_type MU_C2(variable_name, _enum); \
    };
```

Where variable_name is used with interlocked functions and variable_name_enum is used for nicer debugging.

### DIFF
--- a/c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
+++ b/c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
@@ -1,0 +1,26 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef INTERLOCKED_MACROS_H
+#define INTERLOCKED_MACROS_H
+
+#ifdef __cplusplus
+#include <cstdint>
+
+#else
+#include <stdint.h>
+#endif
+
+#include "macro_utils/macro_utils.h"
+
+#include "c_pal/interlocked.h"
+
+/*Codes_SRS_INTERLOCKED_MACROS_42_001: [ INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM shall generate a union with two fields: a volatile_atomic int32_t and a variable of the type enum_type. ]*/
+#define INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM(enum_type, variable_name) \
+    union \
+    { \
+        volatile_atomic int32_t variable_name; \
+        enum_type MU_C2(variable_name, _enum); \
+    }
+
+#endif // INTERLOCKED_MACROS_H
--- a/c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
+++ b/c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
@@ -13,14 +13,13 @@
 
 #include "macro_utils/macro_utils.h"
 
-#include "c_pal/interlocked.h"
-
 /*Codes_SRS_INTERLOCKED_MACROS_42_001: [ INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM shall generate a union with two fields: a volatile_atomic int32_t and a variable of the type enum_type. ]*/
 #define INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM(enum_type, variable_name) \
     union \
     { \
         volatile_atomic int32_t variable_name; \
         enum_type MU_C2(variable_name, _enum); \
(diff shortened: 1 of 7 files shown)


c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
```diff
@@ -0,0 +1,61 @@
...
+// Mostly a case of "if it compiles..."
+/*Tests_SRS_INTERLOCKED_MACROS_42_001: [ INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM shall generate a union with two fields: a volatile_atomic int32_t and a variable of the type enum_type. ]*/
+TEST_FUNCTION(INTERLOCKED_DEFINE_VOLATILE_STATE_ENUM_works_with_some_enum)
+{
+    ///arrange
+    TEST_STRUCT test_struct;
+
+    ///act
+    test_struct.state_enum = MY_ENUM_VALUE_1;
+    (void)interlocked_exchange(&test_struct.state, MY_ENUM_VALUE_2);
```
darobs, line 52: Would it make sense to compare the results of `interlocked_exchange` to the original value?

darobs (approved):
Share and enjoy!

c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
mattdurak: I just added some more asserts to read it later

darobs (dismissed):

c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
```diff
@@ -0,0 +1,65 @@
...
+
+    ///assert
+    size_t size_of_int32_t = sizeof(int32_t);
+    size_t size_of_enum = sizeof(MY_ENUM);
+    size_t expected_size = (size_of_int32_t > size_of_enum ? size_of_int32_t : size_of_enum);
```
anporumb, line 59: I think the only assert here is that  sizeof(enum)==sizeof(int32_t). It would be a VERY weird platform.

When they are no equal, due to endianess we can end up looking at unintended values.

Maybe it can be checked with some static assert in the header directly, like declaring an external array with size equal to "sizeof(enum)==sizeof(int32_t)".

c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
```diff
@@ -0,0 +1,65 @@
...
+    ASSERT_ARE_EQUAL(size_t, expected_size, sizeof(test_struct));
```
anporumb, line 60: >ASSERT_ARE_EQUAL(size_t, expected_size, sizeof(test_struct)); [](http://example.com/codeflow?start=4&length=61)

I don't understand the need to check that the whole size of the struct is made of the size of the union.

c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef INTERLOCKED_MACROS_H
+#define INTERLOCKED_MACROS_H
```
anporumb, line 5: I vote for this header to be included directly in #include "c_pal/interlocked.h" :)

anporumb (approved):
:shipit:

c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
mattdurak: added a static assert to the macro by doing a char[?1:-1] in the union

c_pal_ll/interfaces/inc/c_pal/interlocked_macros.h
mattdurak: sure, one less include to add everywhere else

c_pal_ll/interfaces/tests/interlocked_macros_ut/interlocked_macros_ut.c
mattdurak: Sure, I just added a union by itself to do the sizeof check in the test (as in, this would fail if we generated two fields instead of a union). Removed the extra struct which could possibly have some padding on some platforms.

anporumb (approved):
:shipit:

PR #308, opened 2024-02-27
Add real_thandle_helper.h and real_interlocked_undo_rename.h to c-pal…

darobs:
… repo.

Add a document and unit test to show how to use.


common/devdoc/real_thandle_helper_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+4. In The test suite initialization, call `REGISTER_REAL_THANDLE_MOCK_HOOK` to register the mocked type.
+
+```c
+    REGISTER_REAL_THANDLE_MOCK_HOOK(MOCKED_STRUCT);
+```
+
+Now, calls in the code to `THANDLE` functions will be redirected to the mocked type.
+
```
mattdurak, line 58: calls are already mocked and redirected to the mocks by default. This will essentially hook up reals for all of the `THANDLE_X(TYPE)` functions.

common/reals/real_thandle_helper.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "c_pal/thandle.h"
+#include "c_pal/thandle_ll.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "real_gballoc_hl.h"
+
+// Creates a THANDLE alias of TYPE called REAL_TYPE
+// Must have ENABLE_MOCKS undefined
+// Recommend including real_interlocked_renames.h before this call and real_interlocked_undo_renames.h after this call
```
mattdurak, line 16: >ecommend including real_interlocked_renames.h before this call and real_interlocked_undo_renames.h after this call [](http://example.com/codeflow?start=4&length=114)

I think we should have this in the MD as a note and as part of the example code there

c_pal_ll/linux/linux_reals/real_interlocked_undo_rename.h
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// Intentionally leave out include guard
+// This file may be included after real_interlocked_renames.h in order to undo the rename
+// For example, to use reals for a THANDLE definition but mocks after that
+
+#undef interlocked_add
```
mattdurak, line 7: >interlocked_add [](http://example.com/codeflow?start=7&length=15)

let's put this file in c_pal_ll/interfaces/reals so we only have 1 copy of it

common/tests/real_thandle_helper_ut/real_thandle_helper_ut.c
```diff
@@ -0,0 +1,111 @@
...
+typedef struct MOCKED_STRUCT_TAG
+{
+    uint8_t dummy;
+} MOCKED_STRUCT;
+
+REAL_THANDLE_DECLARE(MOCKED_STRUCT);
+
+REAL_THANDLE_DEFINE(MOCKED_STRUCT);
+
+#include "real_interlocked_undo_rename.h" // IWYU pragma: keep
```
mattdurak, line 46: >real_interlocked_undo_rename [](http://example.com/codeflow?start=10&length=28)

consider a test that checks the normal `umock_c_get_expected_calls` mock things to make sure the interlocked functions generated by THANDLE are ignored.

I guess you could get even fancier and actually call an interlocked function (so you expect the one you call manually to be recorded by the mocking but the calls internal to THANDLE are not).

mattdurak (dismissed):
:shipit:

mattdurak (approved):
:shipit:

PR #313, opened 2024-03-08
Fix async_socket on linux

mattdurak:
- Bring logging to parity with win32 version (LogError on all disconnect cases, have verbose logging support)
- Fix pending call count ordering (this is why we should have sm everywhere...)
- Add an MD to the interfaces/devdoc folder for consistency
- Document that the send/recv callbacks, if provided, must return -1 for all failures and use errno to report the reason code
- Remove unused .c file


dcristoloveanu (approved):
:shipit:

PR #314, opened 2024-03-11
Jemalloc vcpkg

dcristoloveanu:


avranju (dismissed):
:shipit:

CMakeLists.txt
```diff
@@ -72,6 +81,34 @@ set(run_traceability OFF)
...
+    if ((WIN32) AND (${GBALLOC_LL_TYPE} STREQUAL "JEMALLOC"))
+        find_package(PkgConfig REQUIRED)
+        pkg_check_modules (JEMALLOC jemalloc)
+
+        pkg_search_module(JEMALLOC REQUIRED jemalloc)
+        include_directories(${JEMALLOC_INCLUDE_DIRS})
+
+        #this wraps the jemalloc library in a target so it can be used in the rest of the build easily
+        add_library(jemalloc INTERFACE)
+        target_link_libraries(jemalloc INTERFACE ${pkgcfg_lib_JEMALLOC_jemalloc_s})
```
anporumb, line 105: >pkgcfg_lib_JEMALLOC_jemalloc_s [](http://example.com/codeflow?start=51&length=30)

There's a level of "magic" concerning this name here, like ... where is it coming from and how was it discovered.

Would be nice to have some comments that explain that, or links to documentation. What if *I* need to maintain this in 2 months?

CMakeLists.txt
```diff
@@ -28,6 +28,15 @@ option(use_cppunittest "set use_cppunittest to ON to build CppUnitTest tests on
 option(run_traceability "run traceability tool (default is ON)" ON)
 option(run_reals_check "set run_reals_check to ON to run reals check (default is OFF)." OFF)
 
+if((NOT "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo") AND (NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug"))
+    message(FATAL_ERROR "CMAKE_BUILD_TYPE must be specified and be precisely one of Debug or RelWithDebInfo")
+endif()
+
+set(CMAKE_CONFIGURATION_TYPES "${CMAKE_BUILD_TYPE}" CACHE STRING "" FORCE)
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG $ENV{BUILD_BINARIESDIRECTORY}/Debug)
```
anporumb, line 37: >BUILD_BINARIESDIRECTORY [](http://example.com/codeflow?start=46&length=23)

I know we verify this env var to exist in EBS repo, but I think this pushes the check all the way down to c-pal.

Let's make sure that the env variable exists at cmake time!

CMakeLists.txt
```diff
@@ -28,6 +28,15 @@ option(use_cppunittest "set use_cppunittest to ON to build CppUnitTest tests on
 option(run_traceability "run traceability tool (default is ON)" ON)
 option(run_reals_check "set run_reals_check to ON to run reals check (default is OFF)." OFF)
 
+if((NOT "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo") AND (NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug"))
+    message(FATAL_ERROR "CMAKE_BUILD_TYPE must be specified and be precisely one of Debug or RelWithDebInfo")
+endif()
+
+set(CMAKE_CONFIGURATION_TYPES "${CMAKE_BUILD_TYPE}" CACHE STRING "" FORCE)
```
anporumb, line 35: >CMAKE_CONFIGURATION_TYPES [](http://example.com/codeflow?start=4&length=25)

a comment explaining why this line is needed would be really nice for future maintainers of the code.

build/gated_jobs/build_and_run_tests.yml
```diff
@@ -0,0 +1,109 @@
...
+    - visualstudio
+
+  workspace:
+    clean: all
+
+  steps:
+  - checkout: self
+    submodules: true
+    clean: false
+    persistCredentials: true
```
anporumb, line 43: >persistCredentials [](http://example.com/codeflow?start=4&length=18)

just wondering why this is needed.

build/gated_jobs/build_and_run_tests.yml
```diff
@@ -0,0 +1,109 @@
...
+  - task: BatchScript@1
+    displayName: 'Setup VS Vars'
+    inputs:
+      filename: '"c:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat"'
+      modifyEnvironment: true
+
+  - task: CMake@1
+    displayName: 'CMake .. -Drun_reals_check:bool=ON -Drun_unittests:bool=ON -Drun_int_tests:bool=ON -Duse_cppunittest:bool=ON ${{ variables.use_vld_option }} -DGBALLOC_LL_TYPE:string=${{ parameters.GBALLOC_LL_TYPE }} -DCMAKE_BUILD_TYPE=${{ parameters.build_configuration }} -G "Visual Studio 17 2022" -A ${{ parameters.ARCH_TYPE }}'
+    inputs:
+      workingDirectory: 'build_x64'
```
anporumb, line 72: >build_x64 [](http://example.com/codeflow?start=25&length=9)

can it be that the working directory should be different for different architectures/build types?

anporumb (approved):
:shipit:

CMakeLists.txt
dcristoloveanu: Added comments

CMakeLists.txt
dcristoloveanu: Added check for it

CMakeLists.txt
dcristoloveanu: Added more comments

build/gated_jobs/build_and_run_tests.yml
dcristoloveanu: I believe that is a copy/pasta remnant from the main repo. Not needed => removed.

build/gated_jobs/build_and_run_tests.yml
dcristoloveanu: Renamed to cmake_build which should be generic. I don't think it should matter what name we use, but yeah x64 would be weird if we're actually building x86 later.

CMakeLists.txt
```diff
@@ -28,6 +28,24 @@ option(use_cppunittest "set use_cppunittest to ON to build CppUnitTest tests on
...
+if((NOT "${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo") AND (NOT "${CMAKE_BUILD_TYPE}" STREQUAL "Debug"))
+    message(FATAL_ERROR "CMAKE_BUILD_TYPE must be specified and be precisely one of Debug or RelWithDebInfo")
+endif()
+
+# This is done because jemalloc is obtained from vcpkg and brought in with PkgConfig
+# PkgConfig does not support defining different variables with the libraries to be linked for different configurations
+# This forces us to only be able to generate one build configuration (either Debug or RelWithDebInfo)
+# Setting CMAKE_CONFIGURATION_TYPES is the way to limit what configurations are generated
+# so that the user gets only one configuration as option in the Visual Studio GUI
+set(CMAKE_CONFIGURATION_TYPES "${CMAKE_BUILD_TYPE}" CACHE STRING "" FORCE)
```
mattdurak, line 40: Should these checks be under "if ((WIN32) AND (${GBALLOC_LL_TYPE} STREQUAL "JEMALLOC"))" because they are only needed for jemalloc?

I guess not a big deal other than changing how we work locally to have to re-run cmake to switch release/debug instead of using the UI.

anporumb (approved):
:shipit:

CMakeLists.txt
dcristoloveanu: Well, let's see if it gets to be annoying or not :-).

PR #324, opened 2024-04-05
Adding MALLOC_MULTI_FLEX macro

ipg2013:
`MALLOC_MULTI_FLEX` allows to allocate memory for a structure containing multiple variable sized array members using a single malloc.


interfaces/devdoc/malloc_multi_flex_requirements.md
```diff
@@ -0,0 +1,124 @@
...
+```c
+static void* malloc_multi_flex_PARENT_STRUCT(size_t parent_struct_size, uint32_t array_1_count, uint32_t array_1_size, uint32_t array_2_count, uint32_t array_2_size, uint32_t array_3_count, uint32_t array_3_size) 
+{
+    // perform overflow checks while calculating total size to be allocated
+    size_t size_tracker = parent_struct_size; 
+    if ((array_1_size != 0 && SIZE_MAX / array_1_size < array_1_count) || (SIZE_MAX - size_tracker < array_1_count * array_1_size)) 
+    {
+        return NULL;
+    } 
+    size_tracker += array_1_count * array_1_size;
```
mattdurak, line 61: >ize_tracker += array_1_count * array_1_size; [](http://example.com/codeflow?start=5&length=44)

I think one more thing we need to take into consideration as we do something more complex like this is alignment.

Possibly helpful reading [https://en.cppreference.com/w/c/language/object#Alignment](https://en.cppreference.com/w/c/language/object#Alignment)

I'm fairly sure this will work if we also compute the `alignof` for the type of the array element and add padding before each array to make sure it is aligned.

But to be honest, this isn't something I'm completely familiar with and I'd like to defer to maybe Dan or Andrei to review this as well. Andrei has an example of this in `async_op_create`

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+        MU_FOR_EACH_1(ASSIGN_INTERNAL_STRUCT_PTR, __VA_ARGS__)
+
+    //Codes_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]
+    #define MALLOC_MULTI_FLEX(type) \
+        MU_C2(malloc_multi_flex, type) \
+
+    #define DEFINE_MALLOC_MULTI_FLEX(type, ...)\
+        static void* MALLOC_MULTI_FLEX(type)(size_t parent_struct_size ARG_LIST_VALUES(__VA_ARGS__))   \
+        {\
+            //Codes_SRS_MALLOC_MULTI_FLEX_24_001: [ If the total amount of memory required to allocate the type along with its members exceeds SIZE_MAX then DEFINE_MALLOC_MULTI_FLEX shall fail and return NULL. ]\
```
mattdurak, line 50: >// [](http://example.com/codeflow?start=12&length=2)

within the macro you should have multi-line comments like `/*comment*/`

interfaces/inc/c_pal/gballoc_hl.h
```diff
@@ -11,6 +11,8 @@
 #include <stdint.h>
 #endif
 
+#include "malloc_multi_flex.h"
```
mattdurak, line 14: >malloc_multi_flex [](http://example.com/codeflow?start=10&length=17)

I think this is definitely a special-case and we can just have callers include malloc_multi_flex as needed rather than having it in this header.

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#endif
+
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    #define ARG_OVERFLOW_CHECK(arg) \
```
mattdurak, line 19: >ARG_OVERFLOW_CHECK [](http://example.com/codeflow?start=12&length=18)

we should probably prefix names with something so they aren't likely to collide with other defines that may be included together. E.g. `MALLOC_MULTI_FLEX_ARG_OVERFLOW_CHECK`

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+        if ((MU_C2(arg, _size) != 0 && SIZE_MAX / MU_C2(arg, _size) < MU_C2(arg, _count)) || (SIZE_MAX - size_tracker < MU_C2(arg, _count) * MU_C2(arg, _size)))\
+        {\
+            return NULL;\
+        }\
+        size_tracker += MU_C2(arg, _count) * MU_C2(arg, _size);
+
+    #define ARGS_OVERFLOW_CHECK(...) \
+        size_t size_tracker = parent_struct_size;\
```
mattdurak, line 27: >size_tracker [](http://example.com/codeflow?start=15&length=12)

poetry: maybe `size_required`

mattdurak, line 27: >size_t size_tracker = parent_struct_size;\ [](http://example.com/codeflow?start=8&length=42)

slightly more readable IMO if this line is in the main function like around line 49-50

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+        MU_FOR_EACH_1(ARG_LIST_VALUE_ARG, __VA_ARGS__)
+
+    #define ASSIGN_INTERNAL_STRUCT_PTR(arg) \
+        parent_struct_pointer->arg = pointer_iterator;\
+        pointer_iterator = (char*)pointer_iterator + MU_C2(arg, _count) * MU_C2(arg, _size);;
+
+    #define ASSIGN_INTERNAL_STRUCT_PTRS(...) \
+        MU_FOR_EACH_1(ASSIGN_INTERNAL_STRUCT_PTR, __VA_ARGS__)
+
+    //Codes_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]
```
mattdurak, line 43: >Codes_MALLOC_MULTI_FLEX_24_005 [](http://example.com/codeflow?start=6&length=30)

We should have tests before checking this in

interfaces/inc/c_pal/malloc_multi_flex.h
mattdurak: here and everywhere for the "internal" things

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+        MU_FOR_EACH_1(ARG_OVERFLOW_CHECK, __VA_ARGS__)
+
+    #define ARG_LIST_VALUE_ARG(arg) \
+        , uint32_t MU_C2(arg, _count), uint32_t MU_C2(arg, _size)
+
+    #define ARG_LIST_VALUES(...) \
+        MU_FOR_EACH_1(ARG_LIST_VALUE_ARG, __VA_ARGS__)
+
+    #define ASSIGN_INTERNAL_STRUCT_PTR(arg) \
+        parent_struct_pointer->arg = pointer_iterator;\
```
mattdurak, line 37: >pointer_iterator [](http://example.com/codeflow?start=37&length=16)

as mentioned, we should be adding some padding to make sure each the start of each array is aligned based on the alignment requirements of the type (of course you also need this in the overflow check and the size calculation).

interfaces/devdoc/malloc_multi_flex_requirements.md
```diff
@@ -0,0 +1,124 @@
...
+
+Lets say the user wants to allocate memory for `PARENT_STRUCT` which has three array members:
+- array_1 of type `uint32_t`
+- array_2 of type `uint64_t`
+- array_3 of type `INNER_STRUCT`
+
+The caller first needs to specify the struct and member details using `DEFINE_MALLOC_MULTI_FLEX` macro in `.c` file to create a custom `malloc_multi_flex_<type>` function as defined below -
+
+```c
+DEFINE_MALLOC_MULTI_FLEX(PARENT_STRUCT, array_1, array_2, array_3)
```
mattdurak, line 47: if we actually include the type of each array field in this macro, then it can internally generate the `sizeof(type)` and `alignof(type)` code and the generated function could just take the array counts.

/consider

interfaces/inc/c_pal/malloc_multi_flex.h
ipg2013: Added padding in `MALLOC_MULTI_FLEX_ASSIGN_INTERNAL_STRUCT_PTR` macro and updated overflow check macro - `MALLOC_MULTI_FLEX_ARG_OVERFLOW_CHECK` too

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,72 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#endif
+
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    #define alignof _Alignof
```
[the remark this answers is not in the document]
mattdurak: Actually we do build with C17, I don't think we need the define

```
macro(set_default_build_options)
    #compile the code as C17
    set(CMAKE_C_STANDARD 17)
```

But we should have `#include <stdalign.h>` and `#include <cstdalign>`

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,72 @@
...
+
+    #define MALLOC_MULTI_FLEX_ARG_OVERFLOW_CHECK(arg1, arg2) \
+        if ((sizeof(arg1) != 0 && SIZE_MAX / sizeof(arg1) < MU_C2(arg2, _count)) || (SIZE_MAX - size_required < MU_C2(arg2, _count) * sizeof(arg1)) || (SIZE_MAX - (size_required + MU_C2(arg2, _count) * sizeof(arg1)) < alignof(arg1)))\
+        {\
+            return NULL;\
+        }\
+        size_required += MU_C2(arg2, _count) * sizeof(arg1) + alignof(arg1);
```
mattdurak, line 26: >alignof(arg1) [](http://example.com/codeflow?start=62&length=13)

worst case we only need `alignof(T) - 1`

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,140 @@
...
+
+static void assign_struct(PARENT_STRUCT* test_struct_handle)
+{
+    test_struct_handle->int_1 = 3;
+    test_struct_handle->int_2 = 6;
+    test_struct_handle->int_3 = 9;
+
+    for (int i = 0; i < array1_size; i++)
+    {
+        test_struct_handle->array_1[i] = i;
```
mattdurak, line 58: >test_struct_handle->array_1[i] = i; [](http://example.com/codeflow?start=8&length=35)

we should probably set all of the bits to really make sure we didn't have some overlap.

ipg2013: Not sure how set all bits.

I tried this -

```
size_t total_size = sizeof(PARENT_STRUCT) + (array1_size * sizeof(uint32_t)) + alignof(uint32_t) - 1 + (array2_size * sizeof(uint64_t)) + alignof(uint64_t) - 1 + (array3_size * sizeof(INNER_STRUCT)) + alignof(INNER_STRUCT) - 1;
memset(test_struct_handle, 1, total_size);
```

But this memset causes the program to crash. Need help here.

interfaces/devdoc/malloc_multi_flex_requirements.md
dcristoloveanu: +1. When I read the synopsis of this I was thinking that we have to take into account alignment. So I was then looking for this specific part where we compute the size ... 

I believe we should be using alignof

interfaces/inc/c_pal/gballoc_hl.h
dcristoloveanu: I think malloc_multi_flex is built on top of malloc if we want to not copy/pasta it in all the gballoc implementations.

Arguably malloc_flex and malloc_2 is actually the same, but those are already in place there in the gballoc API.

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+    //Codes_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]
+    #define MALLOC_MULTI_FLEX(type) \
+        MU_C2(malloc_multi_flex, type) \
+
+    #define DEFINE_MALLOC_MULTI_FLEX(type, ...)\
+        static void* MALLOC_MULTI_FLEX(type)(size_t parent_struct_size ARG_LIST_VALUES(__VA_ARGS__))   \
+        {\
+            //Codes_SRS_MALLOC_MULTI_FLEX_24_001: [ If the total amount of memory required to allocate the type along with its members exceeds SIZE_MAX then DEFINE_MALLOC_MULTI_FLEX shall fail and return NULL. ]\
+            ARGS_OVERFLOW_CHECK(__VA_ARGS__)\
+            //Codes_SRS_MALLOC_MULTI_FLEX_24_002: [ DEFINE_MALLOC_MULTI_FLEX shall call malloc to allocate memory for the struct and its members. ]\
```
dcristoloveanu: >Codes [](http://example.com/codeflow?start=14&length=5)

Seems that we have Codes, but no Tests?

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+    #define ASSIGN_INTERNAL_STRUCT_PTR(arg) \
+        parent_struct_pointer->arg = pointer_iterator;\
+        pointer_iterator = (char*)pointer_iterator + MU_C2(arg, _count) * MU_C2(arg, _size);;
+
+    #define ASSIGN_INTERNAL_STRUCT_PTRS(...) \
+        MU_FOR_EACH_1(ASSIGN_INTERNAL_STRUCT_PTR, __VA_ARGS__)
+
+    //Codes_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]
+    #define MALLOC_MULTI_FLEX(type) \
+        MU_C2(malloc_multi_flex, type) \
```
dcristoloveanu: >MU_C2 [](http://example.com/codeflow?start=8&length=5)

Needs macro utils include

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+
+    #define DEFINE_MALLOC_MULTI_FLEX(type, ...)\
```
dcristoloveanu, line 48: >DEFINE_MALLOC_MULTI_FLEX [](http://example.com/codeflow?start=12&length=24)

I think this should have a DECLARE/DEFINE pair like any other template that we build.

dcristoloveanu: declare would be used in a header, define would be used in .c.
That gives the flexibility to the user to split the declaration/definition in .h and .c or to keep them both in the same file.

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+
+    #define ARG_LIST_VALUE_ARG(arg) \
+        , uint32_t MU_C2(arg, _count), uint32_t MU_C2(arg, _size)
+
+    #define ARG_LIST_VALUES(...) \
+        MU_FOR_EACH_1(ARG_LIST_VALUE_ARG, __VA_ARGS__)
+
+    #define ASSIGN_INTERNAL_STRUCT_PTR(arg) \
+        parent_struct_pointer->arg = pointer_iterator;\
+        pointer_iterator = (char*)pointer_iterator + MU_C2(arg, _count) * MU_C2(arg, _size);;
```
dcristoloveanu: >; [](http://example.com/codeflow?start=92&length=1)

Extra ;

dcristoloveanu: >(char*) [](http://example.com/codeflow?start=27&length=7)

Nit: char* makes one think we're dealing with a string, I'd use uint8_t*

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+#include <stdint.h>
+#endif
+
+#include "umock_c/umock_c_prod.h"
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    #define ARG_OVERFLOW_CHECK(arg) \
+        if ((MU_C2(arg, _size) != 0 && SIZE_MAX / MU_C2(arg, _size) < MU_C2(arg, _count)) || (SIZE_MAX - size_tracker < MU_C2(arg, _count) * MU_C2(arg, _size)))\
```
dcristoloveanu: >_MAX - size_tracker < MU_C2(arg, _coun [](http://example.com/codeflow?start=98&length=38)

There is a mix of size_t and uint32_t

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,65 @@
...
+        {\
+            return NULL;\
```
dcristoloveanu, line 24: > return NULL;\ [](http://example.com/codeflow?start=11&length=14)

I think we'd want to log here

interfaces/inc/c_pal/malloc_multi_flex.h
dcristoloveanu: It does seem like this is meant to be the .h and .md actually, so maybe this comment is relevant in the sense that we should not have the Codes here as they would anyway be seen as mismatched by traceability.

dcristoloveanu:
Interested to see the next iteration.

dcristoloveanu (dismissed):
:clock1:

interfaces/inc/c_pal/malloc_multi_flex.h
ipg2013: `malloc_muli_flex_int` has the int tests

interfaces/inc/c_pal/malloc_multi_flex.h
ipg2013: uintptr_t has been used in the latest iteration

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
ipg2013: Fixed

interfaces/devdoc/malloc_multi_flex_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+
+The caller first needs to specify the struct name, member types and names using `DECLARE_MALLOC_MULTI_FLEX` macro to define the struct and declare the malloc function:
+
+```c
+DECLARE_MALLOC_MULTI_FLEX(PARENT_STRUCT,
+    FIELDS(uint64_t, int_1, uint32_t, int_2, uint32_t, int_3),
+    ARRAY_FIELDS(uint32_t, array_1, uint64_t, array_2, INNER_STRUCT, array_3))
+```
+
+`DECLARE_MALLOC_MULTI_FLEX` will define the struct as follows
```
dcristoloveanu, line 48: >ws [](http://example.com/codeflow?start=59&length=2)

I would add a :

interfaces/devdoc/malloc_multi_flex_requirements.md
```diff
@@ -0,0 +1,135 @@
...
+{
+    uint32_t inner_int_1;
+    uint64_t inner_int_2;
+}INNER_STRUCT;
+```
+
+The caller first needs to specify the struct name, member types and names using `DECLARE_MALLOC_MULTI_FLEX` macro to define the struct and declare the malloc function:
+
+```c
+DECLARE_MALLOC_MULTI_FLEX(PARENT_STRUCT,
```
dcristoloveanu, line 43: >DECLARE_MALLOC_MULTI_FLEX [](http://example.com/codeflow?start=0&length=25)

Looking at this name it does not really tell me we have to deal with a struct. Maybe (just maybe) it would be a good thing to contain in its name the word "struct" somehow. Not married to the idea or what the name should be, just a Shakespearian consider

interfaces/inc/c_pal/malloc_multi_flex.h
```diff
@@ -0,0 +1,133 @@
...
+            /* Codes_SRS_MALLOC_MULTI_FLEX_24_002: [ DEFINE_MALLOC_MULTI_FLEX shall call malloc to allocate memory for the struct and its members. ]*/ \
+            type* parent_struct_pointer = malloc(size_required);\
+            if (parent_struct_pointer == NULL)\
+            {\
+                LogError("malloc(size_required = %zu) failed", size_required);\
+                return NULL;\
+            }\
+            uintptr_t pointer_iterator = (uintptr_t)parent_struct_pointer + parent_struct_size; \
+            MALLOC_MULTI_FLEX_ARGS_VALIDATION(array_fields)\
+            /* Codes_SRS_MALLOC_MULTI_FLEX_24_003: [ DEFINE_MALLOC_MULTI_FLEX shall assign address pointers to all the member arrays. ]*/ \
```
dcristoloveanu, line 123: I think this leaks. Should have been caught by unit tests.

dcristoloveanu: It seems like all checks should be done before we malloc.

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+    FIELDS(uint64_t, int_1, uint32_t, int_2, uint32_t, int_3),
+    ARRAY_FIELDS(uint32_t, array_1, uint64_t, array_2, INNER_STRUCT, array_3))
+
+DEFINE_MALLOC_MULTI_FLEX(PARENT_STRUCT,
+    FIELDS(uint64_t, int_1, uint32_t, int_2, uint32_t, int_3),
+    ARRAY_FIELDS(uint32_t, array_1, uint64_t, array_2, INNER_STRUCT, array_3))
+
+static void set_all_bits(PARENT_STRUCT* test_struct_handle)
+{
+    size_t bytes_to_set = (array1_size * sizeof(uint32_t)) + alignof(uint32_t) - 1 + (array2_size * sizeof(uint64_t)) + alignof(uint64_t) - 1 + (array3_size * sizeof(INNER_STRUCT)) + alignof(INNER_STRUCT) - 1;
```
dcristoloveanu, line 51: >size_t [](http://example.com/codeflow?start=4&length=6)

Needs some stddef include I think

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+    memset(test_struct_handle + sizeof(PARENT_STRUCT), 0xFF, bytes_to_set);
```
dcristoloveanu, line 52: >memset [](http://example.com/codeflow?start=4&length=6)

(void)

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+        ASSERT_ARE_EQUAL(int, i + 2000, test_struct_handle->array_3[i].inner_int_2);
+    }
+}
+
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_001: [ If the total amount of memory required to allocate the type along with its members exceeds SIZE_MAX then DEFINE_MALLOC_MULTI_FLEX shall fail and return NULL. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_002: [ DEFINE_MALLOC_MULTI_FLEX shall call malloc to allocate memory for the struct and its members. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_003: [ DEFINE_MALLOC_MULTI_FLEX shall assign address pointers to all the member arrays. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_004: [ DEFINE_MALLOC_MULTI_FLEX shall succeed and return the address returned by malloc. ]*/
+/* Tests_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]*/
+TEST_FUNCTION(test_malloc_multi_flex_allocates_memory_and_assigns_address_ptr_for_parent_and_member_arrays)
```
dcristoloveanu, line 106: I think we should add a bit more tests

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+        ASSERT_ARE_EQUAL(int, i + 1000, test_struct_handle->array_3[i].inner_int_1);
+        ASSERT_ARE_EQUAL(int, i + 2000, test_struct_handle->array_3[i].inner_int_2);
+    }
+}
+
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_001: [ If the total amount of memory required to allocate the type along with its members exceeds SIZE_MAX then DEFINE_MALLOC_MULTI_FLEX shall fail and return NULL. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_002: [ DEFINE_MALLOC_MULTI_FLEX shall call malloc to allocate memory for the struct and its members. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_003: [ DEFINE_MALLOC_MULTI_FLEX shall assign address pointers to all the member arrays. ]*/
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_004: [ DEFINE_MALLOC_MULTI_FLEX shall succeed and return the address returned by malloc. ]*/
+/* Tests_MALLOC_MULTI_FLEX_24_005: [ MALLOC_MULTI_FLEX shall expand type to the name of the malloc function in the format of: MALLOC_MULTI_FLEX_type. ]*/
```
dcristoloveanu, line 105: >s_MALLOC_ [](http://example.com/codeflow?start=7&length=9)

SRS missing. Is traceability not running in the gate?

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+    ASSERT_ARE_EQUAL(int, 9, test_struct_handle->int_3);
+
+    for (int i = 0; i < array3_size; i++)
+    {
+        ASSERT_ARE_EQUAL(int, i + 1000, test_struct_handle->array_3[i].inner_int_1);
+        ASSERT_ARE_EQUAL(int, i + 2000, test_struct_handle->array_3[i].inner_int_2);
+    }
+}
+
+/* Tests_SRS_MALLOC_MULTI_FLEX_24_001: [ If the total amount of memory required to allocate the type along with its members exceeds SIZE_MAX then DEFINE_MALLOC_MULTI_FLEX shall fail and return NULL. ]*/
```
dcristoloveanu, line 101: I think this should be tested as a separate test

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
```diff
@@ -0,0 +1,122 @@
...
+
+#include <stdint.h>
+
+#include "testrunnerswitcher.h"
+
+#include "c_pal/gballoc_hl.h"
+#include "c_pal/gballoc_hl_redirect.h"
+
+#include "c_pal/malloc_multi_flex.h"
+
```
dcristoloveanu, line 12: Why are these not unit tests?
How do we simulate malloc failing? We probably do not. If we did we would have caught the leak I spotted in the review ...

dcristoloveanu (changes requested):
:clock1:

interfaces/inc/c_pal/malloc_multi_flex.h
ipg2013: >I think this leaks. Should have been caught by unit tests.

As per the code 
```c
#define MALLOC_MULTI_FLEX_ARG_VALIDATION(arg1, arg2) \
        if (MU_C2(arg2, _count) > 0 && sizeof(arg1) != sizeof(parent_struct_pointer->arg2[0]))\
        {\
            LogError("Data type size type mistmatch. Declared type has size = %zu but actual member has size = %zu", sizeof(arg1), sizeof(parent_struct_pointer->arg2[0]));\
            free(parent_struct_pointer);\
            return NULL;\
        }
```
This doesn't leak because we are freeing the pointer if validation fails.

> It seems like all checks should be done before we malloc.

This check validates size discrepancy. `parent_struct_pointer->arg2[0]` is only accessible after malloc. So, we cannot change the order.

ipg2013: I have removed this arg validation check now because while reproing this scenario in UTs I realized that complier takes care of arg validation and we do not need to do this explicitly.

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
ipg2013: Fixed

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
ipg2013: Added

win32/tests/malloc_multi_flex_int/malloc_multi_flex_int.c
ipg2013: Added `malloc_multi_flex_fails_when_size_exceeds_SIZE_MAX`

interfaces/devdoc/malloc_multi_flex_requirements.md
ipg2013: Fixed

win32/tests/malloc_multi_flex_ut/malloc_multi_flex_ut.c
```diff
@@ -0,0 +1,134 @@
...
+{
+    // arrange
+
+    // act
+    PARENT_STRUCT* parent_struct = create_parent_struct(UINT64_MAX, 20, 30);
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_IS_NULL(parent_struct);
+}
```
dcristoloveanu, line 132: Nice!

dcristoloveanu (approved):
:shipit:

PR #326, opened 2024-04-08
Add a guard around the work function in timer callback.

darobs:
Add a guard around the work function in timer callback so the timer can not be deleted while the event has been triggered.
Add a slight delay on timer destroy to allow thread cleanup and prevent the timer from being executed even though timer has been destroyed.

### DIFF
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -59,6 +59,9 @@ stateDiagram-v2
     OK_TO_WORK --> TIMER_DELETING : threadpool_timer_destroy 
 ```
 
+Even with this guard in place, during `threadpool_timer_delete` there is a small window where an event was triggered just before and a thread is being created. 
+The guard prevents any work from happening, but the event callback can still be called after the `timer_delete` call. A small wait before the free keeps the timer instance alive long enough for the callback to use it if we hit this window.
+
 ## Exposed API
 
 ```C
@@ -331,7 +334,7 @@ static void on_timer_callback(sigval_t timer_data);
 
 **SRS_THREADPOOL_LINUX_45_002: [** `on_timer_callback` shall set the timer instance to `timer_data.sival_ptr`. **]**
 
-**SRS_THREADPOOL_LINUX_45_001: [** If timer instance is NULL, then `on_timer_callback` shall return. **]**
+**SRS_THREADPOOL_LINUX_45_001: [** If timer instance is `NULL`, then `on_timer_callback` shall return. **]**
 
 **SRS_THREADPOOL_LINUX_45_003: [** `on_timer_callback` shall call `interlocked_compare_exchange` with the `timer_work_guard` of this timer instance with `OK_TO_WORK` as the comparison, and `TIMER_WORKING` as the exchange. **]**
 
@@ -340,7 +343,6 @@ static void on_timer_callback(sigval_t timer_data);
 **SRS_THREADPOOL_LINUX_45_005: [** `on_timer_callback` shall call `interlocked_compare_exchange` with the `timer_work_guard` of this timer instance with `TIMER_WORKING` as the comparison, and `OK_TO_WORK` as the exchange. **]**
 
 **SRS_THREADPOOL_LINUX_45_006: [** If `timer_work_guard` is successfully set to `OK_TO_WORK`, then then `on_timer_callback` shall call `wake_by_address_single` on `timer_work_guard`. **]**
-
 
 ### threadpool_work_func
 
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -43,6 +43,21 @@ graph TD
     B--> C(gaps occured)-->E(memmove)-->F(resize done)
 
 ```
+### Threadpool timer
+
+The threadpool timer currently uses the POSIX timer implementation with the timer events generating a new thread (`SIGEV_THREAD`) when an event triggers.
+When the event is triggered, it needs the timer instance to stay allocated until it is completed, so there is a guard to prevent the timer from being deleted while the timer is executing its work function.
+
+```mermaid
+---
+title: Event Triggered State
+---
+stateDiagram-v2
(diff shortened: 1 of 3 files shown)


linux/src/threadpool_linux.c
```diff
@@ -716,6 +757,8 @@ void threadpool_timer_destroy(TIMER_INSTANCE_HANDLE timer)
         {
             // Do Nothing
         }
+        /* Codes_SRS_THREADPOOL_LINUX_45_012: [ threadpool_timer_cancel shall call ThreadAPI_Sleep to allow timer resources to clean up. ]*/
```
mattdurak, line 760: >Codes_SRS_THREADPOOL_LINUX_45_012 [](http://example.com/codeflow?start=11&length=33)

I guess we should have a comment here explaining why this is needed (the timer thread can be scheduled even though we called timer_delete.

Also I had no idea the win32 API's were so much better for this, `WaitForThreadpoolTimerCallbacks` makes this much easier in Win32!

mattdurak (dismissed):
:shipit:

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -43,6 +43,21 @@ graph TD
     B--> C(gaps occured)-->E(memmove)-->F(resize done)
 
 ```
+### Threadpool timer
```
jebrando, line 46: >Threadpool timer [](http://example.com/codeflow?start=4&length=16)

One thing that may be helpful is to mention the bug that we found with the delete.

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -283,10 +307,41 @@ MOCKABLE_FUNCTION(, void, threadpool_timer_destroy, TIMER_INSTANCE_HANDLE, timer
...
+
+```C
+static void on_timer_callback(sigval_t timer_data);
+```
+
+`on_timer_callback` executes on a new thread when the POSIX timer is triggered.
+
+**SRS_THREADPOOL_LINUX_45_002: [** `on_timer_callback` shall set the timer instance to `timer_data.sival_ptr`. **]**
+
+**SRS_THREADPOOL_LINUX_45_001: [** If timer instance is NULL, then `on_timer_callback` shall return. **]**
```
jebrando, line 334: >NULL [](http://example.com/codeflow?start=56&length=4)

tick marks here.

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -283,10 +307,41 @@ MOCKABLE_FUNCTION(, void, threadpool_timer_destroy, TIMER_INSTANCE_HANDLE, timer
...
+**SRS_THREADPOOL_LINUX_45_003: [** `on_timer_callback` shall call `interlocked_compare_exchange` with the `timer_work_guard` of this timer instance with `OK_TO_WORK` as the comparison, and `TIMER_WORKING` as the exchange. **]**
+
+**SRS_THREADPOOL_LINUX_45_004: [** If `timer_work_guard` is successfully set to `TIMER_WORKING`, then `on_timer_callback` shall call the timer's `work_function` with `work_function_ctx`. **]**
+
+**SRS_THREADPOOL_LINUX_45_005: [** `on_timer_callback` shall call `interlocked_compare_exchange` with the `timer_work_guard` of this timer instance with `TIMER_WORKING` as the comparison, and `OK_TO_WORK` as the exchange. **]**
+
+**SRS_THREADPOOL_LINUX_45_006: [** If `timer_work_guard` is successfully set to `OK_TO_WORK`, then then `on_timer_callback` shall call `wake_by_address_single` on `timer_work_guard`. **]**
+
+
 ### threadpool_work_func
```
jebrando, line 345: extra newline

jebrando (dismissed):
:clock1:

jebrando (dismissed):
:shipit:

linux/src/threadpool_linux.c
darobs: Added a comment before the wait which explains why it's there.

linux/devdoc/threadpool_linux_requirements.md
darobs: Added some text to describe the problem.

avranju (approved):

linux/src/threadpool_linux.c
```diff
@@ -48,12 +49,19 @@
 MU_DEFINE_ENUM(TASK_RESULT, TASK_RESULT_VALUES);
 MU_DEFINE_ENUM_STRINGS(TASK_RESULT, TASK_RESULT_VALUES)
 
+typedef enum TIMER_GUARD_TAG
+{
+    OK_TO_WORK,
+    TIMER_WORKING,
+    TIMER_DELETING,
+} TIMER_GUARD;
```
dcristoloveanu, line 57: This should be MU_DEFINE_ENUM so we can use proper strings if needed

linux/src/threadpool_linux.c
```diff
@@ -716,6 +757,11 @@ void threadpool_timer_destroy(TIMER_INSTANCE_HANDLE timer)
         {
             // Do Nothing
         }
+        // Even though timer_delete does cancel any events, there is a small window where an event was triggered just before and a thread is being created.
+        // So the callback can still execute after the timer_delete call. A small wait here keeps the timer instance alive long enough for
+        // the callback to use it if we hit this window.
+        /* Codes_SRS_THREADPOOL_LINUX_45_012: [ threadpool_timer_cancel shall call ThreadAPI_Sleep to allow timer resources to clean up. ]*/
+        ThreadAPI_Sleep(10);
```
dcristoloveanu, line 764: >ThreadAPI_Sleep(10); [](http://example.com/codeflow?start=8&length=20)

Hmmmm, why not 20? Seems that any time here is not going to protect in any way. It just makes the chance of the issue happening less, but it will not eliminate it ....

darobs: Will add:
- state it's not a fix.
- suggest proper fix
- suggest how to repro.

darobs: #327

PR #327, opened 2024-04-09
Add additional comments for the known linux timer problem.

darobs:
Add additional comments for the known problem, suggest long-term
improvements and how to reproduce. 
Also use MU_DEFINE_ENUM for timer_work_guard state.


mattdurak (approved):
:shipit:

PR #336, opened 2024-05-09
thandle ptr helper

anporumb:


common/inc/c_pal/thandle_ptr.h
```diff
@@ -0,0 +1,74 @@
...
+#define THANDLE_PTR_DECLARE(T) \
+    THANDLE_PTR_FREE_FUNC_TYPE(T);                                                                              \
+    PTR_STRUCT_TYPE_TYPEDEF(T);                                                                                 \
+    THANDLE_TYPE_DECLARE(PTR(T));                                                                               \
+    THANDLE(PTR(T)) THANDLE_PTR_CREATE_WITH_MOVE(T)(T pointer, THANDLE_PTR_FREE_FUNC_TYPE_NAME(T) dispose );    \
+
+/*this introduces the definition of the function declared above*/
+#define THANDLE_PTR_DEFINE(T)                                                                                                                                           \
+    THANDLE_TYPE_DEFINE(PTR(T));                                                                                                                                        \
+    static void THANDLE_PTR_DISPOSE(T)(PTR(T)* ptr)                                                                                                                     \
```
dcristoloveanu, line 50: >static [](http://example.com/codeflow?start=4&length=6)

I like this static

dcristoloveanu:
> #include <stddef.h>

File seems quite untouched, can be reverted methinks

---
Refers to: common/tests/thandle_2_ut/thandle_2_ut.c:5 in b5d2c5d. [](commit_id = b5d2c5de2efe03d485598d3249a16b0697318a94, deletion_comment = False)

dcristoloveanu:
A quite dry review, I might say, quite what I would expect - nicely done.

dcristoloveanu:
In order to make things easier for plopping reals for various types, I think having something like real_thandle_helper.h would not be bad at all.

dcristoloveanu (approved):
:shipit:

darobs (approved):
:shipit:

common/devdoc/thandle_ptr_requirements.md
```diff
@@ -0,0 +1,91 @@
...
+Then use it like:
+```c
+THANDLE(PTR(A_S_PTR)) one = THANDLE_PTR_CREATE_WITH_MOVE(A_S_PTR)(a_s, dispose);
+```
+
+`a_s` is a pointer of type `A_S_PTR`.
+
+The above code "moves" ("captures") the pointer `a_s` and produces the `THANDLE` `one`. `dispose` is a user function that takes the original `a_s` pointer and frees all resource used by `a_s`.
+
+`one->pointer` produces the original `a_s` pointer. Other fields of the `one` pointer should not be used (it also captures the `dispose` function for example).
```
darobs, line 48: >one->pointer [](http://example.com/codeflow?start=1&length=12)

Should the access to this pointer be made a macro as well?

common/inc/c_pal/thandle_ptr.h
anporumb: Cheers!

anporumb:
> #include <stddef.h>

extraneous spaces have been automatically removed by an extension.

Let's not have extraneous spaces!

---
In reply to: [2103647243](https://github.com/Azure/c-pal/pull/336#issuecomment-2103647243) [](http://example.com/codeflow?ancestors=2103647243)

---
Refers to: common/tests/thandle_2_ut/thandle_2_ut.c:5 in b5d2c5d. [](commit_id = b5d2c5de2efe03d485598d3249a16b0697318a94, deletion_comment = False)

common/devdoc/thandle_ptr_requirements.md
anporumb: The name of the parameter passed to CREATE is `pointer` so maaaaybe this is good enough as it is. I think usage will tell if enough or not.

anporumb: However, this "because it is a THANDLE", the GET_T of the THANDLE should get the pointer out.

... with->pointer I am afraid the *pointer is read only. Give me 1 second to write another test :)

anporumb: I wrote some more tests where I prove that the ->pointer's fields have the same qualifiers as the original one.

anporumb:
Would not be, and let's have a separate task to do that. I've create task no 28026392 to track that effort

---
In reply to: [2103650525](https://github.com/Azure/c-pal/pull/336#issuecomment-2103650525) [](http://example.com/codeflow?ancestors=2103650525)

dcristoloveanu (approved):
:shipit:

PR #338, opened 2024-05-15
changing async_socket open close signature

jebrando:


mattdurak:
> The `async_socket_win32` takes ownership of the provided `SOCKET_HANDLE` and is responsible for closing it. For that reason, it is not possible to re-open the `async_socket` after closing it. This is due to the fact that the socket is created by some external event, such as accepting an incoming connection and closing the underlying socket is not reversible.

well I guess you've solved this limitation :)
Can probably remove these 2 sentences

---
Refers to: win32/devdoc/async_socket_win32.md:12 in 44bcb42. [](commit_id = 44bcb42773cc669b389ed936aefeb35d1166532c, deletion_comment = False)

mattdurak (dismissed):
:shipit:

jebrando:
> The `async_socket_win32` takes ownership of the provided `SOCKET_HANDLE` and is responsible for closing it. For that reason, it is not possible to re-open the `async_socket` after closing it. This is due to the fact that the socket is created by some external event, such as accepting an incoming connection and closing the underlying socket is not reversible.

Well, It's technically not solved, but has changed.  The ownership is still an issue and since async_socket doesn't create the socket, and it has to close the socket the ownership question is still an issue.  This is a major problem in my (and Andrei's) opinion that will only be resolve by a rewrite.

---
In reply to: [2113391057](https://github.com/Azure/c-pal/pull/338#issuecomment-2113391057) [](http://example.com/codeflow?ancestors=2113391057)

---
Refers to: win32/devdoc/async_socket_win32.md:12 in 44bcb42. [](commit_id = 44bcb42773cc669b389ed936aefeb35d1166532c, deletion_comment = False)

mattdurak:
> The `async_socket_win32` takes ownership of the provided `SOCKET_HANDLE` and is responsible for closing it. For that reason, it is not possible to re-open the `async_socket` after closing it. This is due to the fact that the socket is created by some external event, such as accepting an incoming connection and closing the underlying socket is not reversible.

IMO it just means we move/transfer the ownership on open, but it's cleaner now because you can't do open-close-open without getting a new SOCKET.

But yeah, probably plenty of rewriting needed here

---
In reply to: [2113447657](https://github.com/Azure/c-pal/pull/338#issuecomment-2113447657) [](http://example.com/codeflow?ancestors=2113447657,2113391057)

---
Refers to: win32/devdoc/async_socket_win32.md:12 in 44bcb42. [](commit_id = 44bcb42773cc669b389ed936aefeb35d1166532c, deletion_comment = False)

linux/src/async_socket_linux.c
```diff
@@ -514,30 +517,24 @@ int async_socket_open_async(ASYNC_SOCKET_HANDLE async_socket, ON_ASYNC_SOCKET_OP
...
-                // Codes_SRS_ASYNC_SOCKET_LINUX_11_032: [ async_socket_open_async shall set the state to OPEN. ]
-                (void)interlocked_exchange(&async_socket->state, ASYNC_SOCKET_LINUX_STATE_OPEN);
-                // Codes_SRS_ASYNC_SOCKET_LINUX_11_033: [ On success async_socket_open_async shall call on_open_complete_context with ASYNC_SOCKET_OPEN_OK. ]
-                on_open_complete(on_open_complete_context, ASYNC_SOCKET_OPEN_OK);
-
-                // Codes_SRS_ASYNC_SOCKET_LINUX_11_028: [ On success, async_socket_open_async shall return 0. ]
-                result = 0;
-                goto all_ok;
-            }
+            async_socket->socket_handle = socket_handle;
```
dcristoloveanu, line 526: >socket_handle [](http://example.com/codeflow?start=42&length=13)

Would be nice to have a structure like OPEN_DATA, where the data relevant for the OPEN state is kept. That way it is harder to inadvertently access it from create/destroy (not impossible, but would require the developer to write a little bit more). Also later it is wasier to make the OPEN_DATA a handle by itself.

Given the time constraints and the fact that this would be squirrling a bit, I give it a consider, but I did notice it, so "see something say something"

win32/devdoc/async_socket_win32.md
```diff
@@ -127,22 +125,22 @@ MOCKABLE_FUNCTION(, void, async_socket_destroy, ASYNC_SOCKET_HANDLE, async_socke
...
 ```c
-MOCKABLE_FUNCTION(, int, async_socket_open_async, ASYNC_SOCKET_HANDLE, async_socket, ON_ASYNC_SOCKET_OPEN_COMPLETE, on_open_complete, void*, on_open_complete_context);
+MOCKABLE_FUNCTION(, int, async_socket_open_async, ASYNC_SOCKET_HANDLE, async_socket, SOCKET_HANDLE, socket_handle, ON_ASYNC_SOCKET_OPEN_COMPLETE, on_open_complete, void*, on_open_complete_context);
 ```
 
 `async_socket_open_async` opens the async socket.
 
 **SRS_ASYNC_SOCKET_WIN32_01_007: [** If `async_socket` is NULL, `async_socket_open_async` shall fail and return a non-zero value. **]**
 
+**SRS_ASYNC_SOCKET_WIN32_01_034: [** If `socket_handle` is `INVALID_SOCKET`, `async_socket_open_async` shall fail and return NULL. **]**
```
dcristoloveanu, line 142: >NULL [](http://example.com/codeflow?start=125&length=4)

a non-zero value

win32/src/async_socket_win32.c
```diff
@@ -324,18 +320,11 @@ void async_socket_destroy(ASYNC_SOCKET_HANDLE async_socket)
             }
             else if (current_state == (LONG)ASYNC_SOCKET_WIN32_STATE_CLOSED)
             {
-                if ((SOCKET)async_socket->socket_handle != INVALID_SOCKET)
-                {
-                    LogInfo("async socket destroy, closesocket(%p);", async_socket->socket_handle);
-                    /* Codes_SRS_ASYNC_SOCKET_WIN32_42_007: [ If the socket was not OPEN then async_socket_destroy shall call closesocket on the underlying socket. ]*/
-                    (void)closesocket((SOCKET)async_socket->socket_handle);
```
dcristoloveanu, line 331: This smells like a workaround done by someone (Matt) at some point. I am happy we're removing it. We should double check with Matt tomorrow whether there is some hidden thing here (I suspect that it is indeed not needed any longer, but just to make sure ... ).

jebrando: Will do.

linux/src/async_socket_linux.c
jebrando: I will make a task for this, I would like to do more here and I have ideas, but due to time constraints a task will do for now.

jebrando: Task here: https://msazure.visualstudio.com/One/_workitems/edit/28084746

win32/devdoc/async_socket_win32.md
jebrando: Will update.

PR #341, opened 2024-06-13
spec and header for socket transport module

jebrando:


linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+#define SOCKET_TYPE_VALUES \
+    SOCKET_CLIENT, \
+    SOCKET_SERVER
+
+MU_DEFINE_ENUM(SOCKET_TYPE, SOCKET_TYPE_VALUES)
+
+typedef struct SOCKET_BUFFER_TAG
+{
+    uint32_t length;
+    void* buffer;
```
anporumb, line 50: >void [](http://example.com/codeflow?start=4&length=4)

let's have this as "unsigned char*'

Also since this "sort of" moves data, each SOCKET_BUFFER owns its data? Brings up the question of how to avoid (if possible) memory copies (if any...)

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_create, SOCKET_TYPE, type);
+MOCKABLE_FUNCTION(, void, socket_transport_destroy, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, int, socket_transport_connect, SOCKET_TRANSPORT_HANDLE, socket_transport, const char*, hostname, uint16_t, port, uint32_t, connection_timeout_ms);
+MOCKABLE_FUNCTION(, int, socket_transport_listen, SOCKET_TRANSPORT_HANDLE, socket_transport, uint16_t, port);
+MOCKABLE_FUNCTION(, void, socket_transport_disconnect, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_accept, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_written, uint32_t, flags, void*, data);
```
anporumb, line 62: >SOCKET_BUFFER [](http://example.com/codeflow?start=106&length=13)

const? looks like it doesn't ever modify the data inside seeing as it is "send".

anporumb, line 62: >bytes_written [](http://example.com/codeflow?start=166&length=13)

`written`? if the API is called "XXXsend", maybe "send"?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+MOCKABLE_FUNCTION(, SOCKET_HANDLE, socket_transport_get_underlying_socket, SOCKET_TRANSPORT_HANDLE, socket_transport);
+```
+
+### socket_transport_create
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_create, SOCKET_TYPE, type);
+```
+
+`socket_transport_create` creates a socket transport.
```
anporumb, line 74: >socket_transport_create [](http://example.com/codeflow?start=1&length=23)

I think there needs to be some sort of verification that `type` is not 0xFF

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_connect` shall connect to a specified endpoint.
+
+If `socket_transport` is `NULL`, `socket_transport_connect` shall fail and return a non-zero value.
+
+If `hostname` is `NULL`, `socket_transport_connect` shall fail and return a non-zero value.
+
+If `port` is `0`, `socket_transport_connect` shall fail and return a non-zero value.
+
+If the transport type is not `SOCKET_CLIENT`, `socket_transport_connect` shall fail and return a non-zero value.
```
anporumb, line 112: >transport [](http://example.com/codeflow?start=7&length=9)

"socket"

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_connect` shall call `sm_open_begin` to begin the open.
+
+If `sm_open_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_connect` shall fail and return a non-zero value.
+
+`socket_transport_connect` shall call `socket` with the params `AF_INET`, `SOCK_STREAM` and `0`.
+
+`socket_transport_connect` shall call `connect` to connect to the endpoint.
+
+`socket_transport_connect` shall set the socket to non-blocking.
```
anporumb, line 122: >set the socket to non-blocking. [](http://example.com/codeflow?start=33&length=31)

missing details how it does that.

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_disconnect` shall disconnect a connected socked from its endpoint.
+
+If `socket_transport` is `NULL`, `socket_transport_disconnect` shall fail and return.
+
+`socket_transport_disconnect` shall call `sm_close_begin` to begin the closing process.
+
+If `sm_close_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_disconnect` shall fail and return.
+
+`socket_transport_disconnect` shall call `close` to disconnect the connected socket.
```
anporumb, line 142: >close [](http://example.com/codeflow?start=42&length=5)

closesocket?

mattdurak:
> # socket transport Win32

can we have the generic description part in the interfaces/devdoc folder? Then there is less copy pasted between win32/linux versions (like this table of client vs. server) and they just need the exposed API and the per-function requirements.

---
Refers to: win32/devdoc/socket_transport_win32.md:1 in c15a3b1. [](commit_id = c15a3b1327bb2fad6245e40ef4b55d83f5be8b0f, deletion_comment = False)

linux/devdoc/socket_transport_linux.md
anporumb, line 142: >close [](http://example.com/codeflow?start=42&length=5)

consider calling shutdown before, because that's the nice thing to do.

[https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-shutdown) tells how to gracefully shutdown...

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_written, uint32_t, flags, void*, overlapped_data);
+```
+
+If `socket_transport` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `payload` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `buffer_count` is `0`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
```
anporumb, line 156: >SOCKET_SEND_ERROR [](http://example.com/codeflow?start=73&length=17)

I think buffer_count==0 and payload==NULL is valid use case which results in 0 bytes over the wire.

buffer_count might be the result of some "higher-up" computation.

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_send` shall call `sm_exec_begin`.
+
+If `sm_exec_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+For each buffer count in payload `socket_transport_send` shall call `send` to send data with `flags`.
```
anporumb, line 162: >For each buffer count in payload [](http://example.com/codeflow?start=0&length=32)

would be nice to coalesce these bytes in 1 buffer OR use scatter/gather APIs to do that... (read: WSASend).

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+If `send` returns a value less then 0, `socket_transport_send` shall stop sending and return `SOCKET_SEND_ERROR`.
```
anporumb, line 164: >send [](http://example.com/codeflow?start=4&length=4)

what happens if there are 3 buffers, and the second calls to send returns a value less than 0?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+- If the errno is equal to ECONNRESET, `socket_transport_send` shall return `SOCKET_SEND_SHUTDOWN`.
```
anporumb, line 166: >errno [](http://example.com/codeflow?start=9&length=5)

IS this the real errno or some sort of GetLastError?

anporumb: Oh wait, we are in linux world here.

linux/devdoc/socket_transport_linux.md
anporumb: actually mremory copy by us would be detrimental, but what about sendmsg?

anporumb: And if we are rewriting these... liburing? :D ./runs away

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+Otherwise `socket_transport_send` shall continue calling send until the data sent is equal to payload.length.
```
anporumb, line 168: >data sent is equal to payload.length. [](http://example.com/codeflow?start=72&length=37)

until buffer_count is reached?

and why does it have "bytes_written" when the "unit of transmission" seems to be "SOCKET_BUFFER"? I would expect to get... the number of socket_buffers send?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+If `sm_close_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_disconnect` shall fail and return.
+
+`socket_transport_disconnect` shall call `close` to disconnect the connected socket.
+
+`socket_transport_disconnect` shall call `sm_close_end`.
+
+### socket_transport_send
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_written, uint32_t, flags, void*, overlapped_data);
```
anporumb, line 149: >flags [](http://example.com/codeflow?start=191&length=5)

unclear how this flags is used

anporumb, line 149: >overlapped_data [](http://example.com/codeflow?start=205&length=15)

unclear how overlapped_data is used.

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+If `socket_transport` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `payload` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `buffer_count` is `0`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+`socket_transport_send` shall call `sm_exec_begin`.
```
anporumb, line 158: >sm_exec_begin [](http://example.com/codeflow?start=36&length=13)

what's the threading here? What happens if 2 socket_transport_send are called at the same time? Will the 2 threads compete over the `send` to the same socket?

anporumb: only asking because if only 1 can execute at the same time then it is sm_barrier...

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+For each buffer count in payload `socket_transport_receive` shall call `recv` with the `payload` flags.
+
+If `recv` a value less then 0, `socket_transport_receive` shall do the following:
+
+- If `errno` is `EAGAIN` or `EWOULDBLOCK`, `socket_transport_receive` shall break out of loop and return `SOCKET_RECEIVE_WOULD_BLOCK`.
+
+- If `errno` is `ECONNRESET`, `socket_transport_receive` shall break out of the looop and return `SOCKET_RECEIVE_SHUTDOWN`.
+
+- else `socket_transport_receive` shall break out of the looop and return `SOCKET_RECEIVE_ERROR`.
```
anporumb, line 198: >looop [](http://example.com/codeflow?start=57&length=5)

loooooop

linux/devdoc/socket_transport_linux.md
jebrando: We shall try to not copy anything.

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+If `payload` is `NULL`, `socket_transport_receive` shall fail and return `SOCKET_RECEIVE_ERROR`.
+
+If `buffer_count` is `0`, `socket_transport_receive` shall fail and return `SOCKET_RECEIVE_ERROR`.
+
+`socket_transport_receive` shall call `sm_exec_begin`.
+
+If `sm_exec_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_receive` shall fail and return `SOCKET_RECEIVE_ERROR`.
+
+For each buffer count in payload `socket_transport_receive` shall call `recv` with the `payload` flags.
```
anporumb, line 190: >`payload` flags [](http://example.com/codeflow?start=87&length=15)

with the what flags now?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+Otherwise `socket_transport_send` shall continue calling send until the data sent is equal to payload.length.
+
+If `bytes_written` is not `NULL`, `socket_transport_send` shall set `bytes_written` the total bytes sent.
+
+`socket_transport_send` shall call `sm_exec_end`.
+
+### socket_transport_receive
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_RECEIVE_RESULT, socket_transport_receive, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_recv, uint32_t, flags, void*, overlapped_data);
```
anporumb, line 177: >bytes_recv [](http://example.com/codeflow?start=172&length=10)

this cannot be `NULL`, can it?

linux/devdoc/socket_transport_linux.md
jebrando: Will do

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_accept, SOCKET_TRANSPORT_HANDLE, socket_transport);
+```
+
+`socket_transport_accept` accepts the incoming connections.
+
+If `socket_transport` is `NULL`, `socket_transport_accept` shall fail and return a non-zero value.
+
+If the transport type is not `SOCKET_SERVER`, `socket_transport_accept` shall fail and return a non-zero value.
```
anporumb, line 250: >non-zero value. [](http://example.com/codeflow?start=96&length=15)

I think NULL, seeing as the return type is a handle...

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_accept` shall call `sm_exec_begin`.
+
+If `sm_exec_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_accept` shall fail and return `SOCKET_SEND_ERROR`.
```
anporumb, line 254: >SOCKET_SEND_ERROR [](http://example.com/codeflow?start=103&length=17)

why is an API called socket_transport_`accept` returning an error code of `SOCKET_SEND_ERROR`?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+`socket_transport_accept` shall call `accept` to accept the incoming socket connection.
+
+`socket_transport_accept` shall set the incoming socket to non-blocking.
```
anporumb, line 258: > set the incoming socket to non-blocking [](http://example.com/codeflow?start=31&length=40)

can it be set after the call to accept?

linux/devdoc/socket_transport_linux.md
```diff
@@ -0,0 +1,284 @@
...
+
+### socket_transport_destroy
+
+```c
+MOCKABLE_FUNCTION(, void, socket_transport_destroy, SOCKET_TRANSPORT_HANDLE, socket_transport);
+```
+
+`socket_transport_destroy` destroys all data stored in the `SOCKET_TRANSPORT_HANDLE` object.
+
+If `socket_transport` is `NULL` `socket_transport_destroy` shall do nothing.
```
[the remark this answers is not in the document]
jebrando: I am going to be  doing this on both the linux and Windows version.

linux/devdoc/socket_transport_linux.md
jebrando: Done.

linux/devdoc/socket_transport_linux.md
jebrando: Sounds good.

linux/devdoc/socket_transport_linux.md
jebrando: I would like to talk to you about this offline.

linux/devdoc/socket_transport_linux.md
jebrando: I will add that as a todo (create a task).  My intention is to move the current functionality to this module.  Then we can improve things.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+If `socket_transport` is `NULL` `socket_transport_destroy` shall do nothing.
+
+`socket_transport_destroy` shall call `sm_destroy` to destroy the sm object.
+
+`socket_transport_destroy` shall free the `SOCKET_TRANSPORT_HANDLE` object.
+
+### socket_transport_connect
+
+```c
+MOCKABLE_FUNCTION(, int, socket_transport_connect, SOCKET_TRANSPORT_HANDLE, socket_transport, const char*, hostname, uint16_t, port, uint32_t, connection_timeout_ms);
```
anporumb, line 101: >connection_timeout_ms [](http://example.com/codeflow?start=143&length=21)

not clear what happnes with this parameter

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+`socket_transport_connect` shall call `connect_to_endpoint` to connect to the endpoint.
+
+If successful `socket_transport_connect` shall call `sm_open_end` with `true`.
+
+If any failure is encountered, `socket_transport_connect` shall call `sm_open_end` with `false`, fail and return a non-zero value.
+
+### connect_to_endpoint
+
+```c
+static int connect_to_endpoint(SOCKET client_socket, const ADDRINFO* addrInfo, uint32_t timeout_usec)
```
anporumb, line 129: >timeout_usec [](http://example.com/codeflow?start=88&length=12)

microsecond?

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+`connect_to_endpoint` shall call the `connect` API to asynchronously connect to the endpoint.
```
anporumb, line 132: >asynchronously [](http://example.com/codeflow?start=54&length=14)

... and how does it tell to the user that it connected? Seeing as it doesn't have a callback or something...

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+
+If the `connect` call fails, `connect_to_endpoint` shall check to `WSAGetLastError` for `WSAEWOULDBLOCK`.
+
+On `WSAEWOULDBLOCK` `connect_to_endpoint` shall call the `select` API with a timeout set in the `zrpc_tcp_client_io_create` function and check the return value:
+
+- If the return is `SOCKET_ERROR`, this indicates a connect and `connect_to_endpoint` shall fail.
+
+- If the return value is 0, this indicates a timeout and `connect_to_endpoint` shall fail.
+
+- Any other value this indicates a possible success and and `connect_to_endpoint` shall test if the socket is writable by calling `FD_ISSET`.
```
anporumb, line 142: >FD_ISSET [](http://example.com/codeflow?start=131&length=8)

Is this a windows thingie? I can find it in Linux man pages:

linux/devdoc/socket_transport_linux.md
jebrando: Currently we are failing the call because it does not send all the data, but we return the amount of data that we have sent.  I will add that in the spec also.

linux/devdoc/socket_transport_linux.md
jebrando: Good catch.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+
+```c
+static int connect_to_endpoint(SOCKET client_socket, const ADDRINFO* addrInfo, uint32_t timeout_usec)
+```
+
+`connect_to_endpoint` shall call the `connect` API to asynchronously connect to the endpoint.
+
+If the `connect` call fails, `connect_to_endpoint` shall check to `WSAGetLastError` for `WSAEWOULDBLOCK`.
+
+On `WSAEWOULDBLOCK` `connect_to_endpoint` shall call the `select` API with a timeout set in the `zrpc_tcp_client_io_create` function and check the return value:
```
anporumb, line 136: >zrpc_tcp_client_io_create [](http://example.com/codeflow?start=97&length=25)

I think zrpc is a layer way above c-pal and should not be mentioned here. Also - nonpublic :(

anporumb (approved):
:shipit:

linux/devdoc/socket_transport_linux.md
jebrando: The linux world is a different place sometimes.

linux/devdoc/socket_transport_linux.md
jebrando: changed.

linux/devdoc/socket_transport_linux.md
jebrando: The flags param is used for the send call.  We pass it through unchanged 142

jebrando: Yeah, I need to change this value.  It's unused in linux.

linux/devdoc/socket_transport_linux.md
jebrando: hmm, that's a good question that will require some research.  I believe sockets are thread-safe, but I shall have to either write a test or look at more documentation.

linux/devdoc/socket_transport_linux.md
jebrando: Yes it can be NULL, in a situation where it's fire and forget, if it's NULL then we just don't return the bytes_recv.

linux/devdoc/socket_transport_linux.md
jebrando: Oh, that's a copy paste error

linux/devdoc/socket_transport_linux.md
jebrando: Yes, it has to be set after accept because the socket get returned in the accept call.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+`socket_transport_create` shall call `sm_create` to create a sm object.
+
+On any failure `socket_transport_create` shall return `NULL`.
+
+On success `socket_transport_create` shall return `SOCKET_TRANSPORT_HANDLE`.
+
+### socket_transport_destroy
+
+```c
+MOCKABLE_FUNCTION(, void, socket_transport_destroy, SOCKET_TRANSPORT_HANDLE, socket_transport);
```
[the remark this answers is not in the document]
jebrando: Oh my goodness, that would be so interesting to do.

win32/devdoc/socket_transport_win32.md
jebrando: I will add a spec for it.

win32/devdoc/socket_transport_win32.md
jebrando: Rename

win32/devdoc/socket_transport_win32.md
jebrando: Wrong word.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+static int connect_to_endpoint(SOCKET client_socket, const ADDRINFO* addrInfo, uint32_t timeout_usec)
+```
+
+`connect_to_endpoint` shall call the `connect` API to asynchronously connect to the endpoint.
+
+If the `connect` call fails, `connect_to_endpoint` shall check to `WSAGetLastError` for `WSAEWOULDBLOCK`.
+
+On `WSAEWOULDBLOCK` `connect_to_endpoint` shall call the `select` API with a timeout set in the `zrpc_tcp_client_io_create` function and check the return value:
+
+- If the return is `SOCKET_ERROR`, this indicates a connect and `connect_to_endpoint` shall fail.
```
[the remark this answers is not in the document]
jebrando: fixed.

win32/devdoc/socket_transport_win32.md
jebrando: Yes this is the way of the Windows.

win32/devdoc/socket_transport_win32.md
jebrando: You are correct.

linux/devdoc/socket_transport_linux.md
jebrando: For now I believe it's OK for multiple threads to send on a sockets.

linux/devdoc/socket_transport_linux.md
jebrando: I will leave it for now because that's the way we have it in zrpc.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
+﻿# socket transport Win32
+
+## Overview
+
+socket_transport_win32 is the module that abstracts the socket transport for the windows platform
+
+The `SOCKET_TYPE` determines which API are applicable for this instance of socket_transport.  The following chart shows which API's are allowed under which type:
+
+| API                                    | Allowable Socket Type
```
mattdurak: We should probably have a common doc in interfaces/devdoc (like the other modules do) for the common stuff like this table

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+|----------------------------------------|----------------------
+| socket_transport_create                | BOTH
```
mattdurak: I kind of have to wonder whether we should enforce this at compile time instead. If we had a `SOCKET_TRANSPORT_CLIENT_HANDLE` and `SOCKET_TRANSPORT_SERVER_HANDLE` and then duplicated the interfaces which say "BOTH" in the table, then you can't call `socket_transport_connect` with `SOCKET_TRANSPORT_SERVER_HANDLE` for example.

Means duplicated function declarations, but the code would be common, and we need the tests for both client and server anyway so it doesn't duplicate too much.

I'm just always in favor of catching errors as close to compile time as possible

win32/devdoc/socket_transport_win32.md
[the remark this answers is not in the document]
jebrando: I don't know if this is necessary or not.  I'll spike it out and see if it makes sense and then I'll update it if it's necessary.  Task https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Dilithium/Dt2?workitem=28467607

win32/devdoc/socket_transport_win32.md
mattdurak: "and and"

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+
+MU_DEFINE_ENUM(SOCKET_RECEIVE_RESULT, SOCKET_RECEIVE_RESULT_VALUES)
+
+#define SOCKET_TYPE_VALUES \
+    SOCKET_CLIENT, \
+    SOCKET_SERVER
+
+MU_DEFINE_ENUM(SOCKET_TYPE, SOCKET_TYPE_VALUES)
+
+typedef struct SOCKET_BUFFER_TAG
```
mattdurak: Interesting, no CONSTBUFFER here? I guess that makes sense when you don't need the reference counting in this layer

interfaces/inc/c_pal/socket_transport.h
```diff
@@ -0,0 +1,67 @@
...
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_create, SOCKET_TYPE, type);
+MOCKABLE_FUNCTION(, void, socket_transport_destroy, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, int, socket_transport_connect, SOCKET_TRANSPORT_HANDLE, socket_transport, const char*, hostname, uint16_t, port, uint32_t, connection_timeout_ms);
+MOCKABLE_FUNCTION(, int, socket_transport_listen, SOCKET_TRANSPORT_HANDLE, socket_transport, uint16_t, port);
+MOCKABLE_FUNCTION(, void, socket_transport_disconnect, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_accept, SOCKET_TRANSPORT_HANDLE, socket_transport);
+
+MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_written, uint32_t, flags, void*, data);
```
mattdurak: curious what "flags" and "data" would mean in a PAL sense. Presumably we need to pass different things in Windows and Linux, which isn't very nice. Would be better if those were abstracted and then translated to the platform-specifics in this module. E.g. if we had our own PAL flags for all the things we need, then we convert it to win32 values or Linux equivalents as needed.

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+
+If `socket_transport` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `payload` is `NULL`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+If `buffer_count` is `0`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
+
+`socket_transport_send` shall call `sm_exec_begin`.
+
+If `sm_exec_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_send` shall fail and return `SOCKET_SEND_ERROR`.
```
mattdurak: Can we fail with something like SOCKET_CLOSED to be explicit here with the difference between "something failed to send" and "we didn't even try because we know the socket is not connected"?

mattdurak: Well, I guess we have SOCKET_SEND_FAILED below to differentiate, but still seems that a "NOT_OPEN" or "CLOSED" value may be useful for layers above (vs. invalid args above this).

win32/devdoc/socket_transport_win32.md
```diff
@@ -0,0 +1,296 @@
...
+
+`socket_transport_accept` shall call `sm_exec_begin`.
+
+If `sm_exec_begin` does not return `SM_EXEC_GRANTED`, `socket_transport_accept` shall fail and return `SOCKET_SEND_ERROR`.
+
+`socket_transport_accept` shall call `select` determine if the socket is ready to be read passing a timeout of 10 milliseconds.
+
+`socket_transport_accept` shall call `accept` to accept the incoming socket connection.
+
+`socket_transport_accept` shall allocate a `SOCKET_TRANSPORT` for the incoming connection and call `sm_create` and `sm_open` on the connection.
```
mattdurak: What is the transport type of the returned socket? I assume it's actually `SOCKET_CLIENT`, but should specify here (or if you take my other comment, this just returns a client type).

Is `SOCKET_CLIENT` the right name though? Isn't it really "SOCKET_LISTENER" and "SOCKET_DATA" or something like that? Not really a client/server difference, just a difference in whether you are sending/receiving or you are listening/accepting on the socket.

jebrando: Very true about the name.  I'll specify it in the spec.

interfaces/devdoc/socket_transport_requirements.md
```diff
@@ -0,0 +1,139 @@
...
+```c
+MOCKABLE_FUNCTION(, void, socket_transport_disconnect, SOCKET_TRANSPORT_HANDLE, socket_transport);
+```
+
+`socket_transport_disconnect` shall disconnect a connected socked from its endpoint.
+
+### socket_transport_send
+
+```c
+MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_written, uint32_t, flags, void*, overlapped_data);
```
mattdurak, line 103: Can we at least mention what flags/overlapped_data "means" here

mattdurak: Same below

jebrando: Yes, they are pass-through, I'll add a mention here.

win32/devdoc/socket_transport_win32.md
jebrando: Added.

win32/devdoc/socket_transport_win32.md
jebrando: I had this, at first in the spike, but it starts to get messy because most of the code is common, only 3 functions are different (connect, listen, accept).  It's a lot more code just for 3 function.

win32/devdoc/socket_transport_win32.md
jebrando: And CONSTBUFFER is in c-util and not in PAL.  I thought about moving it, but like you said no ref counting.

interfaces/inc/c_pal/socket_transport.h
jebrando: Currently for flags we only use 1 value here for linux and windows it's 0.  For Data, I thought about an object, but it's only  in windows and in linux it's NULL so I didn't want to over think it.

win32/devdoc/socket_transport_win32.md
jebrando: Will do.

BertKleewein (approved):

PR #344, opened 2024-06-21
Initial check in for socket transport linux

jebrando:


darobs:
>     SOCKET_RECEIVE_OK, \

Add SOCKET_RECEIVE_INVALID_ARG

---
Refers to: linux/devdoc/socket_transport_linux_requirements.md:26 in 69bf383. [](commit_id = 69bf383718361e01ae3bd5eb11be5ea259c31e24, deletion_comment = False)

linux/devdoc/socket_transport_linux_requirements.md
```diff
@@ -65,15 +66,15 @@ MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_create, SOCKET_TYP
 
 `socket_transport_create` creates a socket transport.
 
-`socket_transport_create` shall ensure `type` is either `SOCKET_CLIENT`, or `SOCKET_SERVER`.
+**SOCKET_TRANSPORT_LINUX_11_001: [** `socket_transport_create` shall ensure `type` is either `SOCKET_CLIENT`, or `SOCKET_SERVER`. **]**
```
dcristoloveanu, line 69: I think this spec should normally read "If type is not SOCKET_CLIENT or SOCKET_SERVER, then socket_transport_create shall fail and return NULL" - just so it is clear what will be returned.

linux/src/socket_transport_linux.c
```diff
@@ -0,0 +1,696 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#include <stdlib.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+#include <arpa/inet.h>
```
dcristoloveanu, line 8: oooh, have not seen arpa in some time now.

linux/tests/socket_transport_linux_ut/socket_transport_linux_ut.c
```diff
@@ -0,0 +1,1485 @@
...
+    umock_c_reset_all_calls();
+    umock_c_negative_tests_init();
+}
+
+TEST_FUNCTION_CLEANUP(cleanup)
+{
+    umock_c_negative_tests_deinit();
+
+    //ASSERT_ARE_EQUAL(uint32_t, 0, test_socket, "socket were not matched with close");
+}
```
dcristoloveanu, line 203: Commented out code detected

linux/tests/socket_transport_linux_ut/socket_transport_linux_ut.c
```diff
@@ -0,0 +1,1485 @@
...
+
+    REGISTER_UMOCK_ALIAS_TYPE(SM_HANDLE, void*);
+    REGISTER_UMOCK_ALIAS_TYPE(SOCKET_HANDLE, int);
+    REGISTER_UMOCK_ALIAS_TYPE(ssize_t, long);
+    REGISTER_UMOCK_ALIAS_TYPE(socklen_t, int);
+
+    g_addrInfo.ai_canonname = "ai_canonname";
+    for (size_t index = 0; index < TEST_BYTES_RECV; index++)
+    {
+        recv_buff[index] = 'a';
```
dcristoloveanu, line 182: Consider initializing with something more aggressive for testing - like 'a', 'b', ...

linux/src/socket_transport_linux.c
```diff
@@ -0,0 +1,696 @@
...
+    SOCKET_HANDLE client_socket = socket(AF_INET, SOCK_STREAM, 0);
+    if (client_socket == INVALID_SOCKET)
+    {
+        LogErrorNo("Failure: socket create failure.");
+        result = INVALID_SOCKET;
+    }
+    else
+    {
+        char port_string[16];
+        if (sprintf(port_string, "%u", port) < 0)
```
dcristoloveanu, line 85: PRIu16

linux/src/socket_transport_linux.c
```diff
@@ -0,0 +1,696 @@
...
+        }
+    }
+    return result;
+}
+
+static SOCKET_HANDLE connect_to_client(const char* hostname, uint16_t port, uint32_t connection_timeout)
+{
+    SOCKET_HANDLE result;
+    // Codes_SOCKET_TRANSPORT_LINUX_11_015: [ socket_transport_connect shall call socket with the params AF_INET, SOCK_STREAM and 0. ]
+    SOCKET_HANDLE client_socket = socket(AF_INET, SOCK_STREAM, 0);
```
dcristoloveanu, line 76: I doubt that we need both result and client_socket

linux/src/socket_transport_linux.c
```diff
@@ -0,0 +1,696 @@
...
+                freeaddrinfo(addrInfo);
+            }
+        }
+        close(client_socket);
+    }
+all_ok:
+    return result;
+}
+
+SOCKET_TRANSPORT_HANDLE socket_transport_create(SOCKET_TYPE type)
```
dcristoloveanu, line 136: Generally having 2 APIs socket_transport_create_server and socket_transport_create_client is better, because the check is done at compile time. As in one cannot compile some code with some weird type passed in.
The only case where such a construct would be beneficial is for interfaces where there will be lots of different values added (like a set_option API).
But in all other cases, which is most, it's better to have separate APIs which make it explicitly checked at compile time.

Code will work with this too, but I'd say strong consider.

linux/src/socket_transport_linux.c
```diff
@@ -0,0 +1,696 @@
...
+                sm_exec_end(socket_transport->sm);
+            }
+        }
+        result = NULL;
+    }
+all_ok:
+    return result;
+}
+
+SOCKET_HANDLE socket_transport_get_underlying_socket(SOCKET_TRANSPORT_HANDLE socket_transport)
```
dcristoloveanu, line 667: I have to question why is this API needed...

dcristoloveanu: It's better if the underlying socket is not exposed - so there should be a very strong case for why it is needed

jebrando: You have to have it exposed for epoll on linux and Overlapped on Windows.  They both require the raw socket.

linux/src/socket_transport_linux.c
jebrando: https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Dilithium/Dt2?workitem=28591226

samrach9 (approved):

PR #345, opened 2024-06-26
Add PRI macros for T_PTR

parth21999:


BertKleewein (approved):
:shipit:

common/inc/c_pal/thandle_ptr.h
```diff
@@ -31,6 +31,12 @@ typedef struct PTR_STRUCT_TAG_TYPE_NAME(T)
     THANDLE_PTR_FREE_FUNC_TYPE_NAME(T) dispose; /*original dispose passed to THANDLE_PTR_CREATE_WITH_MOVE*/     \
 } PTR(T);
 
+#define PRI_T_PTR "p"
+
+#define T_PTR_VALUE(t_ptr) ((t_ptr)->pointer)
+
+#define T_PTR_VALUE_OR_NULL(t_ptr) ((t_ptr) == NULL ? NULL : (t_ptr)->pointer)
```
jebrando, line 38: Why don't you just combine these and have 1 define 

T_PTR_VALUE(t_ptr) ((t_ptr) == NULL ? NULL : (t_ptr)->pointer)

jebrando: This would make it so it never will crash is you accidently called the wrong one.

jebrando (approved):
:shipit:

common/inc/c_pal/thandle_ptr.h
parth21999: We have different `OR_NULL` print macros elsewhere in the repo. This allows the user to be more specific.

If the user thinks that the value cannot be NULL in a certain situation, then it is better to crash then to continue and get in a weird state.

PR #349, opened 2024-07-09
Socket transport sr

samrach9:


interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+    CHAOS_TEST_SOCKETS* chaos_test = parameter;
+
+    chaos_test->listen_socket = socket_transport_create_server();
+    ASSERT_IS_NOT_NULL(chaos_test->listen_socket);
+
+    ASSERT_ARE_EQUAL(int, 0, socket_transport_listen(chaos_test->listen_socket, g_port_num));
+
+    for (i = 0; i < CHAOS_THREAD_COUNT; i++)
+    {
+        ASSERT_ARE_EQUAL(int, 0, socket_transport_connect(chaos_test->client_socket_handles[i], "localhost", g_port_num, 10000));
```
jebrando, line 277: >10000 [](http://example.com/codeflow?start=121&length=5)

Best to use the define TEST_CONN_TIMEOUT here.

interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+
+        ASSERT_IS_NOT_NULL(chaos_knight_test.client_socket_handles[i]);
+
+    }
+
+    THREAD_HANDLE thread;
+
+    interlocked_exchange(&chaos_knight_test.API_create_result, 0);
+    ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Create(&thread, thread_worker_func, &chaos_knight_test));
+    wait_for_value(&chaos_knight_test.API_create_result, 1);
```
jebrando, line 313: After looking at this, we don't need to have this in a thread since we're waiting for the values.  Just change thread_worker_func to be a normal function and call it normally.

interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+
+    for (size_t i = 0; i < CHAOS_THREAD_COUNT; i++)
```
jebrando, line 315: From here to line 392 can be combined with just another loop so the code isn't repeated.

interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+
+            free(send_data[inner].buffer);
+            free(recv_data[inner].buffer);
+        }
+        free(send_data);
+        free(recv_data);
+    }
+
+    // server disconnects
+    socket_transport_disconnect(chaos_knight_test.incoming_socket_handles[FAILED_SERVER_NUM]);
```
jebrando, line 349: >FAILED_SERVER_NUM [](http://example.com/codeflow?start=74&length=17)

Failed server number should be a random number from 0 to CHAOS_THREAD_COUNT so any of them can be disconnected

interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+        SOCKET_SEND_RESULT send_result = socket_transport_send(chaos_knight_test.client_socket_handles[i], send_data, buffer_count, &bytes_sent, flag, NULL);
+        if (send_result == SOCKET_SEND_OK)
+        {
+            ASSERT_ARE_EQUAL(int, total_bytes, bytes_sent);
+        }
+
+        if (i == FAILED_SERVER_NUM)
+        {
+            ASSERT_ARE_EQUAL(SOCKET_RECEIVE_RESULT, SOCKET_RECEIVE_ERROR, socket_transport_receive(chaos_knight_test.incoming_socket_handles[i], recv_data, buffer_count, &bytes_recv, 0, NULL));
+            ASSERT_ARE_NOT_EQUAL(int, total_bytes, bytes_recv);
```
jebrando, line 377: >ASSERT_ARE_NOT_EQUAL(int, total_bytes, bytes_recv); [](http://example.com/codeflow?start=12&length=51)

If the above function fail, then this call is invalid and doesn't make any sense.

interfaces/tests/socket_transport_int/socket_transport_int.c
```diff
@@ -234,4 +253,160 @@ TEST_FUNCTION(send_and_receive_random_buffer_of_random_byte_succeeds)
...
+
+        for (uint32_t inner = 0; inner < buffer_count; inner++)
+        {
+            ASSERT_ARE_EQUAL(int, 0, memcmp(send_data[inner].buffer, recv_data[inner].buffer, send_data[inner].length), "mismatch data from array index %" PRIu32 "", inner);
+
+            free(send_data[inner].buffer);
+            free(recv_data[inner].buffer);
+        }
+        free(send_data);
+        free(recv_data);
```
jebrando, line 345: /consider

This code could be in a thread for each and you'd have a bunch of threads sending and receiving then you close one and it should stop and the others should continue for a little bit.

jebrando: You don't have to do this right away, just think about it.

jebrando (approved):
:shipit:

PR #359, opened 2024-08-15
adding two functions to socket_transport API

samrach9:


interfaces/inc/c_pal/socket_transport.h
```diff
@@ -69,6 +70,7 @@ MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_
 MOCKABLE_FUNCTION(, SOCKET_RECEIVE_RESULT, socket_transport_receive, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_recv, uint32_t, flags, void*, data);
 
 MOCKABLE_FUNCTION(, SOCKET_HANDLE, socket_transport_get_underlying_socket, SOCKET_TRANSPORT_HANDLE, socket_transport);
+MOCKABLE_FUNCTION(, int, socket_transport_check_valid_handle, SOCKET_TRANSPORT_HANDLE, socket_transport_handle);
```
jebrando, line 73: >socket_transport_check_valid_handle [](http://example.com/codeflow?start=25&length=35)

Technically you're not checking that the handle is value, your checking if the underlying socket is valid, probably should change the name to

socket_transport_check_valid_socket

linux/src/socket_transport_linux.c
```diff
@@ -191,6 +191,42 @@ SOCKET_TRANSPORT_HANDLE socket_transport_create_server(void)
...
+            LogError("failure allocating SOCKET_TRANSPORT: %zu", sizeof(SOCKET_TRANSPORT));
+        }
+        else
+        {
+            result->sm = sm_create("Socket_transport_win32");
+            if (result->sm == NULL)
+            {
+                LogError("sm_create failed.");
+            }
+            else
```
jebrando, line 216: >else [](http://example.com/codeflow?start=12&length=4)

You need to call sm_open_begin here just like you do on line 703 and then you need to call sm_open_end with true.

linux/src/socket_transport_linux.c
```diff
@@ -728,4 +764,27 @@ SOCKET_HANDLE socket_transport_get_underlying_socket(SOCKET_TRANSPORT_HANDLE soc
...
+            LogError("Invalid socket in argument: SOCKET_TRANSPORT_HANDLE socket_transport_handle: %p", socket_transport_handle);
+            result = -1;
+        }
+        else
+        {
+            result = 0;
+        }
+    }
+    return result;
 }
```
jebrando, line 790: Always end the file with a newline.

interfaces/inc/c_pal/socket_transport.h
jebrando, line 73: >int [](http://example.com/codeflow?start=20&length=3)

You probably should return a bool here, so it's a valid or not.

jebrando: Once you add bool, you will need to include <stdbool.h>, search in the repo for it to see how you add that to a header.

linux/devdoc/socket_transport_linux_requirements.md
```diff
@@ -64,6 +65,7 @@ MOCKABLE_FUNCTION(, SOCKET_SEND_RESULT, socket_transport_send, SOCKET_TRANSPORT_
 MOCKABLE_FUNCTION(, SOCKET_RECEIVE_RESULT, socket_transport_receive, SOCKET_TRANSPORT_HANDLE, socket_transport, SOCKET_BUFFER*, payload, uint32_t, buffer_count, uint32_t*, bytes_recv, uint32_t, flags, void*, data);
 
 MOCKABLE_FUNCTION(, SOCKET_HANDLE, socket_transport_get_underlying_socket, SOCKET_TRANSPORT_HANDLE, socket_transport);
+MOCKABLE_FUNCTION(, int, socket_transport_check_valid_handle, SOCKET_TRANSPORT_HANDLE, socket_transport_handle);
```
jebrando, line 68: >int [](http://example.com/codeflow?start=20&length=3)

Need to fix in the spec.

win32/src/socket_transport_win32.c
```diff
@@ -760,3 +812,24 @@ SOCKET_HANDLE socket_transport_get_underlying_socket(SOCKET_TRANSPORT_HANDLE soc
     }
     return result;
 }
+
+bool socket_transport_is_valid_socket(SOCKET_TRANSPORT_HANDLE socket_transport_handle)
+{
+    bool result = false;
```
jebrando, line 818: >bool result = false; [](http://example.com/codeflow?start=4&length=20)

Nitpick:  There's a bunch of schools of thought about assigning a result variable once you create it.  In general you probably want to not assign it during creation so the compiler will let you know if there is a branch of code that you forgotten to assign the return value.  This is a little function so it doesn't matter to much, but not assigning has saved me a few times, so it's just a good practice to get into.

jebrando (approved):
:shipit:

jebrando:
> MOCKABLE_FUNCTION(, SOCKET_TRANSPORT_HANDLE, socket_transport_create_client);

Add the new functions in this spec.

---
Refers to: interfaces/devdoc/socket_transport_requirements.md:61 in c9307f4. [](commit_id = c9307f41f849632fb980b92dfa66d663966b5fa3, deletion_comment = False)

PR #362, opened 2024-08-28
Initial checkin with specs

nishudeshu:
Initial Check-In with Specs for Win32 and Linux


nishudeshu:
Part of Story https://msazure.visualstudio.com/One/_workitems/edit/29233939

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -31,6 +31,12 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
 MOCKABLE_FUNCTION(, int, threadpool_open, THANDLE(THREADPOOL), threadpool);
 MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 
+MOCKABLE_FUNCTION(, PVOID, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, PVOID, work_function_context);
+
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, PVOID, work_item_context);
+
+MOCKABLE_FUNCTION(, void, threadpool_work_context_destroy, PVOID, work_item_context);
```
mattdurak, line 38: >threadpool_work_context_destroy [](http://example.com/codeflow?start=26&length=31)

poetry: should likely be `threadpool_destroy_work_item`

mattdurak:
> typedef struct TIMER_INSTANCE_TAG* TIMER_INSTANCE_HANDLE;

We should have a new opaque type here like `typedef struct THREADPOOL_WORK_ITEM_TAG* THREADPOOL_WORK_ITEM_HANDLE`

---
Refers to: interfaces/devdoc/threadpool_requirements.md:23 in 31fb301. [](commit_id = 31fb301b28230275223d8dae1b369676b397e158, deletion_comment = False)

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -31,6 +31,12 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
 MOCKABLE_FUNCTION(, int, threadpool_open, THANDLE(THREADPOOL), threadpool);
 MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 
+MOCKABLE_FUNCTION(, PVOID, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, PVOID, work_function_context);
```
mattdurak, line 34: >PVOID [](http://example.com/codeflow?start=20&length=5)

And this should return the `THREADPOOL_WORK_ITEM_HANDLE`, which should be the argument type for the other two functions (instead of `PVOID`)

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, PVOID, work_item_context);
+```
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool.
+
+**NON_THREADPOOL_05_008: [** If `threadpool` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_009: [** If `work_function` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_010: [** If `ptp_work` within the context is `NULL`, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
```
mattdurak, line 147: This should not be possible as we should not return a handle in `threadpool_create_work_item` with a `NULL` `ptp_work` and there is no other way to get these handles. So we can't have this happen/can't write a test for it/don't need the spec.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_002: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_003: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_004: [** Otherwise `threadpool_create_work_item` shall allocate a context where `work_function`, `work_function_context`, and `ptp_work` shall be saved. **]**
+
+**NON_THREADPOOL_05_005: [** If any error occurs, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_006: [** `threadpool_create_work_item` shall create `PTP_WORK` by calling `CreateThreadpoolWork` to set the callback function as `on_work_callback_v2` and pass the newly created context whenever the callback is called in the tp_environment. **]**
```
mattdurak, line 131: >PTP_WORK [](http://example.com/codeflow?start=73&length=8)

We should scrub the references to PTP_WORK, `CreateThreadpoolWork`, and all of the win32 API's/types from this file.

The "interfaces/devdoc" requirements file should just describe the generic interface and what it is expected to do.

Ideally (I'm not sure if we do), we should be able to just take these specs and write int tests against "all" (windows/Linux) platforms.

Then the win32 API specifics belong in "win32/devdoc/threadpool_win32_requirements.md". That file should have these specifics (like calling `CreateThreadpoolWork` and `SubmitThreadpoolWork`, etc.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool.
+
+**NON_THREADPOOL_05_008: [** If `threadpool` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_009: [** If `work_function` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_010: [** If `ptp_work` within the context is `NULL`, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_011: [** `threadpool_schedule_work_item` shall call `SubmitThreadpoolWork` to submit the work item for execution. **]**
```
mattdurak, line 149: >SubmitThreadpoolWork` to submit the work item for execution. **]** [](http://example.com/codeflow?start=73&length=66)

As an example of my comment above, this spec is good for the win32 specs. In this file it should say something like "shall schedule the work item to be run in the threadpool". In an integration test for all platforms, we could validate that (make sure the work function is called after calling the schedule function), but the SubmitThreadpoolWork won't happen on Linux.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
 
 **NON_THREADPOOL_01_019: [** If `threadpool` is not OPEN, `threadpool_close` shall return. **]**
 
+### threadpool_create_work_item
+
+```c
+MOCKABLE_FUNCTION(, PVOID, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, PVOID, work_function_context);
+```
+
+`threadpool_create_work_item` creates a work item to be executed by the threadpool.
```
mattdurak, line 119: >uted by the threadpool. [](http://example.com/codeflow?start=60&length=23)

Here or somewhere we should document *why* we have this separate API. Like "This function separates the creation of the work item from scheduling it so that `threadpool_schedule_work_item` can be called later and cannot fail."

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_007: [** If any error occurs, `threadpool_create_work_item` shall fail, free the newly created context and return a `NULL` value. **]**
+
+### threadpool_schedule_work_item
+
+```c
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, PVOID, work_item_context);
+```
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool.
```
mattdurak, line 141: >cuted by the threadpool [](http://example.com/codeflow?start=63&length=23)

Maybe should have a note that this may be called multiple times on the same `THREADPOOL_WORK_ITEM_HANDLE`.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_011: [** `threadpool_schedule_work_item` shall call `SubmitThreadpoolWork` to submit the work item for execution. **]**
+
+### threadpool_work_context_destroy
+
+```c
+MOCKABLE_FUNCTION(, void, threadpool_work_context_destroy, PVOID, work_item_context);
+```
+
+`threadpool_work_context_destroy` closes the `ptp_work` member variable in `work_item_context` of type `PWORK_ITEM_CONTEXT_V2` and then frees the `work_item_context`
```
mattdurak, line 157: >closes the `ptp_work` member variable in `work_item_context` of type `PWORK_ITEM_CONTEXT_V2` and then frees the `work_item_context` [](http://example.com/codeflow?start=34&length=131)

As a generic interface description this "frees the work item created by `threadpool_create_work_item`".

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_012: [** If `work_item_context` is `NULL`, `threadpool_work_context_destroy` shall fail and not do anything before returning. **]**
+
+**NON_THREADPOOL_05_013: [** If `ptp_work` within the `work_item_context` is not `NULL`, then `threadpool_work_context_destroy` shall call `CloseThreadpoolWork`. **]**
+
+**NON_THREADPOOL_05_014: [** If `ptp_work` is `NULL`, `threadpool_work_context_destroy` shall NOT call `CloseThreadpoolWork`. **]**
```
mattdurak, line 163: >**NON_THREADPOOL_05_014: [** If `ptp_work` is `NULL`, [](http://example.com/codeflow?start=0&length=53)

also shouldn't be possible

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,60 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+```c
+MOCKABLE_FUNCTION(, void, threadpool_work_context_destroy, PVOID, work_item_context);
+```
+
+`threadpool_work_context_destroy` closes the `ptp_work` member variable in `work_item_context` of type `PWORK_ITEM_CONTEXT_V2` and then frees the `work_item_context`
+
+**NON_THREADPOOL_05_012: [** If `work_item_context` is `NULL`, `threadpool_work_context_destroy` shall fail and not do anything before returning. **]**
+
+**NON_THREADPOOL_05_013: [** If `ptp_work` within the `work_item_context` is not `NULL`, then `threadpool_work_context_destroy` shall call `CloseThreadpoolWork`. **]**
```
mattdurak, line 161: >If `ptp_work` within the `work_item_context` is not `NULL` [](http://example.com/codeflow?start=29&length=58)

Can remove this condition

interfaces/inc/c_pal/threadpool.h
```diff
@@ -22,7 +22,7 @@ extern "C" {
 
 typedef struct THREADPOOL_TAG THREADPOOL;
 typedef struct TIMER_INSTANCE_TAG* TIMER_INSTANCE_HANDLE;
-
+typedef void* PVOID;
```
mattdurak, line 25: should remove

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +249,63 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+
+**SRS_THREADPOOL_WIN32_05_003: [** The `work_function` callback passed to `threadpool_create_work_item` shall be called with the `work_function_context` as an argument. `work_function_context` was set inside the `threadpool_create_work_item` as an argument to `CreateThreadpoolContext`. **]**
+
+```c
+MOCKABLE_FUNCTION(, PVOID, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, PVOID, work_function_context);
+```
+
+`threadpool_create_work_item` creates a work item to be executed by the threadpool.
+
+**SRS_THREADPOOL_WIN32_05_004: [** `threadpool_create_work_item` shall initialize the `work_item_context` to `NULL`, where `NULL` would be a failure indicator to the calling functions **]**
```
mattdurak, line 271: This is not needed, spec 05_008 covers it

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +249,63 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+
+**SRS_THREADPOOL_WIN32_05_005: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**SRS_THREADPOOL_WIN32_05_006: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**SRS_THREADPOOL_WIN32_05_007: [** Otherwise `threadpool_create_work_item` shall allocate a context where `work_function`, `work_function_context`, and `ptp_work` shall be saved. **]**
+
+**SRS_THREADPOOL_WIN32_05_008: [** If any error occurs, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**SRS_THREADPOOL_WIN32_05_009: [** `threadpool_create_work_item` shall create `PTP_WORK` by calling `CreateThreadpoolWork` to set the callback function as `on_work_callback_v2` and pass the newly created context whenever the callback is called in the tp_environment. **]**
```
mattdurak, line 281: >and pass the newly created context whenever the callback is called in the tp_environment [](http://example.com/codeflow?start=177&length=88)

I don't think we need this part of the spec ("and pass...") because that is 05_003

interfaces/devdoc/threadpool_requirements.md
mattdurak: And also worth noting that this cannot fail unless the arguments are `NULL`.

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +249,63 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, PVOID, work_item_context);
+```
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool.
+
+**SRS_THREADPOOL_WIN32_05_011: [** If `threadpool` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**SRS_THREADPOOL_WIN32_05_012: [** If `work_function` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**SRS_THREADPOOL_WIN32_05_013: [** If `ptp_work` within the context is `NULL`, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
```
mattdurak, line 295: >5_013: [** If `ptp_work` within the context is `NULL`, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]** [](http://example.com/codeflow?start=24&length=134)

same comment here and below that the ptp_work cannot be null

win32/src/threadpool_win32.c
```diff
@@ -265,6 +286,59 @@ void threadpool_close(THANDLE(THREADPOOL) threadpool)
     }
 }
 
+/* threadpool_create_work_item */
+
+/* Codes_SRS_THREADPOOL_WIN32_05_004: [ threadpool_create_work_item shall initialize the work_item_context to NULL, where NULL would be a failure indicator to the calling functions ] */
```
mattdurak, line 291: >Codes_SRS_THREADPOOL_WIN32_05_004 [](http://example.com/codeflow?start=3&length=33)

for new code, we typically do something in the MD like change it to "S_R_S" or something so that traceability tool doesn't flag it as an error.

That is better than adding specs to the code/test which are not yet implemented.

mattdurak: So we should remove all of the specs in the .c files for now until it is implemented

win32/src/threadpool_win32.c
```diff
@@ -265,6 +286,59 @@ void threadpool_close(THANDLE(THREADPOOL) threadpool)
...
+
+/* Codes_SRS_THREADPOOL_WIN32_05_008 : [ If any error occurs, threadpool_create_work_item shall fail and return a NULL value. ] */
+
+/* Codes_SRS_THREADPOOL_WIN32_05_009 : [ threadpool_create_work_item shall create PTP_WORK by calling CreateThreadpoolWork to set the callback function as on_work_callback_v2 and pass the newly created context whenever the callback is called in the tp_environment. ] */
+
+/* Codes_SRS_THREADPOOL_WIN32_05_010 : [ If any error occurs, threadpool_create_work_item shall fail, free the newly created context and return a NULL value. ] */
+
+PVOID threadpool_create_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WORK_FUNCTION work_function, PVOID work_function_context)
+{
+    LogWarning("Printing %p %p %p", threadpool, work_function, work_function_context);
```
mattdurak, line 307: >LogWarning("Printing %p %p %p", threadpool, work_function, work_function_context); [](http://example.com/codeflow?start=4&length=82)

this is fine, alternatively we will often just put lines like:

```
(void)threadpool;
(void)work_function;
(void)work_function_context;
```

For the otherwise unused variables

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -381,3 +387,63 @@ static int threadpool_work_func(void* param);
...
+**SRS_THREADPOOL_LINUX_05_015: [** `work_item_context` can be a `NULL` value **]**
+
+**SRS_THREADPOOL_LINUX_05_016: [** If `threadpool` is `NULL`, `threadpool_schedule_work_item` shall fail and return a non-zero value. **]**
+
+**SRS_THREADPOOL_LINUX_05_017: [** `threadpool_schedule_work_item` shall call `sm_exec_begin`. **]**
+
+**SRS_THREADPOOL_LINUX_05_018: [** If `sm_exec_begin` returns `SM_EXEC_REFUSED`, `threadpool_schedule_work_item` shall fail and return a non-zero value. **]**
+
+**SRS_THREADPOOL_LINUX_05_019: [** `threadpool_schedule_work_item` shall unblock the `threadpool` semaphore by calling `sem_post`. **]**
+
```
mattdurak, line 440: likely this will have a problem in that we can't call `threadpool_schedule_work_item` multiple times.

I think we'll need to fix that in the Linux implementation.

interfaces/devdoc/threadpool_requirements.md
mattdurak: Actually, calling it multiple times should be a spec (and we should have int tests for it :) )

interfaces/devdoc/threadpool_requirements.md
mattdurak: I'm not sure I follow...we need to have some function exposed here to do that cleanup, and my comment was just that the name (poetry) of this function should be `threadpool_destroy_work_item` to match the other function names.

nishudeshu: I am just saying that work_item aka PTP_WORK is different than WORK_ITEM_CONTEXT. Currently in the `on_work_callback`, both PTP_WORK and WORK_ITEM_CONTEXT is getting freed for every scheduling call. We will have to remove the below statements and free them. In the new implementation PTP_WORK will be saved inside WORK_ITEM_CONTEXT, hence I would call the function poetically as `threadpool_destroy_work_item_context` instead of just `threadpool_destroy_work_item`

```


/* Codes_SRS_THREADPOOL_WIN32_01_038: [ on_work_callback shall call CloseThreadpoolWork. ]*/
CloseThreadpoolWork(work);

/* Codes_SRS_THREADPOOL_WIN32_01_039: [ on_work_callback shall free the context allocated in threadpool_schedule_work. ]*/
free(context);
```

interfaces/devdoc/threadpool_requirements.md
nishudeshu: We are returning the pointer to `WORK_ITEM_CONTEXT_V2` which contains `work_function`, `work_function_context`. and `ptp_work` and saving it in the Queue Processor or any caller for that matter. 
In either case WORK ITEM or WORK ITEM CONTEXT are not available in Linux, hence using PVOID for return and argument type.
`THREADPOOL_WORK_FUNCTION` is available in both WIN32 and LINUX

interfaces/devdoc/threadpool_requirements.md
nishudeshu: These checks are needed in case the variable is accidentally assigned `NULL` after creation of the handles or changed to `NULL` during transmission by a hacker.

interfaces/devdoc/threadpool_requirements.md
nishudeshu: Ok

interfaces/devdoc/threadpool_requirements.md
nishudeshu: Ok

interfaces/devdoc/threadpool_requirements.md
mattdurak: I think the terminology here is the problem. It's a difference between the *interface* and the *implementation*.

As a "user" of the threadpool interface, I don't care about how threadpool stores its state and I don't know what a context or ptp_work is. I just know that I create some "thing" (`threadpool_create_work_item`) and when I'm done with it I call something to clean it up (`threadpool_destroy_work_item`).

In the implementation, yes, you are going to free the ptp_work and free the context. But that's an implementation detail. You are going to return the context as `THREADPOOL_WORK_ITEM_HANDLE` and take that handle for all the other calls.

interfaces/devdoc/threadpool_requirements.md
mattdurak: So generally, our conventions are to provide a header with an opaque type and then functions to operate on it.

As in we have `typedef struct SOMETHING* SOMETHING_HANDLE`.

"users" of the module cannot create a "struct SOMETHING", they don't know anything about the data it stores. They just get a pointer (SOMETHING_HANDLE) and can pass that into the functions. This gives you the minimal type checking that C has (not the same level as C++ but better than nothing) and it encapsulates the implementation.

We should do the same thing for the work item. The interface should have some opaque type which is used for both windows and linux. The implementation can be different, just like the `struct THREADPOOL_TAG` we have now.

interfaces/devdoc/threadpool_requirements.md
mattdurak: This is another reason for opaque handles.

No one can change the data fields outside of threadpool_win32.c (or threadpool_linux.c). The type is encapsulated.

Yes, you could cast it to something, modify the memory directly, etc. But we have to assume that the callers to `threadpool_schedule_work_item` are not adversarial. This module is not at a security boundary and isn't going to get random pointer input.

We can't really guard against "hackers" to this call by checking for NULL anyway. What if the field was changed to "1" (or any other invalid value) instead of NULL?

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +249,63 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+static VOID CALLBACK on_work_callback_v2(PTP_CALLBACK_INSTANCE instance, PVOID context, PTP_TIMER timer);
+```
+
+`on_work_callback_v2` executes the work function passed to `threadpool_create_work_item`.
+
+**SRS_THREADPOOL_WIN32_05_001: [** If `context` is `NULL`, `on_work_callback_v2` shall return. **]**
+
+**SRS_THREADPOOL_WIN32_05_002: [** Otherwise `context` shall be used as the context created in `threadpool_create_work_item`. **]**
+
+**SRS_THREADPOOL_WIN32_05_003: [** The `work_function` callback passed to `threadpool_create_work_item` shall be called with the `work_function_context` as an argument. `work_function_context` was set inside the `threadpool_create_work_item` as an argument to `CreateThreadpoolContext`. **]**
```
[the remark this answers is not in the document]
nishudeshu: Hi Matt, do you mean, the entire specs for `on_work_callback_v2` should be after `threadpool_schedule_work_item` specs?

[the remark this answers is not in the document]
mattdurak: yeah, not a big deal though, just a suggestion

interfaces/devdoc/threadpool_requirements.md
nishudeshu: I think adding `_item` is creating all the confusion where I feel to avoid `threadpool_schedule_work` being declared twice, we are naming `threadpool_schedule_work_item` as the new API. 
So, should we have following APIs:
`threadpool_create_work`   <-- Create everything needed make scheduling work
`threadpool_schedule_work`
`threadpool_destroy_work` <--- Destroy everything that was initialized to make scheduling work
`threadpool_schedule_work_item`

For now, can we create `threadpool_schedule_work_item` as an additional new API and later when we remove `threadpool_schedule_work`, we just rename `threadpool_schedule_work_item` to `threadpool_schedule_work`.

I feel we should introduce versioning of the datastructures starting with `WORK_ITEM_CONTEXT` having a version field where version 1 is the earlier implementation and version 2 is the latest and initialized to 2 by Queue Processor

mattdurak: I think either name is fine, but we should not rename `threadpool_schedule_work` right now so we don't have to update the code everywhere in all the repos.

Personally I don't see a problem with
`threadpool_create_work_item`
`threadpool_schedule_work_item`
`threadpool_destroy_work_item`

Because we are creating a "work item" which can be scheduled.

But sure, we could do a future rename to remove the "item" part.

I'm not sure we gain much with versioning the internal context structure here though. It sounds like it would add extra complexity and shouldn't be needed (it's not something we are persisting or transferring over the network where versioning would be helpful for having a mix of old/new code).

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,54 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+```c
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_ITEM_HANDLE, work_item_context);
+```
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool. This API is called multiple times on the same `THREADPOOL_WORK_ITEM_HANDLE` and fails only if the arguments are passed `NULL`.
+
+**NON_THREADPOOL_05_007: [** If `threadpool` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
+
+**NON_THREADPOOL_05_008: [** If `work_function` is NULL, `threadpool_schedule_work_item` shall fail and return a `non-zero` value. **]**
```
mattdurak, line 143: >work_function [](http://example.com/codeflow?start=33&length=13)

This is called `work_item_context` in the function declaration

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -31,6 +31,12 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
 MOCKABLE_FUNCTION(, int, threadpool_open, THANDLE(THREADPOOL), threadpool);
 MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 
+MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, void*, work_function_context);
+
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_ITEM_HANDLE, work_item_context);
```
mattdurak, line 36: >work_item_context [](http://example.com/codeflow?start=118&length=17)

poetry: Can remove the word "_context" here and in the next function. We typically use "context" along with arbitrary data that is passed along with callbacks (like `work_function_context` above), but not in a case like this (see the other functions here which take a `timer` or a `threadpool` and not a `timer_context`/`threadpool_context`.

win32/devdoc/threadpool_win32.md
```diff
@@ -23,6 +23,12 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
 MOCKABLE_FUNCTION(, int, threadpool_open, THANDLE(THREADPOOL), threadpool);
 MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 
+MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, PVOID, work_function_context);
```
mattdurak, line 26: >PVOID [](http://example.com/codeflow?start=152&length=5)

missing updates from the header I suspect, as this still says "PVOID"

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -31,6 +31,12 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
...
+
+MOCKABLE_FUNCTION(, void, threadpool_work_context_destroy, THREADPOOL_WORK_ITEM_HANDLE, work_item_context);
```
mattdurak, line 38: >threadpool_work_context_destroy [](http://example.com/codeflow?start=26&length=31)

name doesn't match the header (sorry, c-pal is more annoying than other projects because we copy the header into 3 places (interface, windows, and linux requirements)

win32/devdoc/threadpool_win32.md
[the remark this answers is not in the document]
mattdurak: I mean it should probably be around line 301 (on code flow iteration 7), not way up with `on_work_callback`. We should keep it very close to the "threadpool_schedule_work_item" function that causes it to execute.

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +264,51 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+
+**S_R_S_THREADPOOL_WIN32_05_005: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_006: [** Otherwise `threadpool_create_work_item` shall allocate a context where `work_function`, `work_function_context`, and `ptp_work` shall be saved. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_007: [** If any error occurs, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_008: [** `threadpool_create_work_item` shall create `PTP_WORK` by calling `CreateThreadpoolWork` to set the callback function as `on_work_callback_v2`. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_009: [** If any error occurs, `threadpool_create_work_item` shall fail, free the newly created context and return a `NULL` value. **]**
```
mattdurak, line 286: >9: [** If any error occurs, `th [](http://example.com/codeflow?start=30&length=31)

Missing spec about returning successfully ("...shall succeed and return ..."

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -381,3 +387,61 @@ static int threadpool_work_func(void* param);
...
+**S_R_S_THREADPOOL_LINUX_05_014: [** `work_item_context` can be a `NULL` value **]**
+
+**S_R_S_THREADPOOL_LINUX_05_015: [** If `threadpool` is `NULL`, `threadpool_schedule_work_item` shall fail and return a non-zero value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_016: [** `threadpool_schedule_work_item` shall call `sm_exec_begin`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_017: [** If `sm_exec_begin` returns `SM_EXEC_REFUSED`, `threadpool_schedule_work_item` shall fail and return a non-zero value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_018: [** `threadpool_schedule_work_item` shall unblock the `threadpool` semaphore by calling `sem_post`. **]**
```
mattdurak, line 437: >sem_post [](http://example.com/codeflow?start=122&length=8)

still need to figure out how we will be able to call this multiple times and have the work execute each time.

interfaces/devdoc/threadpool_requirements.md
nishudeshu: It's work_item_context in the original threadpool_schedule_work that we are now passing around. Earlier we used to create it and free inside the callback function. But now we are saving it by returning it to the caller from the create function

linux/devdoc/threadpool_linux_requirements.md
nishudeshu: Linux implementation would be safe because all we are doing is creating additional work items and calling schedule. So each call to schedule should have done items to work on

interfaces/devdoc/threadpool_requirements.md
nishudeshu: Good catch

linux/devdoc/threadpool_linux_requirements.md
jebrando: We can have a chat about this offline so I understand the issue.

interfaces/devdoc/threadpool_requirements.md
dcristoloveanu: Matt means that conceptually what gets passed here from the point of view of the user of this API is a work item handle. We generally use context for a pointer associated with something.

Looking from the point of view of a user of the APIs. For example for threadpool_schedule_work, user passes a work function and an opaque context associated with that work_function.

In the case of this API (threadpool_schedule_work_item), the user passes a work item handle, not a context associated with a work item handle.

So it should be named as such: work_itm or work_item_handle.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,56 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_001: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_002: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_003: [** Otherwise `threadpool_create_work_item` shall allocate a context of type `THREADPOOL_WORK_ITEM_HANDLE` to save threadpool related variables. **]**
+
+**NON_THREADPOOL_05_004: [** If any error occurs, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**NON_THREADPOOL_05_005: [** `threadpool_create_work_item` shall create an individual work item context of type `THREADPOOL_WORK_ITEM_HANDLE` and if needed set a local callback function. **]**
```
dcristoloveanu, line 129: > and if needed set a local callback function [](http://example.com/codeflow?start=141&length=44)

I would not know how to test this. Generally our specs should be crisp enough so that we can write tests based on them.

dcristoloveanu: I'd remove the part of "and if needed set a local callback function"

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,56 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_006: [** If there are no errors then this work item context of type `THREADPOOL_WORK_ITEM_HANDLE` would be returned indicating a success to the caller. **]**
```
dcristoloveanu, line 131: >this [](http://example.com/codeflow?start=57&length=4)

I'd say "just created work item". /consider

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -381,3 +387,67 @@ static int threadpool_work_func(void* param);
...
+
+**S_R_S THREADPOOL_LINUX_05_001: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_002: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall create `THREADPOOL_WORK_ITEM_HANDLE` variable `threadpool_work_item` by calling `malloc`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_004: [** If the `malloc` fails and `threadpool_work_item` is `NULL` then log the error and return `NULL`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_005: [** `threadpool_create_work_item` shall copy the `work_function` and `work_function_context` into the `threadpool_work_item` amd return `threadpool_work_item` to indicate success. **]**
```
dcristoloveanu, line 407: >amd [](http://example.com/codeflow?start=158&length=3)

Intel? :-)

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -381,3 +387,67 @@ static int threadpool_work_func(void* param);
...
+MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, void*, work_function_context);
+```
+
+`threadpool_create_work_item` creates a work item to be executed by the threadpool.
+
+**S_R_S THREADPOOL_LINUX_05_001: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_002: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall create `THREADPOOL_WORK_ITEM_HANDLE` variable `threadpool_work_item` by calling `malloc`. **]**
```
dcristoloveanu, line 403: >`THREADPOOL_WORK_ITEM_HANDLE` variable `threadpool_work_item` by calling `malloc` [](http://example.com/codeflow?start=80&length=81)

I'd consider saying "shall allocate memory for a THREADPOOL_WORK_ITEM_HANDLE"

dcristoloveanu:
> 

I don't think that is true, destroy should free the memory allocated in threadpool_create_work_item

---
Refers to: linux/devdoc/threadpool_linux_requirements.md:454 in f069b59. [](commit_id = f069b5953bdb6c833aa893a15d6daa8231b4e9bb, deletion_comment = False)

dcristoloveanu (changes requested):
:clock1:

linux/src/threadpool_linux.c
```diff
@@ -507,6 +515,23 @@ void threadpool_close(THANDLE(THREADPOOL) threadpool)
...
+
+int threadpool_schedule_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WORK_ITEM_HANDLE work_item_context)
+{
+    LogWarning("Not Implemented: threadpool_schedule_work_item(%p)", work_item_context);
+    return 0;
+}
+
+void threadpool_work_context_destroy(THREADPOOL_WORK_ITEM_HANDLE work_item_context)
+{
+    LogWarning("Not Implemented: threadpool_work_context_destroy(%p)", work_item_context);
```
dcristoloveanu, line 532: Cool

win32/devdoc/threadpool_win32.md
```diff
@@ -243,3 +264,54 @@ static VOID CALLBACK on_timer_callback(PTP_CALLBACK_INSTANCE instance, PVOID con
...
+MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_FUNCTION, work_function, void*, work_function_context);
+```
+
+`threadpool_create_work_item` creates a work item to be executed by the threadpool.
+
+**S_R_S_THREADPOOL_WIN32_05_004: [** If `threadpool` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_005: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
+
+**S_R_S_THREADPOOL_WIN32_05_006: [** Otherwise `threadpool_create_work_item` shall allocate a context `work_item_context` of type `THREADPOOL_WORK_ITEM_HANDLE` where `work_function`, `work_function_context`, and `ptp_work` shall be saved. **]**
```
dcristoloveanu, line 280: >shall allocate a context `work_item_context` [](http://example.com/codeflow?start=77&length=44)

Can we avoid this allocation alltogether?

nishudeshu: It happens once and the function return type is THREADPOOL_WORK_ITEM_HANDLE. Or we can ask the caller to allocate it and it can be statically defined on their end? We can also have a double pointer and allocate only if it's content is a NULL pointer.

linux/devdoc/threadpool_linux_requirements.md
nishudeshu: NVDA :)

linux/src/threadpool_linux.c
nishudeshu: Sorry, did not understand.

win32/devdoc/threadpool_win32.md
dcristoloveanu: I was hinting at whether we can cast PTP_WORK to our handle. Since both are pointers and I don't think we need to store anything else in our "work item context"

dcristoloveanu:
> 

If threadpool_create_work_item allocates memory, destroy has to free it.

"**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall create `THREADPOOL_WORK_ITEM_HANDLE` variable `threadpool_work_item` by calling `malloc`. **]**
"

---
In reply to: [2322623928](https://github.com/Azure/c-pal/pull/362#issuecomment-2322623928) [](http://example.com/codeflow?ancestors=2322623928)

---
Refers to: linux/devdoc/threadpool_linux_requirements.md:454 in f069b59. [](commit_id = f069b5953bdb6c833aa893a15d6daa8231b4e9bb, deletion_comment = False)

win32/devdoc/threadpool_win32.md
nishudeshu: In WIN32, this context gets passed to `on_work_callback` where it calls `context->work_function(work_function_context)`.

interfaces/devdoc/threadpool_requirements.md
```diff
@@ -104,6 +110,56 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THREADPOOL_HANDLE, threadpool);
...
+
+**NON_THREADPOOL_05_007: [** If any error occurs, `threadpool_create_work_item` shall fail, free the newly created context and return a `NULL` value. **]**
+
+### threadpool_schedule_work_item
+
+```c
+MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), threadpool, THREADPOOL_WORK_ITEM_HANDLE, work_item_context);
+```
+
+`threadpool_schedule_work_item` schedules a work item to be executed by the threadpool. This API is called multiple times on the same `THREADPOOL_WORK_ITEM_HANDLE` and fails only if the arguments are passed `NULL`.
```
dcristoloveanu, line 141: >is cal [](http://example.com/codeflow?start=97&length=6)

"can be called".

win32/devdoc/threadpool_win32.md
dcristoloveanu: Ok.

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -381,3 +387,67 @@ static int threadpool_work_func(void* param);
...
+
+### threadpool_destroy_work_item
+
+```c
+MOCKABLE_FUNCTION(, void, threadpool_destroy_work_item, THREADPOOL_WORK_ITEM_HANDLE, work_item_context);
+```
+
+`threadpool_work_context_destroy` Does nothing and a placeholder for WIN32 equivalent function stub
+
+**S_R_S_THREADPOOL_LINUX_05_020: [** Free memory allocated to the work item of type `THREADPOOL_WORK_ITEM_HANDLE` created in `threadpool_create_work_item`. **]**
```
dcristoloveanu, line 453: >Free memory allocated to  [](http://example.com/codeflow?start=37&length=25)

Nitpick: `threadpool_destroy_work_item` shall free the memory ...

dcristoloveanu (approved):
:shipit:

PR #369, opened 2024-09-16
Added method to get time elapsed in seconds without creating a timer

flash-sinx:
## Summary

This pull request introduces a new method `timer_global_get_elapsed` to get the time elapsed in seconds without the need to create a timer. This method is useful for avoid using malloc for creating timer.

## Changes

- Added `timer_global_get_elapsed` method to return the elapsed time in seconds from a start time in the past.
- Added a reset function `global_timer_state_reset` to reset the global timer state between tests.

## Testing

- Verified the functionality with unit tests to ensure the correct elapsed time is returned.
- Added tests to check the reset functionality to avoid state interference between tests.

### DIFF
--- a/interfaces/inc/c_pal/timer.h
+++ b/interfaces/inc/c_pal/timer.h
@@ -18,6 +18,7 @@ MOCKABLE_FUNCTION(, int, timer_start, TIMER_HANDLE, handle);
 MOCKABLE_FUNCTION(, double, timer_get_elapsed, TIMER_HANDLE, timer);
 MOCKABLE_FUNCTION(, double, timer_get_elapsed_ms, TIMER_HANDLE, timer);
 MOCKABLE_FUNCTION(, void, timer_destroy, TIMER_HANDLE, timer);
+MOCKABLE_FUNCTION(, void, global_timer_state_reset);
 MOCKABLE_FUNCTION(, double, timer_global_get_elapsed);
 MOCKABLE_FUNCTION(, double, timer_global_get_elapsed_ms);
 MOCKABLE_FUNCTION(, double, timer_global_get_elapsed_us);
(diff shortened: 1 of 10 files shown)


interfaces/inc/c_pal/timer.h
```diff
@@ -18,6 +18,8 @@ MOCKABLE_FUNCTION(, int, timer_start, TIMER_HANDLE, handle);
 MOCKABLE_FUNCTION(, double, timer_get_elapsed, TIMER_HANDLE, timer);
 MOCKABLE_FUNCTION(, double, timer_get_elapsed_ms, TIMER_HANDLE, timer);
 MOCKABLE_FUNCTION(, void, timer_destroy, TIMER_HANDLE, timer);
+MOCKABLE_FUNCTION(, void, global_timer_state_reset);
```
avranju, line 21: Does this exist only for enabling unit testing?

avranju (approved):
As discussed, we should probably add the Linux implementation as well.

linux/src/timer_linux.c
```diff
@@ -180,6 +180,28 @@ double timer_get_elapsed_ms(TIMER_HANDLE timer)
     return result;
 }
 
+double timer_global_get_elapsed_s(void)
+{
+    double result;
+    struct timespec elapsed_time;
+
+    /* Codes_SRS_TIMER_27_001: [ timer_global_get_elapsed_s shall return the elapsed time in seconds from a start time in the past. ]*/
```
avranju, line 188: Should this be `SRS_TIMER_LINUX_27_001`? Also, `SRS_TIMER_LINUX_27_003` seems to be missing.

avranju (dismissed):

PR #370, opened 2024-09-18
design & header specs for job_object_helper

BertKleewein:


win32/devdoc/job_object_helper.md
```diff
@@ -0,0 +1,46 @@
...
+## job_object_helper_limit_resources
+```c
+    MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu);
+```
+
+`job_object_helper_limit_resources` is an API that uses the job object API to limit the amount of physical memory and CPU that the current process can consume.
+
+**SRS_JOB_OBJECT_HELPER_18_001: [** `percent_physical_memory` is allowed to be `0`, which means "do not limit physical memory usage". **]**
+
+**SRS_JOB_OBJECT_HELPER_18_002: [** `percent_cpu` is allowed to be `0`, which means "do not limit CPU usage". **]**
```
parth21999, line 19: We usually don't spec which values are valid. 

How would you test these two specs exclusively without hitting _003?

win32/devdoc/job_object_helper.md
```diff
@@ -0,0 +1,46 @@
...
+
+**SRS_JOB_OBJECT_HELPER_18_005: [** If `percent_cpu` is greater than `100`, `job_object_helper_limit_resources` shall fail and return a non-zero value. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_006: [** `job_object_helper_limit_resources` shall call `CreateJobObject` to create a new job object passing `NULL` for both `lpJobAttributes` and `lpName`. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_007: [** `job_object_helper_limit_resources` shall call `GetCurrentProcess` to get the current process handle. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_008: [** `job_object_helper_limit_resources` shall call `AssignProcessToJobObject` to assign the current process to the new job object. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_009: [** If `percent_physical_memory` is not `0`, `job_object_helper_limit_resources` shall call `GlobalMemoryStatusEx` to get the total amount of physical memory in kb. **]**
```
parth21999, line 33: > If `percent_physical_memory` is not `0` [](http://example.com/codeflow?start=35&length=40)

003 says that the function will return. 

This check seems redundant

parth21999: same for 010

parth21999: and 011

win32/src/job_object_helper.c
```diff
@@ -0,0 +1,17 @@
...
+
+#include "stdint.h"
+
+#include "macro_utils/macro_utils.h"
+#include "c_logging/logger.h"
+
+#include "c_pal/job_object_helper.h"
+
+
+IMPLEMENT_MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_pysical_memory, uint32_t, percent_cpu)
```
parth21999, line 12: >pysical [](http://example.com/codeflow?start=88&length=7)

nti

win32/devdoc/job_object_helper.md
BertKleewein: removed.  003 only comes into play if they are both zero.

win32/devdoc/job_object_helper.md
BertKleewein: 003 comes into play only if they are both `0`.

BertKleewein: rewritten to remove ambiguity.

win32/devdoc/job_object_helper.md
BertKleewein: rewritten to simplify and remove redundancy.

parth21999 (approved):
:shipit:

win32/inc/c_pal/job_object_helper.h
```diff
@@ -0,0 +1,29 @@
...
+#include "macro_utils/macro_utils.h"
+
+#include "umock_c/umock_c_prod.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+    MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu);
```
anporumb, line 23: >job_object_helper_limit_resources [](http://example.com/codeflow?start=29&length=33)

I am going to make a comment (or more) here, which nobody will like, buuuut... 

1) I am hoping that in the big design, these limits are set by the Almighty Resource Governor - and that is SF in the end. Somehow we living in separate process/processes or clumped together should matter differently.

2) Assume we're somehow at our wits end - and everything is "used". Not 1 byte of memory left. The technician comes and wants to ... change the percent physical memory. It is conceivable that this function will fail because it will fail creating the JobObjects. So maybe consider having a handle that keep in memory the JobObject to be used/reused across calls. That is, once the handle is created, adjustment is always possible.

BertKleewein: Good points. Right now, this is a one-time thing that gets called at service start based on SF config parameters. Something that allows for outside control is a much bigger feature and out of scope.

PR #371, opened 2024-09-19
Threadpool WIN32 changes implementation

nishudeshu:
Threadpool changes on WIN32 for splitting the threadpool_schedule_work into threadpool_create_work_item, threadpool_schedule_work_item, and threadpool_destroy_work_item.

### DIFF
--- a/win32/src/threadpool_win32.c
+++ b/win32/src/threadpool_win32.c
@@ -33,7 +33,7 @@ typedef struct THREADPOOL_WORK_ITEM_TAG
 {
     THREADPOOL_WORK_FUNCTION work_function;
     void* work_function_context;
-    PTP_WORK* ptp_work;
+    PTP_WORK ptp_work;
 } THREADPOOL_WORK_ITEM, *THREADPOOL_WORK_ITEM_HANDLE;
 
 typedef struct TIMER_INSTANCE_TAG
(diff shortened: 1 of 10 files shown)


win32/src/threadpool_win32.c
```diff
@@ -87,7 +87,19 @@ static VOID CALLBACK on_work_callback(PTP_CALLBACK_INSTANCE instance, PVOID cont
 
 static VOID CALLBACK on_work_callback_v2(PTP_CALLBACK_INSTANCE instance, PVOID context, PTP_WORK work)
 {
-    LogWarning("Printing %p %p %p", instance, context, work);
+    if (context == NULL)
+    {
+        /* Codes_SRS_THREADPOOL_WIN32_05_001: [ If context is NULL, on_work_callback_v2 shall return. ]*/
+        LogError("Invalid arguments: PTP_CALLBACK_INSTANCE instance=%p, PVOID context=%p, PTP_WORK work=%p",
```
jebrando, line 93: >LogError [](http://example.com/codeflow?start=8&length=8)

This should be a fatal error, it should be LogCriticalAndTerminate

win32/src/threadpool_win32.c
```diff
@@ -277,24 +289,133 @@ void threadpool_close(THANDLE(THREADPOOL) threadpool)
...
+            threadpool, work_function, work_function_context);
+    }
+    else
+    {
+        THREADPOOL* threadpool_ptr = THANDLE_GET_T(THREADPOOL)(threadpool);
+
+        (void)InterlockedIncrement(&threadpool_ptr->pending_api_calls);
+
+        THREADPOOL_WIN32_STATE state = InterlockedAdd(&threadpool_ptr->state, 0);
+        if (state != (LONG)THREADPOOL_WIN32_STATE_OPEN)
```
jebrando, line 315: >(LONG) [](http://example.com/codeflow?start=21&length=6)

I don't think you need a cast here.

jebrando (dismissed):
:shipit:

win32/src/threadpool_win32.c
nishudeshu: Hi Jelani, it's a callback function, do we want to exit Fatal for callbacks? Also, reference `on_work_callback` where we are just doing `LogError`

win32/src/threadpool_win32.c
nishudeshu: Hi Jelani, just referenced from `threadpool_schedule_work`. Should I make these changes for all such occurrences in the file?

win32/src/threadpool_win32.c
jebrando: We would terminate where because they context should never be NULL and if it is then something really bad has happened and we can't recover from this.

win32/src/threadpool_win32.c
jebrando: Sure, I think all these types of casts are unnecessary.

jebrando (approved):
:shipit:

nishudeshu:
https://msazure.visualstudio.com/One/_workitems/edit/29311073 
Story board

PR #373, opened 2024-09-25
Spec changes to handle freeing the context after all tasks are comple…

nishudeshu:
Spec changes to handle freeing the context after all tasks are completed and corresponding minor changes in threadpool_schedule_work

### DIFF
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -414,7 +414,7 @@ MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, TH
 
 **S_R_S_THREADPOOL_LINUX_05_005: [** `threadpool_create_work_item` shall allocate memory for `threadpool_work_item` of type `THREADPOOL_WORK_ITEM_HANDLE`. **]**
 
-**S_R_S_THREADPOOL_LINUX_05_006: [** If the `malloc` fails and `threadpool_work_item` is `NULL` then log the error and return `NULL`. **]**
+**S_R_S_THREADPOOL_LINUX_05_006: [** If during the initialization of `threadpool_work_item`, `malloc` fails then log the error and threadpool_create_work_item shall return `NULL`. **]**
 
 **S_R_S_THREADPOOL_LINUX_05_007: [** `threadpool_create_work_item` shall initialize `pending_work_item_count` to `0` then copy the `work_function` and `work_function_context` into the `threadpool_work_item` and return `threadpool_work_item` to indicate success. **]**
 
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -237,7 +237,7 @@ MOCKABLE_FUNCTION(, int, threadpool_schedule_work, THANDLE(THREADPOOL), threadpo
 
 **SRS_THREADPOOL_LINUX_07_048: [** If reallocating the task array fails, `threadpool_schedule_work` shall fail and return a non-zero value. **]**
 
-**SRS_THREADPOOL_LINUX_07_049: [** `threadpool_schedule_work` shall initialize `pending_work_item_count_ptr` with `NULL` then copy the work function and work function context into insert position in the task array and return zero on success. **]**
+**SRS_THREADPOOL_LINUX_07_049: [** `threadpool_schedule_work` shall initialize `pending_work_item_count_ptr` with `NULL` then copy the work function and work function context into insert position in the task array and assign `0` to the return variable to indicate success. **]**
 
 **SRS_THREADPOOL_LINUX_07_050: [** `threadpool_schedule_work` shall set the `task_state` to `TASK_WAITING` and then release the shared SRW lock. **]**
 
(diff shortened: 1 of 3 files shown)


linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -400,11 +408,17 @@ MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, TH
...
 **S_R_S_THREADPOOL_LINUX_05_002: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
 
-**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall allocate memory for `threadpool_work_item` of type `THREADPOOL_WORK_ITEM_HANDLE`. **]**
+**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall call `sm_exec_begin`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_004: [** If `sm_exec_begin` returns `SM_EXEC_REFUSED`, `threadpool_create_work_item` shall fail and return a non-zero value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_005: [** `threadpool_create_work_item` shall allocate memory for `threadpool_work_item` of type `THREADPOOL_WORK_ITEM_HANDLE`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_006: [** If the `malloc` fails and `threadpool_work_item` is `NULL` then log the error and return `NULL`. **]**
```
jebrando, line 417: >`malloc` fails and `threadpool_work_item` is `NULL` [](http://example.com/codeflow?start=44&length=51)

This should just say 

If the malloc fails, threadpool_create_work_item shall return NULL.

because if malloc fails then there is no way for threadpool_work_item to not be NULL

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -237,7 +237,7 @@ MOCKABLE_FUNCTION(, int, threadpool_schedule_work, THANDLE(THREADPOOL), threadpo
 
 **SRS_THREADPOOL_LINUX_07_048: [** If reallocating the task array fails, `threadpool_schedule_work` shall fail and return a non-zero value. **]**
 
-**SRS_THREADPOOL_LINUX_07_049: [** `threadpool_schedule_work` shall copy the work function and work function context into insert position in the task array and return zero on success. **]**
+**SRS_THREADPOOL_LINUX_07_049: [** `threadpool_schedule_work` shall initialize `pending_work_item_count_ptr` with `NULL` then copy the work function and work function context into insert position in the task array and return zero on success. **]**
```
jebrando, line 240: >and return zero on success [](http://example.com/codeflow?start=214&length=26)

This part is not needed since you have essentially the same spec on line 248

jebrando (dismissed):
:shipit:

linux/devdoc/threadpool_linux_requirements.md
nishudeshu: Okay, wanted to specify what variable we are malloc'ing, putting the following:

If during the initialization of `threadpool_work_item`, `malloc` fails then log the error and return `NULL`.

linux/devdoc/threadpool_linux_requirements.md
nishudeshu: I have seen this all over the place. Looks like line 240 wants to specify a jump to line 248 or some kind of a branch around 240.

parth21999: I agree with Jelani. Putting a "return zero on success" clause in this requirement makes it sound like we are return right away without touching the subsequent specs. Seems like we go on to release a lock so that is probably not what we want here.

linux/devdoc/threadpool_linux_requirements.md
parth21999: The "If during the initialization of threadpool_work_item" clause seems confusing. I agree with Jelani, it should simply say "If the malloc fails, threadpool_create_work_item shall return NULL."

parth21999 (changes requested):
:clock1:

parth21999 (dismissed):
:shipit:

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -400,11 +408,17 @@ MOCKABLE_FUNCTION(, THREADPOOL_WORK_ITEM_HANDLE, threadpool_create_work_item, TH
...
 **S_R_S_THREADPOOL_LINUX_05_002: [** If `work_function` is `NULL`, `threadpool_create_work_item` shall fail and return a `NULL` value. **]**
 
-**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall allocate memory for `threadpool_work_item` of type `THREADPOOL_WORK_ITEM_HANDLE`. **]**
+**S_R_S_THREADPOOL_LINUX_05_003: [** `threadpool_create_work_item` shall call `sm_exec_begin`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_004: [** If `sm_exec_begin` returns `SM_EXEC_REFUSED`, `threadpool_create_work_item` shall fail and return a non-zero value. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_005: [** `threadpool_create_work_item` shall allocate memory for `threadpool_work_item` of type `THREADPOOL_WORK_ITEM_HANDLE`. **]**
+
+**S_R_S_THREADPOOL_LINUX_05_006: [** If during the initialization of `threadpool_work_item`, `malloc` fails then log the error and threadpool_create_work_item shall return `NULL`. **]**
```
parth21999, line 417: 05_006 and 05_004 can be combined into one requirement that says 

"If there are any failures, `threadpool_create_work_item` shall fail and return a non-zero value"

This is how a lot of other modules do it (although not all) and seems to be the preferred way to spec the failure case.

linux/devdoc/threadpool_linux_requirements.md
parth21999: I would say that the "and assign `0` to the return variable to indicate success" clause is not really required. There is no a good way you can write a test for this, and we don't really do this in other places.

parth21999 (approved):
:shipit:

PR #376, opened 2024-09-26
job_object_helper code & tests

BertKleewein:


win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
 
 IMPLEMENT_MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu)
 {
-    (void) percent_physical_memory;
-    (void) percent_cpu;
-    return MU_FAILURE;
+    int result;
+    LogInfo("job_object_helper_limit_resources called with parameters:"
+        " uint32_t percent_physical_memory = %d"
+        ", uint32_t percent_cpu = %d",
```
dcristoloveanu, line 17: >d [](http://example.com/codeflow?start=35&length=1)

PRIu32

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
 
 IMPLEMENT_MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu)
 {
-    (void) percent_physical_memory;
-    (void) percent_cpu;
-    return MU_FAILURE;
+    int result;
+    LogInfo("job_object_helper_limit_resources called with parameters:"
+        " uint32_t percent_physical_memory = %d"
```
dcristoloveanu, line 16: >d [](http://example.com/codeflow?start=46&length=1)

PRIu32

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+    /*Codes_SRS_JOB_OBJECT_HELPER_18_005: [ If percent_cpu is greater than 100, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+    if ((percent_physical_memory == 0) ||
+        (percent_physical_memory > 100) || 
+        (percent_cpu == 0) ||
+        (percent_cpu > 100))
+    {
+        LogError("Invalid arguments:"
+            " uint32_t percent_physical_memory = %d"
+            ", uint32_t percent_cpu = %d",
+            percent_physical_memory,
```
dcristoloveanu, line 33: Here too, PRIu32

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+        {
+            /*Codes_SRS_JOB_OBJECT_HELPER_18_007: [ job_object_helper_limit_resources shall call GetCurrentProcess to get the current process handle. ]*/
+            HANDLE current_process = GetCurrentProcess();
+            if (current_process == NULL)
+            {
+                LogLastError("GetCurrentProcess() failed");
+                /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+                result = MU_FAILURE;
+            }
+            else
```
dcristoloveanu, line 58: A nitpick that I think this cannot really fail:

[https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocess)

Return value
The return value is a pseudo handle to the current process.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+                        LogLastError("GlobalMemoryStatus(&memory_status_ex = %p) failed", &memory_status_ex);
+                        /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+                        result = MU_FAILURE;
+                    }
+                    else
+                    {
+                        /*Codes_SRS_JOB_OBJECT_HELPER_18_010: [ job_object_helper_limit_resources shall call SetInformationJobObject, passing JobObjectExtendedLimitInformation and a JOBOBJECT_EXTENDED_LIMIT_INFORMATION object with JOB_OBJECT_LIMIT_JOB_MEMORY set and JobMemoryLimit set to the percent_physical_memory percent of the physical memory in bytes. ]*/
+                        JOBOBJECT_EXTENDED_LIMIT_INFORMATION extended_limit_information = {0};
+                        extended_limit_information.BasicLimitInformation.LimitFlags  = JOB_OBJECT_LIMIT_JOB_MEMORY;
+                        extended_limit_information.JobMemoryLimit = percent_physical_memory * memory_status_ex.ullTotalPhys / 100;
```
dcristoloveanu, line 83: >.JobMemoryLimit = percent_physical_memory [](http://example.com/codeflow?start=50&length=41)

Nice, so, we are looking at physical memory.
This is another nail in the coffin to not have paging enabled. I was pushing for that today.

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+#define test_percent_physical_memory 89
+#define test_job_object (HANDLE)0x42001
+#define test_process (HANDLE)0x42002
+#define test_total_physical_memory 20480
+
+MOCK_FUNCTION_WITH_CODE(, BOOL, mocked_GlobalMemoryStatusEx, LPMEMORYSTATUSEX, lpBuffer)
+{
+    ASSERT_IS_NOT_NULL(lpBuffer);
+    ASSERT_ARE_EQUAL(size_t, lpBuffer->dwLength, sizeof(MEMORYSTATUSEX));
+    memset(lpBuffer, 0, sizeof(MEMORYSTATUSEX));
```
dcristoloveanu, line 45: >memset [](http://example.com/codeflow?start=4&length=6)

(void)

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
+// Copyright(C) Microsoft Corporation.All rights reserved.
+
+
+#include <stdlib.h>
+
+#include "macro_utils/macro_utils.h"
+
+#include "testrunnerswitcher.h"
+#include "windows.h"
```
dcristoloveanu, line 9: We usually put windows.h before other of ours includes.
Typical order:
- C standard includes
- OS includes 
- our dependencies includes (usually lowest to highest)
- our current code includes.

Makes it easier to look for them (not that it won't work)

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+MOCK_FUNCTION_END(TRUE)
+
+static JOBOBJECT_EXTENDED_LIMIT_INFORMATION captured_extended_limit_information;
+static JOBOBJECT_CPU_RATE_CONTROL_INFORMATION captured_cpu_rate_control_information;
+MOCK_FUNCTION_WITH_CODE(, BOOL, mocked_SetInformationJobObject, HANDLE, hJob, JOBOBJECTINFOCLASS, JobObjectInformationClass, LPVOID, lpJobObjectInformation, DWORD, cbJobObjectInformationLength)
+{
+    ASSERT_IS_NOT_NULL(lpJobObjectInformation);
+    if (JobObjectInformationClass == JobObjectExtendedLimitInformation)
+    {
+        memcpy((void*)&captured_extended_limit_information, lpJobObjectInformation, cbJobObjectInformationLength);
```
dcristoloveanu, line 58: >memcpy [](http://example.com/codeflow?start=8&length=6)

(void)

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+    }
+    else if (JobObjectInformationClass == JobObjectCpuRateControlInformation)
+    {
+        memcpy((void*)&captured_cpu_rate_control_information, lpJobObjectInformation, cbJobObjectInformationLength);
```
dcristoloveanu, line 62: >memcpy [](http://example.com/codeflow?start=8&length=6)

(void)

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+    }
+    else
+    {
+        ASSERT_FAIL();
```
dcristoloveanu, line 66: >(); [](http://example.com/codeflow?start=19&length=3)

Can we have a hint of a  message?

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+    int result;
+
+    ASSERT_ARE_EQUAL(int, 0, real_gballoc_hl_init(NULL, NULL));
+
+    result = umock_c_init(on_umock_c_error);
+    ASSERT_ARE_EQUAL(int, 0, result, "umock_c_init failed");
+
+    result = umocktypes_windows_register_types();
+    ASSERT_ARE_EQUAL(int, 0, result, "umocktypes_windows_register_types failed");
+
```
dcristoloveanu, line 84: Consider having these as one-liners

win32/tests/job_object_helper_ut/job_object_helper_ut.c
```diff
@@ -0,0 +1,238 @@
...
+/*Tests_SRS_JOB_OBJECT_HELPER_18_010: [ job_object_helper_limit_resources shall call SetInformationJobObject, passing JobObjectExtendedLimitInformation and a JOBOBJECT_EXTENDED_LIMIT_INFORMATION object with JOB_OBJECT_LIMIT_JOB_MEMORY set and JobMemoryLimit set to the percent_physical_memory percent of the physical memory in bytes. ]*/
+/*Tests_SRS_JOB_OBJECT_HELPER_18_011: [ job_object_helper_limit_resources shall call SetInformationJobObject passing JobObjectCpuRateControlInformation and a JOBOBJECT_CPU_RATE_CONTROL_INFORMATION object with JOB_OBJECT_CPU_RATE_CONTROL_ENABLE and JOB_OBJECT_CPU_RATE_CONTROL_HARD_CAP set, and CpuRate set to percent_cpu times 100. ]*/
+/*Tests_SRS_JOB_OBJECT_HELPER_18_012: [ job_object_helper_limit_resources shall call CloseHandle to close the handle of the current process. ]*/
+/*Tests_SRS_JOB_OBJECT_HELPER_18_013: [ job_object_helper_limit_resources shall call CloseHandle to close the handle to the job object. ]*/
+/*Tests_SRS_JOB_OBJECT_HELPER_18_014: [ job_object_helper_limit_resources shall succeed and return 0. ]*/
+TEST_FUNCTION(job_object_helper_limit_resources_succeeds)
```
dcristoloveanu, line 195: I think it would be nice to add an extra test with some different values

win32/tests/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ if(${run_unittests})
     endif()
 
     build_test_folder(gballoc_hl_passthrough_ut)
+    build_test_folder(job_object_helper_ut)
```
dcristoloveanu, line 36: >job_object_helper_ut [](http://example.com/codeflow?start=22&length=20)

I think we should add a task for an int test for this.
The int test could just try to allocate more than what the limit is for memory - to check that the job object is setup properly.

dcristoloveanu: We can do the same for CPU, we could spin n threads in a tight loop and check that CPU does not go over.
I highly suggest delegating this task to Parth so we do not put you with it on the critical path.

dcristoloveanu (dismissed):
:shipit:

anporumb:
> #include "stdint.h"

use <> for this include

---
Refers to: win32/src/job_object_helper.c:4 in 36b59d0. [](commit_id = 36b59d03d6987d81acdc5a376170e22ff5b5e14a, deletion_comment = False)

anporumb:
> #include "macro_utils/macro_utils.h"

probably missing "windows.h"

Not sure about jobapi2.h

---
Refers to: win32/src/job_object_helper.c:6 in 36b59d0. [](commit_id = 36b59d03d6987d81acdc5a376170e22ff5b5e14a, deletion_comment = False)

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
 
 IMPLEMENT_MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu)
```
anporumb, line 12: >int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu [](http://example.com/codeflow?start=30&length=96)

again me, some comments that nobody likes.

This function does 2 things: set memory and CPU.

Imagine a situation where only 1 is in need of adjustment. As it is written now, the user of this module MUST rememeber the OTHER value what it was set to before.

Suggestion is to split in 2 separate APIs.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
 {
-    (void) percent_physical_memory;
-    (void) percent_cpu;
-    return MU_FAILURE;
+    int result;
+    LogInfo("job_object_helper_limit_resources called with parameters:"
```
anporumb, line 15: >LogInfo [](http://example.com/codeflow?start=4&length=7)

georeplication-ism detected.

Only georeplication logs extensively parameters at every function entry/exit. c-pal and all the way up to EBS they do not.

Remove plz.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+    /*Codes_SRS_JOB_OBJECT_HELPER_18_002: [ If percent_cpu is 0, job_object_helper_limit_resources shall fail and return a non-zero value. ] */
+    /*Codes_SRS_JOB_OBJECT_HELPER_18_004: [ If percent_physical_memory is greater than 100, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+    /*Codes_SRS_JOB_OBJECT_HELPER_18_005: [ If percent_cpu is greater than 100, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+    if ((percent_physical_memory == 0) ||
+        (percent_physical_memory > 100) || 
+        (percent_cpu == 0) ||
+        (percent_cpu > 100))
+    {
+        LogError("Invalid arguments:"
+            " uint32_t percent_physical_memory = %d"
```
anporumb, line 31: >d [](http://example.com/codeflow?start=50&length=1)

PRiu32 (here and eveywhere)

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+            ", uint32_t percent_cpu = %d",
+            percent_physical_memory,
+            percent_cpu);
+        /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+        result = MU_FAILURE;
+    }
+    else
+    {
+        /*Codes_SRS_JOB_OBJECT_HELPER_18_006: [ job_object_helper_limit_resources shall call CreateJobObject to create a new job object passing NULL for both lpJobAttributes and lpName. ]*/
+        HANDLE job_object = CreateJobObject(NULL, NULL);
```
anporumb, line 41: >job_object [](http://example.com/codeflow?start=15&length=10)

This job_object should stay behind a HANDLE of the module

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+        if (job_object == NULL)
+        {
+            LogLastError("CreateJobObject(NULL, NULL) failed");
+            /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+            result = MU_FAILURE;
+        }
+        else
+        {
+            /*Codes_SRS_JOB_OBJECT_HELPER_18_007: [ job_object_helper_limit_resources shall call GetCurrentProcess to get the current process handle. ]*/
+            HANDLE current_process = GetCurrentProcess();
```
anporumb, line 51: >GetCurrentProcess [](http://example.com/codeflow?start=37&length=17)

needs processthreadsapi.h

also it cannot fail.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+            if (current_process == NULL)
+            {
+                LogLastError("GetCurrentProcess() failed");
+                /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+                result = MU_FAILURE;
+            }
+            else
+            {
+                /*Codes_SRS_JOB_OBJECT_HELPER_18_008: [ job_object_helper_limit_resources shall call AssignProcessToJobObject to assign the current process to the new job object. ]*/
+                if (AssignProcessToJobObject(job_object, current_process) == FALSE)
```
anporumb, line 61: > == FALSE [](http://example.com/codeflow?start=73&length=9)

probably nicer to write "!Assign..." like in line 84 and 103

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+                    LogLastError("AssignProcessToJobObject(HANDLE job_object = %p, HANDLE current_process = %p) failed", job_object, current_process); 
+                    /*Codes_SRS_JOB_OBJECT_HELPER_18_015: [ If there are any failures, job_object_helper_limit_resources shall fail and return a non-zero value. ]*/
+                    result = MU_FAILURE;
+                }
+                else
+                {
+                    MEMORYSTATUSEX memory_status_ex;
+                    memory_status_ex.dwLength = sizeof(memory_status_ex);
+                    /*Codes_SRS_JOB_OBJECT_HELPER_18_009: [ job_object_helper_limit_resources shall call GlobalMemoryStatusEx to get the total amount of physical memory. ]*/
+                    if (!GlobalMemoryStatusEx(&memory_status_ex))
```
anporumb, line 72: >GlobalMemoryStatusEx [](http://example.com/codeflow?start=25&length=20)

I would postpone line 61 until after this line.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+                    else
+                    {
+                        /*Codes_SRS_JOB_OBJECT_HELPER_18_010: [ job_object_helper_limit_resources shall call SetInformationJobObject, passing JobObjectExtendedLimitInformation and a JOBOBJECT_EXTENDED_LIMIT_INFORMATION object with JOB_OBJECT_LIMIT_JOB_MEMORY set and JobMemoryLimit set to the percent_physical_memory percent of the physical memory in bytes. ]*/
+                        JOBOBJECT_EXTENDED_LIMIT_INFORMATION extended_limit_information = {0};
+                        extended_limit_information.BasicLimitInformation.LimitFlags  = JOB_OBJECT_LIMIT_JOB_MEMORY;
+                        extended_limit_information.JobMemoryLimit = percent_physical_memory * memory_status_ex.ullTotalPhys / 100;
+                        if (!SetInformationJobObject(job_object, JobObjectExtendedLimitInformation, &extended_limit_information, sizeof(extended_limit_information)))
+                        {
+                            LogLastError("SetInformationJobObject("
+                                " job_object = %p"
```
anporumb, line 87: >  [](http://example.com/codeflow?start=33&length=1)

extra space in logging

anporumb: here and everywhere

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+                        }
+                        else
+                        {
+                            /*Codes_SRS_JOB_OBJECT_HELPER_18_011: [ job_object_helper_limit_resources shall call SetInformationJobObject passing JobObjectCpuRateControlInformation and a JOBOBJECT_CPU_RATE_CONTROL_INFORMATION object with JOB_OBJECT_CPU_RATE_CONTROL_ENABLE and JOB_OBJECT_CPU_RATE_CONTROL_HARD_CAP set, and CpuRate set to percent_cpu times 100. ]*/
+                            JOBOBJECT_CPU_RATE_CONTROL_INFORMATION cpu_rate_control_information = {0};
+                            cpu_rate_control_information.ControlFlags = JOB_OBJECT_CPU_RATE_CONTROL_ENABLE | JOB_OBJECT_CPU_RATE_CONTROL_HARD_CAP;
+                            cpu_rate_control_information.CpuRate = percent_cpu * 100;
+                            if (!SetInformationJobObject(job_object, JobObjectCpuRateControlInformation, &cpu_rate_control_information, sizeof(cpu_rate_control_information)))
+                            {
+                                LogLastError("SetInformationJobObject("
```
anporumb, line 105: >LogLastError [](http://example.com/codeflow?start=32&length=12)

assume line 105 gets executed. That implies that line 84 succeeded and did set the memory limit. However, the CPU limit failed.

At this moment I believe that either: the RAM should be reverted to "what it used to be before" or some other remedies.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+            }
+            /*Codes_SRS_JOB_OBJECT_HELPER_18_013: [ job_object_helper_limit_resources shall call CloseHandle to close the handle to the job object. ]*/
+            if (!CloseHandle(job_object))
+            {
+                LogLastError("CloseHandle(job_object = %p) failed", job_object);
+            }
+        }
+    }
+
+    LogInfo("job_object_helper_limit_resources exited with result = %d", result);
```
anporumb, line 138: >LogInfo("job_object_helper_limit_resources exited with result = %d", result); [](http://example.com/codeflow?start=4&length=77)

georeplication-ism detected.

win32/src/job_object_helper.c
```diff
@@ -11,7 +11,130 @@
...
+    return result;
 }
```
anporumb, line 140: file should end with an empty line

win32/tests/CMakeLists.txt
anporumb, line 36: >job_object_helper_ut [](http://example.com/codeflow?start=22&length=20)

I believe we need some job_object_int tests.

I would like to see if the API `job_object_helper_limit_resources` can be called twice, and if so, what's the effects.

Specifically, can the memory/CPU be un-restricted or only more-and-more restricted?

Also we can check that "it really works" with int tests,, like restrict memory to 1GB, try to allocate 1GB => fail. 

"unrestrict" memory, try to allocate 1GB=> success.

And so on.

anporumb: Job Object nesting and chains are... interesting.

anporumb: Anyway, I would be waiting because reasons, but can only sign off.

anporumb (approved):
:shipit:

win32/src/job_object_helper.c
BertKleewein: In retrospect, yes, I should have done that. Maybe for the next version.

win32/src/job_object_helper.c
BertKleewein: next time.

win32/src/job_object_helper.c
BertKleewein: Good idea, and this wouldn't be an issue if it were 2 separate APIs. I like it, but I don't want to change it now because of time.

win32/src/job_object_helper.c
BertKleewein: OK, I can see that. Easy enough.

win32/src/job_object_helper.c
BertKleewein: You're right. I'm not sure if I agree with removing the check, but it probably doesn't matter. I'll follow your lead and remove it.

win32/tests/CMakeLists.txt
BertKleewein: will create a task

PR #377, opened 2024-10-01
job_object_helper_v2 spec and header

BertKleewein:


win32/devdoc/job_object_helper.md
```diff
@@ -1,44 +1,123 @@
...
-    MOCKABLE_FUNCTION(, int, job_object_helper_limit_resources, uint32_t, percent_physical_memory, uint32_t, percent_cpu);
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
 ```
+``job_object_helper_create`` is used to create a `JOB_OBJECT_HELPER` object.
 
-`job_object_helper_limit_resources` is an API that uses the job object API to limit the amount of physical memory and CPU that the current process can consume.
+**SRS_JOB_OBJECT_HELPER_18_016: [** `job_object_helper_create` shall allocate a `JOB_OBJECT_HELPER` object. **]**
 
-**SRS_JOB_OBJECT_HELPER_18_001: [** If `percent_physical_memory` is `0`, `job_object_helper_limit_resources` shall fail and return a non-zero value. **]**
+**SRS_JOB_OBJECT_HELPER_18_017: [** `job_object_helper_create` shall call `sm_create`. **]**
```
anporumb, line 22: >sm_create [](http://example.com/codeflow?start=75&length=9)

why is SM needed here? This module is 
1) THANDLEd
2) doesn't have callbacks.

So I suspect that sm is not needed.

win32/inc/c_pal/job_object_helper.h
```diff
@@ -15,13 +15,20 @@
 
 #include "umock_c/umock_c_prod.h"
```
anporumb, line 16: keep this header right near `ifdef __cpluplus`, nothing in between.

anporumb:
> #include "stdint.h"

use the other way of `include` with `<` and `>`

---
Refers to: win32/src/job_object_helper.c:4 in 56a473b. [](commit_id = 56a473b7adb522048d4543227c408685a583c6e2, deletion_comment = False)

win32/devdoc/job_object_helper.md
```diff
@@ -1,44 +1,123 @@
...
+`job_object_helper_open` prepares the `JOB_OBJECT_HELPER` object for use by initializing the resources that it needs. 
+
+**SRS_JOB_OBJECT_HELPER_18_021: [** If `job_object_helper` is `NULL`, `job_object_helper_open` shall fail and return a non-zero value. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_022: [** `job_object_helper_open` shall call `sm_open_begin`. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_023: [** `job_object_helper_open` shall call `GlobalMemoryStatusEx` to get the total amount of physical memory in kb. **]**
 
-**SRS_JOB_OBJECT_HELPER_18_007: [** `job_object_helper_limit_resources` shall call `GetCurrentProcess` to get the current process handle. **]**
+**SRS_JOB_OBJECT_HELPER_18_024: [** `job_object_helper_open` shall call `CreateJobObject` to create a new job object passing `NULL` for both `lpJobAttributes` and `lpName`. **]**
```
anporumb, line 50: >CreateJobObject [](http://example.com/codeflow?start=73&length=15)

this for sure cannot be in the "open", needs to be moved to "create". 

I also suspect open and close and not needed (together with SM).

win32/devdoc/job_object_helper.md
```diff
@@ -1,44 +1,123 @@
...
+
+**SRS_JOB_OBJECT_HELPER_18_025: [** `job_object_helper_open` shall call `GetCurrentProcess` to get the current process handle. **]**
+
+**SRS_JOB_OBJECT_HELPER_18_026: [** `job_object_helper_open` shall call `AssignProcessToJobObject` to assign the current process to the new job object. **]**
```
anporumb, line 54: >AssignProcessToJobObject [](http://example.com/codeflow?start=73&length=24)

So in the following sequence of API calls:

create
open
close
open
close
open
close
destroy

How many distinct JOB_OBJECTS will end up being assigned to the process? And what happens when the parameters for these job objects (CPU/memory) are different/conflicting?

anporumb (approved):
:shipit:

win32/devdoc/job_object_helper.md
BertKleewein: good point.  getting rid of open/close and only using create/dispose makes this a non-issue

anporumb (approved):
:shipit:

PR #382, opened 2024-10-07
Change Timeouts to UINT32_MAX

nishudeshu:
Change InterlockedHL_WaitForValue timeouts to UINT32_MAX and remove the do while loop around it.

### DIFF
--- a/linux/tests/threadpool_linux_int/threadpool_linux_int.c
+++ b/linux/tests/threadpool_linux_int/threadpool_linux_int.c
@@ -175,11 +175,7 @@ TEST_FUNCTION(one_work_item_schedule_works)
 
     // assert
     LogInfo("Waiting for task to complete");
-    do
-    {
-        (void)InterlockedHL_WaitForValue(&thread_counter, num_threads, TEST_TIMEOUT_VALUE);
-    } while (interlocked_add(&thread_counter, 0) != num_threads);
-
+    ASSERT_ARE_EQUAL(int, INTERLOCKED_HL_OK, InterlockedHL_WaitForValue(&thread_counter, num_threads, UINT32_MAX));
     ASSERT_ARE_EQUAL(int32_t, thread_counter, num_threads, "Thread counter has timed out");
 
     // cleanup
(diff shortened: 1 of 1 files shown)


darobs (approved):

anporumb (approved):
:shipit:

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -884,11 +855,8 @@ TEST_FUNCTION(schedule_after_close_works)
 
     // assert
     LogInfo("Waiting for task to complete");
-    do
-    {
-        (void)wait_on_address(&thread_counter, num_threads, TEST_TIMEOUT_VALUE);
-    } while (interlocked_add(&thread_counter, 0) != num_threads);
-
+    // assert
+    ASSERT_ARE_EQUAL(int, INTERLOCKED_HL_OK, InterlockedHL_WaitForValue(&thread_counter, num_threads, UINT32_MAX));
```
dcristoloveanu, line 859: >int [](http://example.com/codeflow?start=21&length=3)

This should be INTERLOCKED_HL_RESULT. Here and everywhere.

dcristoloveanu:
>     } while (interlocked_add(&thread_counter, 0) != num_threads);

I think this loop is not needed also. Here and in the other tests also.

---
Refers to: linux/tests/threadpool_linux_int/threadpool_linux_int.c:838 in b09fb8a. [](commit_id = b09fb8ae6650fa753a8708a1d975ac58fada3b13, deletion_comment = False)

dcristoloveanu:
>     // arrange

COuld not help but notice that this test function does not have an // Act.
We usually have 

// Arrange
... test stup

// Act
.. here we act on the API

// Assert
... verify that when we performed the test action the expected things happened


---
Refers to: linux/tests/threadpool_linux_int/threadpool_linux_int.c:490 in b09fb8a. [](commit_id = b09fb8ae6650fa753a8708a1d975ac58fada3b13, deletion_comment = False)

dcristoloveanu:
>     // arrange

Might want to look at the other tests in the file just to have AAA everywhere

---
In reply to: [2397851395](https://github.com/Azure/c-pal/pull/382#issuecomment-2397851395) [](http://example.com/codeflow?ancestors=2397851395)

---
Refers to: linux/tests/threadpool_linux_int/threadpool_linux_int.c:490 in b09fb8a. [](commit_id = b09fb8ae6650fa753a8708a1d975ac58fada3b13, deletion_comment = False)

PR #384, opened 2024-10-10
add the smallest precompiled header and use it in 1x project.

anporumb:


jebrando (approved):
:shipit:

win32/reals/real_threadpool.c
```diff
@@ -3,6 +3,8 @@
 
 #include "real_gballoc_hl_renames.h"
 
+#include "real_ps_util_renames.h"
```
jebrando, line 6: duplicate include with line 16

PR #391, opened 2024-10-18
fixes for EBS repo and remove c_pal_interfacesmocks

anporumb:


common/reals/real_socket_transport.h
```diff
@@ -0,0 +1,49 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+SOCKET_TRANSPORT_HANDLE real_socket_transport_create_client(void);
+SOCKET_TRANSPORT_HANDLE real_socket_transport_create_server(void);
+SOCKET_TRANSPORT_HANDLE real_socket_transport_create_from_socket(SOCKET_HANDLE socket_handle);
+void real_socket_transport_destroy(SOCKET_TRANSPORT_HANDLE socket_transport);
+int real_socket_transport_connect(SOCKET_TRANSPORT_HANDLE socket_transport, const char* hostname, uint16_t port, uint32_t connection_timeout_ms);
+int real_socket_transport_listen(SOCKET_TRANSPORT_HANDLE socket_transport, uint16_t port);
```
dcristoloveanu, line 37: >uint16_t [](http://example.com/codeflow?start=75&length=8)

Likely could use an stdint include...

interfaces/inc/c_pal/async_socket.h
```diff
@@ -76,6 +73,10 @@ typedef struct ASYNC_SOCKET_BUFFER_TAG
     uint32_t length;
 } ASYNC_SOCKET_BUFFER;
 
+#include "umock_c/umock_c_prod.h"
```
dcristoloveanu, line 76: Consider having the include up there next to the other ones. Having includes as close to the top as possible can reduce developer annoyance  - because who scrolls down for includes :-). /consider

linux/linux_reals/real_async_socket.h
```diff
@@ -0,0 +1,45 @@
...
+extern "C" {
+#endif
+
+    ASYNC_SOCKET_HANDLE real_async_socket_create(EXECUTION_ENGINE_HANDLE execution_engine);
+    ASYNC_SOCKET_HANDLE real_async_socket_create_with_transport(EXECUTION_ENGINE_HANDLE execution_engine, ON_ASYNC_SOCKET_SEND on_send, void* on_send_context, ON_ASYNC_SOCKET_RECV on_recv, void* on_recv_context);
+    void real_async_socket_destroy(ASYNC_SOCKET_HANDLE async_socket);
+
+    int real_async_socket_open_async(ASYNC_SOCKET_HANDLE async_socket, SOCKET_TRANSPORT_HANDLE socket_transport, ON_ASYNC_SOCKET_OPEN_COMPLETE on_open_complete, void* on_open_complete_context);
+    void real_async_socket_close(ASYNC_SOCKET_HANDLE async_socket);
+    ASYNC_SOCKET_SEND_SYNC_RESULT real_async_socket_send_async(ASYNC_SOCKET_HANDLE async_socket, const ASYNC_SOCKET_BUFFER* payload, uint32_t buffer_count, ON_ASYNC_SOCKET_SEND_COMPLETE on_send_complete, void* on_send_complete_context);
```
dcristoloveanu, line 38: >uint32_t [](http://example.com/codeflow?start=133&length=8)

I sense the possibility of adding an stdint include here too

linux/src/async_socket_linux.c
```diff
@@ -176,12 +176,12 @@ static int on_socket_recv(void* context, ASYNC_SOCKET_HANDLE async_socket, void*
 
 #ifdef TEST_SUITE_NAME_FROM_CMAKE
 
-ON_ASYNC_SOCKET_SEND get_async_socket_send_callback()
+ON_ASYNC_SOCKET_SEND get_async_socket_send_callback(void)
 {
     return on_socket_send;
 }
 
-ON_ASYNC_SOCKET_RECV get_async_socket_recv_callback()
```
dcristoloveanu, line 184: >() [](http://example.com/codeflow?start=51&length=2)

Somebody writing C++ around here?

dcristoloveanu (dismissed):
:shipit:

linux/linux_reals/real_async_socket.h
anporumb: line 9 has it

interfaces/inc/c_pal/async_socket.h
anporumb: I considered, and this is our very special umock_c_prod which should ideally encompass only APIs (which is what this does here). I think I like it to stay here rather than someplace where it can mangle with other headers.

linux/src/async_socket_linux.c
anporumb: :)

dcristoloveanu (approved):
:shipit:

PR #393, opened 2024-10-21
re-enable traceability code in repo

jebrando:


anporumb (approved):
:shipit:

build/devops_gated.yml
```diff
@@ -12,6 +12,7 @@ resources:
 jobs:
 - template: /pipeline_templates/build_all_flavors.yml@c_build_tools
   parameters:
+    cmake_options: -Drun_unittests:BOOL=ON -Drun_e2e_tests:BOOL=ON -Drun_int_tests:BOOL=ON -Drun_perf_tests:BOOL=ON -Drun_traceability:BOOL=ON -Duse_cppunittest:BOOL=ON -Drun_reals_check:BOOL=ON
```
dcristoloveanu, line 15: I think we should enable traceability by default in c_build_tools and then if REALLY does not make sense for a repo, we should override it there, right?

PR #399, opened 2024-10-25
add timer state code+ut

M-iceberg:


win32/src/threadpool_win32.c
```diff
@@ -583,6 +584,7 @@ int threadpool_timer_start(THANDLE(THREADPOOL) threadpool, uint32_t start_delay_
                 }
                 else
                 {
+                    timer_temp->is_destroying = 0;
```
jasmineymlo, line 587: use interlocked_exchange

win32/src/threadpool_win32.c
```diff
@@ -661,13 +671,17 @@ void threadpool_timer_destroy(TIMER_INSTANCE_HANDLE timer)
...
+        (void)interlocked_exchange(&timer->is_destroying, 1);
         /* Codes_SRS_THREADPOOL_WIN32_42_012: [ threadpool_timer_destroy shall call SetThreadpoolTimer with NULL for pftDueTime and 0 for msPeriod and msWindowLength to cancel ongoing timers. ]*/
         /* Codes_SRS_THREADPOOL_WIN32_42_013: [ threadpool_timer_destroy shall call WaitForThreadpoolTimerCallbacks. ]*/
         threadpool_internal_cancel_timer_and_wait(timer->timer);
 
         /* Codes_SRS_THREADPOOL_WIN32_42_014: [ threadpool_timer_destroy shall call CloseThreadpoolTimer. ]*/
         CloseThreadpoolTimer(timer->timer);
 
+        /* Codes_SRS_THREADPOOL_WIN32_07_003: [ threadpool_timer_destroy shall indicate the timer is destroyed. ]*/
+        (void)interlocked_exchange(&timer->is_destroying, 0);
```
jasmineymlo, line 684: why are we setting it to 0? we are destroying it...

M-iceberg: will move it after the free

win32/src/threadpool_win32.c
M-iceberg: done

win32/src/threadpool_win32.c
```diff
@@ -625,6 +627,14 @@ int threadpool_timer_restart(TIMER_INSTANCE_HANDLE timer, uint32_t start_delay_m
             timer, start_delay_ms, timer_period_ms);
         result = MU_FAILURE;
     }
+    else if(
+        /* Codes_SRS_THREADPOOL_WIN32_07_001: [ If timer state is in destroying state, threadpool_timer_restart shall fail and return a non - zero value. ]*/
+        interlocked_add(&timer->is_destroying, 0) == 1
+        )
+    {
+        LogError("timer is in closing state");
+        result = MU_FAILURE;
```
jasmineymlo, line 636: seems like we shouldn't return MU_FAILURE if it's due to closing...otherwise will fault...

jasmineymlo:
>         threadpool_internal_set_timer(timer->timer, start_delay_ms, timer_period_ms);

seems like we could still have destroy going after we check...then we would end up in same situation?


---
Refers to: win32/src/threadpool_win32.c:641 in ef31a77. [](commit_id = ef31a77ded4d5095db22f5f470dbe8512f00aa98, deletion_comment = False)

win32/devdoc/threadpool_win32.md
```diff
@@ -215,7 +215,7 @@ MOCKABLE_FUNCTION(, int, threadpool_timer_restart, TIMER_INSTANCE_HANDLE, timer,
 
 **SRS_THREADPOOL_WIN32_42_019: [** If `timer` is `NULL`, `threadpool_timer_restart` shall fail and return a non-zero value. **]**
 
-If timer state is `DESTROYING`, `threadpool_timer_restart` shall fail and return a non-zero value.
```
dcristoloveanu, line 218: >DESTROYING [](http://example.com/codeflow?start=19&length=10)

These states should be described at the top of the .md (Otherwise when reading this, I have no idea when we switch to the state and when we transition out of it) ...

win32/devdoc/threadpool_win32.md
```diff
@@ -215,7 +215,7 @@ MOCKABLE_FUNCTION(, int, threadpool_timer_restart, TIMER_INSTANCE_HANDLE, timer,
...
+**SRS_THREADPOOL_WIN32_07_001: [** If timer state is in destroying state, `threadpool_timer_restart` shall fail and return a non-zero value. **]**
```
dcristoloveanu, line 218: >destroying [](http://example.com/codeflow?start=56&length=10)

I think capital and enclosed by tickmarks was better.

dcristoloveanu: Because they are references to states that should be defined in the doc.

win32/src/threadpool_win32.c
```diff
@@ -625,6 +627,14 @@ int threadpool_timer_restart(TIMER_INSTANCE_HANDLE timer, uint32_t start_delay_m
             timer, start_delay_ms, timer_period_ms);
         result = MU_FAILURE;
     }
+    else if(
+        /* Codes_SRS_THREADPOOL_WIN32_07_001: [ If timer state is in destroying state, threadpool_timer_restart shall fail and return a non - zero value. ]*/
+        interlocked_add(&timer->is_destroying, 0) == 1
+        )
+    {
+        LogError("timer is in closing state");
```
dcristoloveanu, line 635: >closing [](http://example.com/codeflow?start=30&length=7)

I thought the states are CREATED/DESTROYING/DESTROYED.
A wild closing appears :-)

win32/src/threadpool_win32.c
```diff
@@ -625,6 +627,14 @@ int threadpool_timer_restart(TIMER_INSTANCE_HANDLE timer, uint32_t start_delay_m
             timer, start_delay_ms, timer_period_ms);
         result = MU_FAILURE;
     }
+    else if(
+        /* Codes_SRS_THREADPOOL_WIN32_07_001: [ If timer state is in destroying state, threadpool_timer_restart shall fail and return a non - zero value. ]*/
+        interlocked_add(&timer->is_destroying, 0) == 1
```
dcristoloveanu, line 632: >is_destroying [](http://example.com/codeflow?start=32&length=13)

Can we have this as a proper enum? With the 3 states we have

win32/devdoc/threadpool_win32.md
```diff
@@ -245,15 +245,15 @@ MOCKABLE_FUNCTION(, void, threadpool_timer_destroy, TIMER_INSTANCE_HANDLE, timer
 
 **SRS_THREADPOOL_WIN32_42_011: [** If `timer` is `NULL`, `threadpool_timer_destroy` shall fail and return. **]**
 
-`threadpool_timer_destroy` shall switch the state to `DESTROYING`.
+**SRS_THREADPOOL_WIN32_07_002: [** `threadpool_timer_destroy` shall indicate the timer is in destroying state. **]**
```
dcristoloveanu, line 248: > [** `threadpool_t [](http://example.com/codeflow?start=30&length=18)

I find these specs rather poor.
As a reader I can only try to imagine what we want to achieve.
I think we should describe the states we have and why we have them:

I assume we have something along the lines of:
CREATED - state in which the timer is after threadpool_timer_start was called.
DESTROYING - state in which the timer is after threadpool_timer_destroy was called. A threadpool_timer_restart call after switching to DESTROYING state is failed.
DESTROYED - terminal state for the timer.

dcristoloveanu (changes requested):
:clock1:

win32/devdoc/threadpool_win32.md
M-iceberg: done

win32/devdoc/threadpool_win32.md
M-iceberg: added

win32/devdoc/threadpool_win32.md
M-iceberg: done

win32/src/threadpool_win32.c
M-iceberg: done

win32/src/threadpool_win32.c
M-iceberg: modified

dcristoloveanu:
> MOCKABLE_FUNCTION(, void, threadpool_timer_destroy, TIMER_INSTANCE_HANDLE, timer);

One more interesting comment on this.
If this handle woutl be THANDLE, isn't it true that we could not have by design a race where restart is called while destroy happens?

---
Refers to: win32/devdoc/threadpool_win32.md:241 in d68b9c8. [](commit_id = d68b9c854a160bffe858a25cbbdb0b931331685c, deletion_comment = False)

PR #404, opened 2024-10-30
updating wait_on_address to return the timeout on waits

jebrando:
Currently on calls to InterlockedHL_WaitForValue and InterlockedHL_WaitForNotValue if the wait_on_address call times out it returns and INTERLOCKED_HL_ERROR, which will work, but sometimes we need to know that the wait call has timed out.


common/src/interlocked_hl.c
```diff
@@ -107,10 +107,19 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
 
             /* Codes_SRS_INTERLOCKED_HL_01_004: [ If the value at address is not equal to value, InterlockedHL_WaitForValue shall wait until the value at address changes in order to compare it again to value by using wait_on_address. ]*/
             /* Codes_SRS_INTERLOCKED_HL_01_005: [ When waiting for the value at address to change, the milliseconds argument value shall be used as timeout. ]*/
-            if (wait_on_address(address, current_value, milliseconds) != WAIT_ON_ADDRESS_OK)
+            WAIT_ON_ADDRESS_RESULT wait_result = wait_on_address(address, current_value, milliseconds);
+            if (wait_result != WAIT_ON_ADDRESS_OK)
             {
-                /* Codes_SRS_INTERLOCKED_HL_01_006: [ If wait_on_address fails, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_ERROR. ]*/
-                result = INTERLOCKED_HL_ERROR;
+                if (wait_result == WAIT_ON_ADDRESS_TIMEOUT)
```
anporumb, line 113: just wondering if this can be done without evaluating twice `wait_result`. Maybe a `switch` statement?

common/src/interlocked_hl.c
```diff
@@ -107,10 +107,19 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
+                {
+                    /* Codes_SRS_INTERLOCKED_HL_11_001: [ If wait_on_address timesout, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_TIMEOUT. ] */
+                    result = INTERLOCKED_HL_TIMEOUT;
+                }
+                else
+                {
+                    /* Codes_SRS_INTERLOCKED_HL_01_006: [ If wait_on_address fails, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_ERROR. ]*/
+                    result = INTERLOCKED_HL_ERROR;
```
anporumb, line 121: >INTERLOCKED_HL_ERROR [](http://example.com/codeflow?start=29&length=20)

Line 171 Logs an error here, probably we should log something here too.

After all - I believe we'll never see it, but for those cases when we'd potentially-virtually-allegedly see it - we'd better see it.

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -401,6 +401,27 @@ TEST_FUNCTION(when_the_wait_on_address_fails_InterlockedHL_WaitForValue_also_fai
...
+TEST_FUNCTION(when_the_wait_on_address_timesout_InterlockedHL_WaitForValue_also_timesout)
+{
+    // arrange
+    INTERLOCKED_HL_RESULT result;
+    volatile_atomic int32_t value = 0x40;
+    int32_t intermediate_value = 0x41;
+
+    STRICT_EXPECTED_CALL(interlocked_add(&value, 0));
+    STRICT_EXPECTED_CALL(wait_on_address(&value, IGNORED_ARG, UINT32_MAX))
+        .CopyOutArgumentBuffer_address(&intermediate_value, sizeof(int32_t))
```
anporumb, line 414: >CopyOutArgumentBuffer_address [](http://example.com/codeflow?start=9&length=29)

why is this needed if it times out?

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -401,6 +401,27 @@ TEST_FUNCTION(when_the_wait_on_address_fails_InterlockedHL_WaitForValue_also_fai
...
+        .SetReturn(WAIT_ON_ADDRESS_TIMEOUT);
+
+    // act
+    result = InterlockedHL_WaitForValue(&value, 0x42, UINT32_MAX);
```
anporumb, line 418: >UINT32_MAX [](http://example.com/codeflow?start=54&length=10)

probably use any other value than UINT32_MAX?

Just for "play pretend".

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -504,6 +525,25 @@ TEST_FUNCTION(when_the_wait_on_address_fails_InterlockedHL_WaitForNotValue_also_
...
+    STRICT_EXPECTED_CALL(interlocked_add(&value, 0));
+    STRICT_EXPECTED_CALL(wait_on_address(&value, IGNORED_ARG, UINT32_MAX))
+        .SetReturn(WAIT_ON_ADDRESS_TIMEOUT);
+
+    // act
+    result = InterlockedHL_WaitForNotValue(&value, 0x42, UINT32_MAX);
+
+    // assert
+    ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_TIMEOUT, result);
```
anporumb, line 544: >INTERLOCKED_HL_TIMEOUT [](http://example.com/codeflow?start=44&length=22)

must have int tests for this API, where we observe this TIMEOUT actually working.

... will wait for those to be written, but weakly.

anporumb (changes requested):
:clock1:

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
jebrando: Added int test

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
jebrando: copy paste from the above test.

common/src/interlocked_hl.c
jebrando: I can make that happen.

jebrando:
> IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotValue, int32_t volatile_atomic*, address, int32_t, value, uint32_t, milliseconds)

I agree with the sentiment but that would be for another PR

---
In reply to: [2450124780](https://github.com/Azure/c-pal/pull/404#issuecomment-2450124780) [](http://example.com/codeflow?ancestors=2450124780)

---
Refers to: common/src/interlocked_hl.c:131 in ef6a042. [](commit_id = ef6a0429b6b0dd68376ffe58423e9c5b981af995, deletion_comment = False)

common/src/interlocked_hl.c
```diff
@@ -149,12 +158,21 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotVal
...
-                /* Codes_SRS_INTERLOCKED_HL_42_007: [ If wait_on_address fails, InterlockedHL_WaitForNotValue shall fail and return INTERLOCKED_HL_ERROR. ]*/
-                result = INTERLOCKED_HL_ERROR;
+                if (wait_result == WAIT_ON_ADDRESS_TIMEOUT)
+                {
+                    /* Codes_SRS_INTERLOCKED_HL_11_002: [ If wait_on_address timesout, InterlockedHL_WaitForNotValue shall fail and return INTERLOCKED_HL_TIMEOUT. ] */
+                    result = INTERLOCKED_HL_TIMEOUT;
+                }
+                else
+                {
+                    LogError("failure in wait_on_address(address=%p, &current_value=%p, milliseconds=%" PRIu32 ")",
```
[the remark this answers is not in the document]
jebrando: I don't want a log in the timeout due to this will be a normal operation in the limiter and I don't want to flood.  the error case is a different subject and I can add it there.

common/src/interlocked_hl.c
```diff
@@ -100,20 +101,41 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
+            WAIT_ON_ADDRESS_RESULT wait_result = wait_on_address(address, current_value, milliseconds);
+            switch (wait_result)
             {
-                /* Codes_SRS_INTERLOCKED_HL_01_006: [ If wait_on_address fails, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_ERROR. ]*/
-                result = INTERLOCKED_HL_ERROR;
-                break;
+                case WAIT_ON_ADDRESS_OK:
+                {
+                    result = INTERLOCKED_HL_OK;
+                    break;
```
anporumb, line 118: Yeah, so there's problems with the code here, and that's VERY unfortunate because these problems also point out to the tests being insufficient.

Here's what CAN happen:
line 82: InterlockedHL_WaitForValue is called with value=1
line 99: observes the value being "2". 
line 112: wait_on_address returns `WAIT_ON_ADDRESS_OK` but it is a spurious wake-up ([https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress)), so the value is still "2". Note: return is "WAIT_ON_ADDRESS_OK".
line 118: returns to the user "OK".

At no time during this execution did the value get to be "1". 

The fact that this is not captured by tests is concerning.

common/src/interlocked_hl.c
```diff
@@ -91,6 +91,7 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
     else
     {
         int32_t current_value;
+        bool continue_eval = true;
```
anporumb, line 94: >bool continue_eval = true; [](http://example.com/codeflow?start=8&length=26)

I highly suspect that the new code can be reformulated without the use of "continue_eval". Probably a combination of "continue" and "break" keywords would do it.

common/src/interlocked_hl.c
```diff
@@ -100,20 +101,41 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
             {
                 /* Codes_SRS_INTERLOCKED_HL_01_003: [ If the value at address is equal to value, InterlockedHL_WaitForValue shall return INTERLOCKED_HL_OK. ]*/
                 result = INTERLOCKED_HL_OK;
+                continue_eval = false;
                 break;
```
anporumb, line 105: this is "break".

common/src/interlocked_hl.c
anporumb, line 118: this is "continue"

common/src/interlocked_hl.c
```diff
@@ -100,20 +101,41 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
+                {
+                    result = INTERLOCKED_HL_OK;
+                    break;
+                }
+                case WAIT_ON_ADDRESS_TIMEOUT:
+                {
+                    /* Codes_SRS_INTERLOCKED_HL_11_001: [ If wait_on_address timesout, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_TIMEOUT. ] */
+                    result = INTERLOCKED_HL_TIMEOUT;
+                    continue_eval = false;
+                    break;
```
anporumb, line 125: this is "break" (from do/while)

common/src/interlocked_hl.c
```diff
@@ -100,20 +101,41 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
+                }
+                case WAIT_ON_ADDRESS_ERROR:
+                default:
+                {
+                    LogError("failure in wait_on_address(address=%p, &current_value=%p, milliseconds=%" PRIu32 ") result: %" PRI_MU_ENUM "",
+                        address, &current_value, milliseconds, MU_ENUM_VALUE(WAIT_ON_ADDRESS_RESULT, wait_result));
+                    /* Codes_SRS_INTERLOCKED_HL_01_006: [ If wait_on_address fails, InterlockedHL_WaitForValue shall fail and return INTERLOCKED_HL_ERROR. ]*/
+                    result = INTERLOCKED_HL_ERROR;
+                    continue_eval = false;
+                    break;
```
anporumb, line 135: this is "break" (from do/while)/.

common/src/interlocked_hl.c
```diff
@@ -100,20 +101,41 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForValue,
...
                 break;
             }
 
             /* Codes_SRS_INTERLOCKED_HL_01_008: [ If the value at address does not match, InterlockedHL_WaitForValue shall issue another call to wait_on_address. ]*/
 
             /* Codes_SRS_INTERLOCKED_HL_01_004: [ If the value at address is not equal to value, InterlockedHL_WaitForValue shall wait until the value at address changes in order to compare it again to value by using wait_on_address. ]*/
             /* Codes_SRS_INTERLOCKED_HL_01_005: [ When waiting for the value at address to change, the milliseconds argument value shall be used as timeout. ]*/
-            if (wait_on_address(address, current_value, milliseconds) != WAIT_ON_ADDRESS_OK)
+            WAIT_ON_ADDRESS_RESULT wait_result = wait_on_address(address, current_value, milliseconds);
+            switch (wait_result)
```
anporumb, line 113: >switch [](http://example.com/codeflow?start=12&length=6)

since "break" is overloaded in C for both switch and do/while, I would suggest re-writing the switch with "if" statements.

common/src/interlocked_hl.c
jebrando: I'm guessing the test writer did not think of surious wake-ups.  I'll add a test

anporumb: I will ... errr... reformulate my comment from above since it is factually inaccurate. 

As I've self-observed, `break` is overloaded (in another comment).

All i've said can be discarded but don't set "result" to anything because that's confusing :)

common/src/interlocked_hl.c
```diff
@@ -142,22 +165,42 @@ IMPLEMENT_MOCKABLE_FUNCTION(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNotVal
...
+                /* Codes_SRS_INTERLOCKED_HL_42_006: [ If the value at address matches, InterlockedHL_WaitForNotValue shall issue another call to wait_on_address. ]*/
+
+                /* Codes_SRS_INTERLOCKED_HL_42_003: [ If the value at address is equal to value, InterlockedHL_WaitForNotValue shall wait until the value at address changes in order to compare it again to value by using wait_on_address. ]*/
+                /* Codes_SRS_INTERLOCKED_HL_42_004: [ When waiting for the value at address to change, the milliseconds argument value shall be used as timeout. ]*/
+                WAIT_ON_ADDRESS_RESULT wait_result = wait_on_address(address, current_value, milliseconds);
+                switch (wait_result)
+                {
+                    case WAIT_ON_ADDRESS_OK:
+                    {
+                        result = INTERLOCKED_HL_OK;
```
anporumb, line 181: >result = INTERLOCKED_HL_OK; [](http://example.com/codeflow?start=24&length=27)

same comments as above for all the code.

wait_on_address can wake up for WHATEVER reason it wants (... tsk tsk tsk).

Assigning here "result" cannot be done, because a determination of the "real" value has not been performed yet (only line 163) does that "for real".

anporumb (approved):
:shipit:

common/src/interlocked_hl.c
jebrando: After looking there is a test for this

common/src/interlocked_hl.c
jebrando: fixed

common/src/interlocked_hl.c
jebrando: fixed

common/src/interlocked_hl.c
jebrando: done

common/src/interlocked_hl.c
jebrando: taken care of

nishudeshu (approved):

PR #405, opened 2024-11-01
Add real_threadpool_thandle for convenience in UTs

mattdurak:
Adds reals just for the THANDLE part of threadpool for leak detection in UTs


jebrando (approved):
:shipit:

PR #406, opened 2024-11-05
add a threadpool chaos test

M-iceberg:


win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -998,6 +998,139 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
     return 0;
 }
 
+//static DWORD WINAPI chaos_thread_with_timers_no_lock_func(LPVOID lpThreadParameter)
```
dcristoloveanu, line 1001: The easier way to have this test disabled is to have an XTEST_FUNCTION defined which does not expand to the ctest code.

Look for it in the code base, you will find some examples.
The advantage is that in that case the code gets compiled and we get checks on the fact that this actually compiles Ok, even though the test does not get to run.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -998,6 +998,139 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
...
+//{
+//    CHAOS_TEST_DATA* chaos_test_data = (CHAOS_TEST_DATA*)lpThreadParameter;
```
dcristoloveanu, line 1003: >(CHAOS_TEST_DATA*) [](http://example.com/codeflow?start=41&length=18)

I don't think the cast is needed

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -998,6 +998,139 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
...
+//
+//    while (InterlockedAdd(&chaos_test_data->chaos_test_done, 0) == 0)
+//    {
+//        int which_action = rand() * 8 / (RAND_MAX + 1);
```
dcristoloveanu, line 1007: >which_action [](http://example.com/codeflow?start=14&length=12)

It is better to have an enum defined with the possible actions.
Then it becomes more readable (instead of having 0, 1, 2, we'd have CHAOS_TEST_TIMER_ACTION_THREADPOOL_OPEN, ... for example

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1058,6 +1191,69 @@ TEST_FUNCTION(chaos_knight_test)
     execution_engine_dec_ref(execution_engine);
 }
 
+// test used for detect race condition between timer_restart/timer_cancel and timer destory, failed due to the race condition for the current code, will uncomment after the fix
+//TEST_FUNCTION(chaos_knight_test_with_timers_no_lock)
```
dcristoloveanu, line 1195: >chaos_knight_test_with_timers_no_lock [](http://example.com/codeflow?start=16&length=37)

Isn't this chaos test stronger than the other one? Because it has no locks supposedly ...?

If there is no other real difference between the 2 tests, we should remove the existing test (seems to do quite the same)

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1058,6 +1191,69 @@ TEST_FUNCTION(chaos_knight_test)
...
+//{
+//    // start a number of threads and each of them will do a random action on the threadpool
+//    EXECUTION_ENGINE_PARAMETERS execution_engine_parameters = { 4, 0 };
+//    EXECUTION_ENGINE_HANDLE execution_engine = execution_engine_create(&execution_engine_parameters);
```
dcristoloveanu, line 1199: >execution_engine [](http://example.com/codeflow?start=30&length=16)

This should be asserted to not be NULL.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1058,6 +1191,69 @@ TEST_FUNCTION(chaos_knight_test)
...
+//    (void)InterlockedExchange64(&chaos_test_data.executed_work_functions, 0);
+//    (void)InterlockedExchange64(&chaos_test_data.executed_timer_functions, 0);
+//    (void)InterlockedExchange(&chaos_test_data.timers_starting, 0);
+//    (void)InterlockedExchange(&chaos_test_data.chaos_test_done, 0);
+//    (void)InterlockedExchange(&chaos_test_data.can_start_timers, 1);
+//
+//    for (i = 0; i < MAX_TIMER_COUNT; i++)
+//    {
+//        chaos_test_data.timers[i].timer = NULL;
+//        InterlockedExchange(&chaos_test_data.timers[i].state, TIMER_STATE_NONE);
```
dcristoloveanu, line 1219: >InterlockedExchange [](http://example.com/codeflow?start=10&length=19)

(void)

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1058,6 +1191,69 @@ TEST_FUNCTION(chaos_knight_test)
...
+//    }
+//
+//    chaos_test_data.work_item_context = threadpool_create_work_item(chaos_test_data.threadpool, work_function, (void*)&chaos_test_data.executed_work_functions);
```
dcristoloveanu, line 1222: >work_item_context [](http://example.com/codeflow?start=22&length=17)

There should be an assert on this not being NULL.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1058,6 +1191,69 @@ TEST_FUNCTION(chaos_knight_test)
...
+//
+//    for (i = 0; i < CHAOS_THREAD_COUNT; i++)
+//    {
+//        thread_handles[i] = CreateThread(NULL, 0, chaos_thread_with_timers_no_lock_func, &chaos_test_data, 0, NULL);
+//        ASSERT_IS_NOT_NULL(thread_handles[i], "thread %zu failed to start", i);
+//    }
+//
+//    // wait for some time
+//    Sleep(10000);
```
dcristoloveanu, line 1231: >10000 [](http://example.com/codeflow?start=12&length=5)

I think this should be some sort of define (like TEST_RUNTIME)

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -998,6 +998,139 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
...
+//        {
+//            // Synchronize with close
+//            (void)InterlockedIncrement(&chaos_test_data->timers_starting);
+//            if (InterlockedAdd(&chaos_test_data->can_start_timers, 0) != 0)
+//            {
+//                int which_timer_slot = rand() * MAX_TIMER_COUNT / (RAND_MAX + 1);
+//                if (InterlockedAdd(&chaos_test_data->timers[which_timer_slot].state, 0) == TIMER_STATE_STARTED)
+//                {
+//                    threadpool_timer_cancel(chaos_test_data->timers[which_timer_slot].timer);
+//                    //(void)InterlockedExchange(&chaos_test_data->timers[which_timer_slot].state, TIMER_STATE_STARTED);
```
dcristoloveanu, line 1095: >//(void)InterlockedExchange(&chaos_test_data->timers[which_timer_slot].state, TIMER_STATE_STARTED); [](http://example.com/codeflow?start=22&length=99)

This seems to be commented out, should it?

dcristoloveanu (changes requested):
:clock1:

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: removed

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: done

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: removed the existing test

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: added

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: added

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: done

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: removed

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: this one is NULL in the chaos test because this test only test if it would crash, schedule work item with non-NULL context was tested in the following several test functions.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -787,6 +799,7 @@ TEST_FUNCTION(open_while_closing_fails)
 }
 
 #define CHAOS_THREAD_COUNT 4
+#define TEST_RUN_TIME 10000
```
dcristoloveanu, line 802: >10000 [](http://example.com/codeflow?start=22&length=5)

Can we have a // ms
comment next to it so we know it is expressed in ms?

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -896,96 +912,94 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
...
+                    ASSERT_ARE_EQUAL(int, 0, threadpool_timer_restart(chaos_test_data->timers[which_timer_slot].timer, timer_start_delay, timer_period));
                 }
-                (void)InterlockedDecrement(&chaos_test_data->timers_starting);
-                WakeByAddressSingle((PVOID)&chaos_test_data->timers_starting);
             }
-        case 7:
+            (void)InterlockedDecrement(&chaos_test_data->timers_starting);
+            WakeByAddressSingle((PVOID)&chaos_test_data->timers_starting);
+        }
+        case TEST_ACTION_SCEDULE_WORK_ITEM:
```
dcristoloveanu, line 1002: >TEST_ACTION_SCEDULE_WORK_ITEM [](http://example.com/codeflow?start=13&length=29)

Nice, ty

win32/tests/threadpool_win32_int/threadpool_win32_int.c
dcristoloveanu: I am confused, doesn't threadpool_create_work_item return something non-NULL always?

dcristoloveanu: >(void*) [](http://example.com/codeflow?start=113&length=7)

Is this cast needed?

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -896,96 +912,94 @@ static DWORD WINAPI chaos_thread_with_timers_func(LPVOID lpThreadParameter)
...
                 {
-                    int which_timer_slot = rand() * MAX_TIMER_COUNT / (RAND_MAX + 1);
-                    if (InterlockedCompareExchange(&chaos_test_data->timers[which_timer_slot].state, TIMER_STATE_STOPPING, TIMER_STATE_STARTED) == TIMER_STATE_STARTED)
-                    {
-                        threadpool_timer_destroy(chaos_test_data->timers[which_timer_slot].timer);
-                        chaos_test_data->timers[which_timer_slot].timer = NULL;
-                        InterlockedExchange(&chaos_test_data->timers[which_timer_slot].state, TIMER_STATE_NONE);
-                    }
+                    threadpool_timer_destroy(chaos_test_data->timers[which_timer_slot].timer);
+                    //chaos_test_data->timers[which_timer_slot].timer = NULL;
```
dcristoloveanu, line 959: >//chaos_test_data->timers[which_timer_slot].timer = NULL; [](http://example.com/codeflow?start=20&length=57)

Commented out code detected.

dcristoloveanu: Either the code is not needed at all in which case it should be removed, or it is neededand then it should be uncommented

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: changed to return a non-NULL context and added the assert.

M-iceberg: yeah it's needed because compiler complains quantifiers are different

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: yeah it's needed, uncomment the code

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -865,20 +878,23 @@ static void chaos_cleanup_all_timers(CHAOS_TEST_DATA* chaos_test_data)
...
 
     while (InterlockedAdd(&chaos_test_data->chaos_test_done, 0) == 0)
     {
-        int which_action = rand() * 8 / (RAND_MAX + 1);
+        int which_action = rand() * (MU_ENUM_VALUE_COUNT(TEST_ACTION_VALUES) - 1) / RAND_MAX + 1;
         switch (which_action)
         {
-        case 0:
+        default:
+            // do nothing
```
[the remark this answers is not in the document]
M-iceberg: added

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: threadpool_create_work_item returns NULL when threadpool is not open, the existing test didn't open the threadpool before this line when I copied from the old test, so get a NULL work_item and didn't actually schedule any work item in the chaos test, the old test directly goes into break in the case of schedule work item/schedule work due to the NULL check. I changed it to return a non-NULL context in the create_work_item and actually run some work items now in the chaos test.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -800,11 +814,12 @@ typedef struct CHAOS_TEST_DATA_TAG
 {
     volatile LONG64 expected_call_count;
     volatile LONG64 executed_work_functions;
-    volatile LONG64 executed_timer_functions;
+    volatile LONG executed_timer_functions;
```
dcristoloveanu, line 817: > LONG executed_t [](http://example.com/codeflow?start=12&length=16)

I think we could keep this 64 bit. It is hard but not impossible to fire 4B timers. But it will be absolutely impossible with current CPUs to fire as many times to exhaust 64 bit.

dcristoloveanu (dismissed):
:shipit:

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: changed back to 64

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Typo: SCHEDULE instead of SCEDULE.

M-iceberg: fixed

nishudeshu (approved):

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1082,22 +1108,28 @@ TEST_FUNCTION(chaos_knight_test_with_timers)
...
     // wait for some time
-    Sleep(10000);
+    Sleep(TEST_RUN_TIME);
 
     (void)InterlockedExchange(&chaos_test_data.chaos_test_done, 1);
 
+    //will change to use InterlockedHL_waitforvalue_64 later, task:https://msazure.visualstudio.com/One/_workitems/edit/30197585
+    while ((int64_t)InterlockedAdd64(&chaos_test_data.expected_call_count, 0) != (int64_t)InterlockedAdd64(&chaos_test_data.executed_work_functions, 0))
+    {
+        continue;
```
dcristoloveanu, line 1130: >continue; [](http://example.com/codeflow?start=8&length=9)

continue not needed

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1082,22 +1108,28 @@ TEST_FUNCTION(chaos_knight_test_with_timers)
...
         ASSERT_IS_NOT_NULL(thread_handles[i], "thread %zu failed to start", i);
     }
 
     // wait for some time
-    Sleep(10000);
+    Sleep(TEST_RUN_TIME);
 
     (void)InterlockedExchange(&chaos_test_data.chaos_test_done, 1);
 
+    //will change to use InterlockedHL_waitforvalue_64 later, task:https://msazure.visualstudio.com/One/_workitems/edit/30197585
```
dcristoloveanu, line 1127: >InterlockedHL_waitforvalue_64 [](http://example.com/codeflow?start=25&length=29)

Nitpick: InterlockedHL_WaitForValue_64 is likely the future API name

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1082,22 +1108,28 @@ TEST_FUNCTION(chaos_knight_test_with_timers)
...
+    while ((int64_t)InterlockedAdd64(&chaos_test_data.expected_call_count, 0) != (int64_t)InterlockedAdd64(&chaos_test_data.executed_work_functions, 0))
```
dcristoloveanu, line 1128: >(int64_t) [](http://example.com/codeflow?start=81&length=9)

Why would this cast be needed?

win32/tests/threadpool_win32_int/threadpool_win32_int.c
dcristoloveanu, line 1130: >continue [](http://example.com/codeflow?start=8&length=8)

We should yield in this while (like a Sleep(1))

dcristoloveanu (dismissed):
:shipit:

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: done

M-iceberg: changes to sleep(1);

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: added in the comment

win32/tests/threadpool_win32_int/threadpool_win32_int.c
M-iceberg: removed

nishudeshu (approved):

dcristoloveanu (approved):
:shipit:

PR #408, opened 2024-11-07
Threadpool open and close function scope reduction

nishudeshu:
threadpool_open functions work will be performed by threadpool_create. threadpool_open will be around to support other modules
threadpool_close functions work will be performed by threadpool_dispose. threadpool_close will be around to support other modules.

### DIFF
--- a/win32/devdoc/threadpool_win32.md
+++ b/win32/devdoc/threadpool_win32.md
@@ -74,13 +74,11 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees all resources associated with threadpool.
 
-**SRS_THREADPOOL_WIN32_01_007: [** `threadpool_dispose` shall perform an implicit close. **]**
+**SRS_THREADPOOL_WIN32_01_030: [** `threadpool_dispose` shall wait for any executing callbacks by calling `CloseThreadpoolCleanupGroupMembers`, passing `FALSE` as `fCancelPendingCallbacks`. **]**
 
-- **SRS_THREADPOOL_WIN32_01_030: [** `threadpool_dispose` shall wait for any executing callbacks by calling `CloseThreadpoolCleanupGroupMembers`, passing `FALSE` as `fCancelPendingCallbacks`. **]**
+**SRS_THREADPOOL_WIN32_01_032: [** `threadpool_dispose` shall close the threadpool cleanup group by calling `CloseThreadpoolCleanupGroup`. **]**
 
-- **SRS_THREADPOOL_WIN32_01_032: [** `threadpool_dispose` shall close the threadpool cleanup group by calling `CloseThreadpoolCleanupGroup`. **]**
-
-- **SRS_THREADPOOL_WIN32_01_033: [** `threadpool_dispose` shall destroy the thread pool environment created in `threadpool_create`. **]**
+**SRS_THREADPOOL_WIN32_01_033: [** `threadpool_dispose` shall destroy the thread pool environment created in `threadpool_create`. **]**
 
 **SRS_THREADPOOL_WIN32_42_028: [** `threadpool_dispose` shall decrement the reference count on the `execution_engine`. **]**
 
--- a/win32/devdoc/threadpool_win32.md
+++ b/win32/devdoc/threadpool_win32.md
@@ -104,7 +104,7 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 Note: `threadpool_close` will be deprecated and threadpool_dispose will perform additional tasks of `threadpool_close`. This function will exist until all the libraries calling this API are modified to use only `threadpool_create`.
 `threadpool_close` closes an open threadpool.
 
-**SRS_THREADPOOL_WIN32_05_018: [** If `threadpool` is `NULL`, `threadpool_close` shall return. **]**
+**SRS_THREADPOOL_WIN32_05_019: [** If `threadpool` is `NULL`, `threadpool_close` shall return. **]**
 
 ### threadpool_schedule_work
 
@@ -304,6 +304,8 @@ MOCKABLE_FUNCTION(, void, threadpool_destroy_work_item, THANDLE(THREADPOOL), thr
 
 **SRS_THREADPOOL_WIN32_05_015: [** If `work_item_context` is `NULL`, `threadpool_destroy_work_item` shall fail and not do anything before returning. **]**
 
-**SRS_THREADPOOL_WIN32_05_016: [** `threadpool_destroy_work_item` shall call `CloseThreadpoolWork` to close `ptp_work`. **]**
+**SRS_THREADPOOL_WIN32_05_016: [** `threadpool_destroy_work_item` shall call `WaitForThreadpoolWorkCallbacks` to wait on all outstanding tasks being scheduled on this `ptp_work`. **]**
 
-**SRS_THREADPOOL_WIN32_05_017: [** `threadpool_destroy_work_item` shall free the `work_item_context`. **]**
+**SRS_THREADPOOL_WIN32_05_017: [** `threadpool_destroy_work_item` shall call `CloseThreadpoolWork` to close `ptp_work`. **]**
+
+**SRS_THREADPOOL_WIN32_05_018: [** `threadpool_destroy_work_item` shall free the `work_item_context`. **]**
--- a/win32/devdoc/threadpool_win32.md
+++ b/win32/devdoc/threadpool_win32.md
@@ -56,13 +56,13 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL), threadpool_create, EXECUTION_ENGINE_HAN
 
 **SRS_THREADPOOL_WIN32_01_025: [** `threadpool_create` shall obtain the PTP_POOL from the execution engine by calling `execution_engine_win32_get_threadpool`. **]**
 
-**S_R_S_THREADPOOL_WIN32_01_026: [** `threadpool_create` shall initialize a thread pool environment by calling `InitializeThreadpoolEnvironment`. **]**
+**SRS_THREADPOOL_WIN32_01_026: [** `threadpool_create` shall initialize a thread pool environment by calling `InitializeThreadpoolEnvironment`. **]**
 
-**S_R_S_THREADPOOL_WIN32_01_027: [** `threadpool_create` shall set the thread pool for the environment to the pool obtained from the execution engine by calling `SetThreadpoolCallbackPool`. **]**
+**SRS_THREADPOOL_WIN32_01_027: [** `threadpool_create` shall set the thread pool for the environment to the pool obtained from the execution engine by calling `SetThreadpoolCallbackPool`. **]**
 
-**S_R_S_THREADPOOL_WIN32_01_028: [** `threadpool_create` shall create a threadpool cleanup group by calling `CreateThreadpoolCleanupGroup`. **]**
+**SRS_THREADPOOL_WIN32_01_028: [** `threadpool_create` shall create a threadpool cleanup group by calling `CreateThreadpoolCleanupGroup`. **]**
 
-**S_R_S_THREADPOOL_WIN32_01_029: [** `threadpool_create` shall associate the cleanup group with the just created environment by calling `SetThreadpoolCallbackCleanupGroup`. **]**
+**SRS_THREADPOOL_WIN32_01_029: [** `threadpool_create` shall associate the cleanup group with the just created environment by calling `SetThreadpoolCallbackCleanupGroup`. **]**
 
 **SRS_THREADPOOL_WIN32_01_003: [** If any error occurs, `threadpool_create` shall fail and return `NULL`. **]**
 
@@ -74,19 +74,13 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees all resources associated with threadpool.
 
-**SRS_THREADPOOL_WIN32_01_005: [** `threadpool_dispose` shall free all resources associated with `threadpool`. **]**
-
-**S_R_S_THREADPOOL_WIN32_01_016: [** If `threadpool` is `NULL`, `threadpool_dispose` shall return. **]**
-
-**SRS_THREADPOOL_WIN32_01_006: [** While `threadpool` is OPENING or CLOSING, `threadpool_dispose` shall wait for the open to complete either successfully or with error. **]**
-
 **SRS_THREADPOOL_WIN32_01_007: [** `threadpool_dispose` shall perform an implicit close. **]**
 
-- **S_R_S_THREADPOOL_WIN32_01_030: [** `threadpool_dispose` shall wait for any executing callbacks by calling `CloseThreadpoolCleanupGroupMembers`, passing `FALSE` as `fCancelPendingCallbacks`. **]**
+- **SRS_THREADPOOL_WIN32_01_030: [** `threadpool_dispose` shall wait for any executing callbacks by calling `CloseThreadpoolCleanupGroupMembers`, passing `FALSE` as `fCancelPendingCallbacks`. **]**
 
-- **S_R_S_THREADPOOL_WIN32_01_032: [** `threadpool_dispose` shall close the threadpool cleanup group by calling `CloseThreadpoolCleanupGroup`. **]**
+- **SRS_THREADPOOL_WIN32_01_032: [** `threadpool_dispose` shall close the threadpool cleanup group by calling `CloseThreadpoolCleanupGroup`. **]**
 
-- **S_R_S_THREADPOOL_WIN32_01_033: [** `threadpool_dispose` shall destroy the thread pool environment created in `threadpool_create`. **]**
+- **SRS_THREADPOOL_WIN32_01_033: [** `threadpool_dispose` shall destroy the thread pool environment created in `threadpool_create`. **]**
 
 **SRS_THREADPOOL_WIN32_42_028: [** `threadpool_dispose` shall decrement the reference count on the `execution_engine`. **]**
 
@@ -100,23 +94,7 @@ Note: `threadpool_open` will be deprecated and threadpool_create will perform ad
 
 **SRS_THREADPOOL_WIN32_01_008: [** If `threadpool` is `NULL`, `threadpool_open` shall fail and return a non-zero value. **]**
 
-**SRS_THREADPOOL_WIN32_01_011: [** Otherwise, `threadpool_open` shall switch the state to OPENING. **]**
-
-**SRS_THREADPOOL_WIN32_01_026: [** `threadpool_open` shall initialize a thread pool environment by calling `InitializeThreadpoolEnvironment`. **]**
-
-**SRS_THREADPOOL_WIN32_01_027: [** `threadpool_open` shall set the thread pool for the environment to the pool obtained from the execution engine by calling `SetThreadpoolCallbackPool`. **]**
-
-**SRS_THREADPOOL_WIN32_01_028: [** `threadpool_open` shall create a threadpool cleanup group by calling `CreateThreadpoolCleanupGroup`. **]**
-
-**SRS_THREADPOOL_WIN32_01_029: [** `threadpool_open` shall associate the cleanup group with the just created environment by calling `SetThreadpoolCallbackCleanupGroup`. **]**
-
-**SRS_THREADPOOL_WIN32_01_015: [** `threadpool_open` shall set the state to OPEN. **]**
-
 **SRS_THREADPOOL_WIN32_01_012: [** On success, `threadpool_open` shall return 0. **]**
-
-**SRS_THREADPOOL_WIN32_01_013: [** If `threadpool` is already OPEN or OPENING, `threadpool_open` shall fail and return a non-zero value. **]**
-
-**SRS_THREADPOOL_WIN32_01_040: [** If any error occurrs, `threadpool_open` shall fail and return a non-zero value. **]**
 
 ### threadpool_close
 
@@ -126,17 +104,7 @@ MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 Note: `threadpool_close` will be deprecated and threadpool_dispose will perform additional tasks of `threadpool_close`. This function will exist until all the libraries calling this API are modified to use only `threadpool_create`.
 `threadpool_close` closes an open threadpool.
 
-**SRS_THREADPOOL_WIN32_01_016: [** If `threadpool` is `NULL`, `threadpool_close` shall return. **]**
-
-**SRS_THREADPOOL_WIN32_01_017: [** Otherwise, `threadpool_close` shall switch the state to CLOSING. **]**
-
-**SRS_THREADPOOL_WIN32_01_030: [** `threadpool_close` shall wait for any executing callbacks by calling `CloseThreadpoolCleanupGroupMembers`, passing `FALSE` as `fCancelPendingCallbacks`. **]**
-
-**SRS_THREADPOOL_WIN32_01_032: [** `threadpool_close` shall close the threadpool cleanup group by calling `CloseThreadpoolCleanupGroup`. **]**
-
-**SRS_THREADPOOL_WIN32_01_033: [** `threadpool_close` shall destroy the thread pool environment created in `threadpool_create`. **]**
-
-**SRS_THREADPOOL_WIN32_01_019: [** If `threadpool` is not OPEN, `threadpool_close` shall return. **]**
+**SRS_THREADPOOL_WIN32_05_018: [** If `threadpool` is `NULL`, `threadpool_close` shall return. **]**
 
 ### threadpool_schedule_work
 
--- a/win32/src/threadpool_win32.c
+++ b/win32/src/threadpool_win32.c
@@ -15,15 +15,6 @@
 #include "c_pal/thandle.h"
 #include "c_pal/threadpool.h"
 
-#define THREADPOOL_WIN32_STATE_VALUES \
-    THREADPOOL_WIN32_STATE_CLOSED, \
-    THREADPOOL_WIN32_STATE_OPENING, \
-    THREADPOOL_WIN32_STATE_OPEN, \
-    THREADPOOL_WIN32_STATE_CLOSING
-
-MU_DEFINE_ENUM(THREADPOOL_WIN32_STATE, THREADPOOL_WIN32_STATE_VALUES)
-MU_DEFINE_ENUM_STRINGS(THREADPOOL_WIN32_STATE, THREADPOOL_WIN32_STATE_VALUES)
-
 typedef struct WORK_ITEM_CONTEXT_TAG
 {
     THREADPOOL_WORK_FUNCTION work_function;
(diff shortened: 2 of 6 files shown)


win32/src/threadpool_win32.c
```diff
@@ -182,14 +153,32 @@ THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
             }
             else
             {
-                /* Codes_SRS_THREADPOOL_WIN32_42_027: [ threadpool_create shall increment the reference count on the execution_engine. ]*/
-                execution_engine_inc_ref(execution_engine);
-                result->execution_engine = execution_engine;
+                /* Codes_SRS_THREADPOOL_WIN32_01_026: [ threadpool_open shall initialize a thread pool environment by calling InitializeThreadpoolEnvironment. ]*/
```
dcristoloveanu, line 156: >threadpool_open [](http://example.com/codeflow?start=56&length=15)

Meh, threadpool_open does not exist anymore.

win32/src/threadpool_win32.c
```diff
@@ -182,14 +153,32 @@ THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
...
             {
-                /* Codes_SRS_THREADPOOL_WIN32_42_027: [ threadpool_create shall increment the reference count on the execution_engine. ]*/
-                execution_engine_inc_ref(execution_engine);
-                result->execution_engine = execution_engine;
+                /* Codes_SRS_THREADPOOL_WIN32_01_026: [ threadpool_open shall initialize a thread pool environment by calling InitializeThreadpoolEnvironment. ]*/
+                InitializeThreadpoolEnvironment(&result->tp_environment);
 
-                (void)InterlockedExchange(&result->pending_api_calls, 0);
-                (void)InterlockedExchange(&result->state, THREADPOOL_WIN32_STATE_CLOSED);
+                /* Codes_SRS_THREADPOOL_WIN32_01_027: [ threadpool_open shall set the thread pool for the environment to the pool obtained from the execution engine by calling SetThreadpoolCallbackPool. ]*/
```
dcristoloveanu, line 159: >threadpool_open [](http://example.com/codeflow?start=56&length=15)

Here also. And below.

dcristoloveanu:
>             result->pool = execution_engine_win32_get_threadpool(execution_engine);

Pretty sure the pool is leaked.

---
Refers to: win32/src/threadpool_win32.c:148 in 257fdd0. [](commit_id = 257fdd0250e1e54d731b994851313a2c480d4e2d, deletion_comment = False)

win32/src/threadpool_win32.c
```diff
@@ -213,81 +201,19 @@ int threadpool_open(THANDLE(THREADPOOL) threadpool)
...
 }
 
 void threadpool_close(THANDLE(THREADPOOL) threadpool)
 {
     if (threadpool == NULL)
     {
-        /* Codes_SRS_THREADPOOL_WIN32_01_016: [ If threadpool is NULL, threadpool_close shall return. ]*/
+        /* Codes_SRS_THREADPOOL_WIN32_05_018: [ If threadpool is NULL, threadpool_close shall return. ]*/
         LogError("Invalid arguments: THREADPOOL_HANDLE threadpool=%p", threadpool);
     }
```
dcristoloveanu, line 216: >} [](http://example.com/codeflow?start=4&length=1)

Consider adding an else branch with a comment of "do_nothing"

win32/src/threadpool_win32.c
```diff
@@ -367,18 +286,9 @@ int threadpool_schedule_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WOR
...
-        else
-        {
-            /* Codes_SRS_THREADPOOL_WIN32_05_013: [ threadpool_schedule_work_item shall call SubmitThreadpoolWork to submit the work item for execution. ]*/
-            SubmitThreadpoolWork(work_item_context->ptp_work);
-            result = 0;
-        }
+        /* Codes_SRS_THREADPOOL_WIN32_05_013: [ threadpool_schedule_work_item shall call SubmitThreadpoolWork to submit the work item for execution. ]*/
+        SubmitThreadpoolWork(work_item_context->ptp_work);
+        result = 0;
         (void)InterlockedDecrement(&threadpool_ptr->pending_api_calls);
```
dcristoloveanu, line 292: We usually have this be the last thing we do (between ln 293 and 294).

That's easier to spot usually than if it is in the middle of other code. /consider.

win32/src/threadpool_win32.c
```diff
@@ -401,17 +311,8 @@ void threadpool_destroy_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WOR
...
-        if (state != THREADPOOL_WIN32_STATE_OPEN)
-        {
-            LogWarning("Bad state: %" PRI_MU_ENUM, MU_ENUM_VALUE(THREADPOOL_WIN32_STATE, state));
-        }
-        else
-        {
-            /* Codes_SRS_THREADPOOL_WIN32_05_016: [ threadpool_destroy_work_item shall call CloseThreadpoolWork to close ptp_work. ]*/
-            CloseThreadpoolWork(work_item_context->ptp_work);
-        }
+        /* Codes_SRS_THREADPOOL_WIN32_05_016: [ threadpool_destroy_work_item shall call CloseThreadpoolWork to close ptp_work. ]*/
```
dcristoloveanu, line 314: I fail to see why do we have pending_api_calls if there is no open/close state any longer.
I think we only wait for pending call count to be 0 in the dispose.
And if we are in the dispose, everyone has relinquished all the references.

So I think this pending_api_calls has to die.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -36,6 +36,8 @@ typedef struct WAIT_WORK_CONTEXT_TAG
     HANDLE wait_event;
 } WAIT_WORK_CONTEXT;
 
+#define WAIT_WORK_FUNCTION_SLEEP 3000
```
dcristoloveanu, line 39: >3000 [](http://example.com/codeflow?start=33&length=4)

Consider adding a comment next to this saying // ms
so that we know what we're measuring time in.
Or even better have it in the name of the define: WAIT_WORK_FUNCTION_SLEEP_IN_MS

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -67,7 +69,7 @@ typedef struct OPEN_WORK_CONTEXT_TAG
 static void open_work_function(void* context)
 {
     OPEN_WORK_CONTEXT* open_work_context = (OPEN_WORK_CONTEXT*)context;
-    ASSERT_ARE_NOT_EQUAL(int, 0, threadpool_open(open_work_context->threadpool));
+    ASSERT_ARE_EQUAL(int, 0, threadpool_open(open_work_context->threadpool));
     (void)InterlockedIncrement(&open_work_context->call_count);
```
dcristoloveanu, line 73: What are we testing here with threadpool_open?
This should be completely removed from the threadpool tests with this PR.
There is no value in having threadpool_open and threadpool_close tested in threadpool_ut and threadpool_int.

Removing from all the other users will come with other PR(s), but after this PR they should not be tested anymore (as they do nothing).

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -816,6 +821,20 @@ typedef struct CHAOS_TEST_DATA_TAG
 #define TIMER_PERIOD_MIN 50
 #define TIMER_PERIOD_MAX 400
 
+static void chaos_delay()
+{
+    // Earlier, the threadpool_open and threadpool_close would start and stop threadpool threads whereby there would be some limit on number of times work would be scheduled
+    // on Windows Threadpool using threadpool_schedule_work and threadpool_schedule_work_item because both functions would test for threadpool open condition and return if threadpool is not open.
+    // With threadpool_open and threadpool_close now only there for backward compatibility and doing nothing, the chaos test for optimized threadpool_schedule_work_item
+    // would create a very large number of work items compared to the previous version of threadpool_schedule_work. The tests would fail because expected_call_count would be very high than the
+    // executed_work_functions count because Windows Threadpool is unable to schedule that many work items. Adding Sleep(1) would significantly slow the tests. So adding following malloc and free
```
dcristoloveanu, line 830: >Windows Threadpool is unable to schedule that many work items [](http://example.com/codeflow?start=45&length=61)

This is interesting. Is there a specific limit here that we can reference?

win32/tests/threadpool_win32_ut/threadpool_win32_ut.c
```diff
@@ -300,16 +297,29 @@ TEST_FUNCTION(threadpool_create_with_NULL_execution_engine_fails)
 }
 
 /* Tests_SRS_THREADPOOL_WIN32_01_001: [ threadpool_create shall allocate a new threadpool object and on success shall return a non-NULL handle. ]*/
-/* Tests_SRS_THREADPOOL_WIN32_42_027: [ threadpool_create shall increment the reference count on the execution_engine. ]*/
 /* Tests_SRS_THREADPOOL_WIN32_01_025: [ threadpool_create shall obtain the PTP_POOL from the execution engine by calling execution_engine_win32_get_threadpool. ]*/
+/* Tests_SRS_THREADPOOL_WIN32_01_026: [ threadpool_open shall initialize a thread pool environment by calling InitializeThreadpoolEnvironment. ]*/
```
dcristoloveanu, line 301: >threadpool_open [](http://example.com/codeflow?start=40&length=15)

In the test also: threadpool_open is does not exist any longer for all practical purposes

win32/tests/threadpool_win32_ut/threadpool_win32_ut.c
```diff
@@ -375,23 +393,28 @@ TEST_FUNCTION(threadpool_dispose_frees_resources)
...
         .CaptureArgumentValue_pcbe(&cbe);
     STRICT_EXPECTED_CALL(mocked_SetThreadpoolCallbackPool(IGNORED_ARG, test_pool))
         .ValidateArgumentValue_pcbe(&cbe);
     STRICT_EXPECTED_CALL(mocked_CreateThreadpoolCleanupGroup())
         .CaptureReturn(&test_cleanup_group);
-    (void)threadpool_open(threadpool);
+    STRICT_EXPECTED_CALL(mocked_SetThreadpoolCallbackCleanupGroup(IGNORED_ARG, IGNORED_ARG, IGNORED_ARG))
+        .ValidateArgumentValue_ptpcg(&test_cleanup_group);
+    STRICT_EXPECTED_CALL(execution_engine_inc_ref(test_execution_engine));
+    THANDLE(THREADPOOL) threadpool = threadpool_create(test_execution_engine);
```
dcristoloveanu, line 417: >threadpool [](http://example.com/codeflow?start=24&length=10)

There should be an assert is not null on this.

dcristoloveanu (changes requested):
:clock1:

nishudeshu:
> static void internal_close(THREADPOOL* threadpool)

Done.

---
In reply to: [2467156160](https://github.com/Azure/c-pal/pull/408#issuecomment-2467156160) [](http://example.com/codeflow?ancestors=2467156160)

---
Refers to: win32/src/threadpool_win32.c:96 in 257fdd0. [](commit_id = 257fdd0250e1e54d731b994851313a2c480d4e2d, deletion_comment = False)

nishudeshu:
>             result->pool = execution_engine_win32_get_threadpool(execution_engine);

`pool` is freed in `execution_engine_dec_ref`

---
In reply to: [2467159316](https://github.com/Azure/c-pal/pull/408#issuecomment-2467159316) [](http://example.com/codeflow?ancestors=2467159316)

---
Refers to: win32/src/threadpool_win32.c:148 in 257fdd0. [](commit_id = 257fdd0250e1e54d731b994851313a2c480d4e2d, deletion_comment = False)

win32/src/threadpool_win32.c
nishudeshu: Changed.

win32/src/threadpool_win32.c
nishudeshu: Changed

win32/src/threadpool_win32.c
nishudeshu: Done

win32/src/threadpool_win32.c
nishudeshu: It is matching the Increment on line 288. I am unsure how much below it can go.

win32/src/threadpool_win32.c
nishudeshu: `pending_api_calls` is incremented by 1 just  before `SubmitThreadpoolWork` on line 290 which may happen in a different thread than where `threadpool_dispose` is happening. This ensures that perhaps the last few jobs to  `threadpool_schedule_work_item` is submitted before allow dispose to continue and destroy the `threadpool`. Not full proof but some kind of protection before dispose closes everything. Nothing to do with open and close.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Renamed to `WAIT_WORK_FUNCTION_SLEEP_IN_MS`

win32/tests/threadpool_win32_ut/threadpool_win32_ut.c
nishudeshu: Done.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -816,6 +821,20 @@ typedef struct CHAOS_TEST_DATA_TAG
...
+    // would ensure delay less than 1 milliseconds, enough to allow Windows Threadpool to handle this chaos test scenarios for threadpool_schedule_work_item. For threadpool_schedule_work there are
+    // already extra operations than threadpool_schedule_work_item like malloc and free (in its callback) that would create more delay than this function. Hence threadpool_schedule_work would pass in
```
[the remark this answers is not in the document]
nishudeshu: Removed this function for InterlockedHL Wait

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Removed this function for InterlockedHL Wait

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: In UT, only the open and close function that remain is tested to satisfy Traceability tool.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -685,9 +625,9 @@ TEST_FUNCTION(close_while_items_are_scheduled_still_executes_all_items)
     ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work(threadpool, wait_work_function, (void*)&wait_work_context));
     ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work(threadpool, work_function, (void*)&wait_work_context.call_count));
 
+    Sleep(WAIT_WORK_FUNCTION_SLEEP_IN_MS);
```
dcristoloveanu, line 628: Why is this Sleep needed? What happens if you remove it?

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -837,24 +679,17 @@ static DWORD WINAPI chaos_thread_func(LPVOID lpThreadParameter)
...
-                break;
-            case 2:
                 // perform a schedule item
                 if (threadpool_schedule_work(chaos_test_data->threadpool, work_function, (void*)&chaos_test_data->executed_work_functions) == 0)
                 {
                     (void)InterlockedIncrement64(&chaos_test_data->expected_call_count);
                 }
-            case 3:
+                break;
+            case 1:
```
dcristoloveanu, line 692: >1 [](http://example.com/codeflow?start=17&length=1)

I think this would be nicer if modeled by an ENUM (like we have below for TEST_ACTION_CLEANUP_TIMER, etc.

This is a consider being that the code was already like that.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1015,7 +844,6 @@ static DWORD WINAPI chaos_thread_with_timers_no_lock_func(LPVOID lpThreadParamet
                     (void)InterlockedIncrement64(&chaos_test_data->expected_call_count);
                 }
             }
-            break;
```
dcristoloveanu, line 1018: >break; [](http://example.com/codeflow?start=12&length=6)

We should leave this break; in, generally each case: should have its own, break. No need to remove it.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1067,14 +895,14 @@ TEST_FUNCTION(chaos_knight_test)
     }
 
     // assert that all scheduled items were executed
-    ASSERT_ARE_EQUAL(int64_t, (int64_t)InterlockedAdd64(&chaos_test_data.expected_call_count, 0), (int64_t)InterlockedAdd64(&chaos_test_data.executed_work_functions, 0));
+
+    InterlockedHL_WaitForValue64(&chaos_test_data.executed_work_functions, chaos_test_data.expected_call_count, UINT32_MAX);
```
dcristoloveanu, line 899: >InterlockedHL_WaitForValue64 [](http://example.com/codeflow?start=4&length=28)

We should ASSERT on the result of this, right?

dcristoloveanu, line 899: >InterlockedHL_WaitForValue64 [](http://example.com/codeflow?start=4&length=28)

Yay for using the new API!

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1137,15 +964,14 @@ XTEST_FUNCTION(chaos_knight_test_with_timers_no_lock)
     }
 
     // assert that all scheduled items were executed
-    ASSERT_ARE_EQUAL(int64_t, (int64_t)InterlockedAdd64(&chaos_test_data.expected_call_count, 0), (int64_t)InterlockedAdd64(&chaos_test_data.executed_work_functions, 0));
+    InterlockedHL_WaitForValue64(&chaos_test_data.executed_work_functions, chaos_test_data.expected_call_count, UINT32_MAX);
```
dcristoloveanu, line 967: >InterlockedHL_WaitForValue64 [](http://example.com/codeflow?start=4&length=28)

Assert missing here also.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
```diff
@@ -1297,9 +1108,9 @@ TEST_FUNCTION(close_while_items_are_scheduled_still_executes_all_items_v2)
     ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work_item(threadpool, wait_work_item_context));
     ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work_item(threadpool, work_item_context));
 
+    Sleep(WAIT_WORK_FUNCTION_SLEEP_IN_MS);
```
dcristoloveanu, line 1111: >Sleep(WAIT_WORK_FUNCTION_SLEEP_IN_MS); [](http://example.com/codeflow?start=4&length=38)

Again, I don't think this Sleep is needed, can we remove it please?

win32/tests/threadpool_win32_ut/threadpool_win32_ut.c
```diff
@@ -375,23 +397,29 @@ TEST_FUNCTION(threadpool_dispose_frees_resources)
     ASSERT_ARE_EQUAL(char_ptr, umock_c_get_expected_calls(), umock_c_get_actual_calls());
 }
 
-/* Tests_SRS_THREADPOOL_WIN32_01_006: [ While threadpool is OPENING or CLOSING, threadpool_dispose shall wait for the open to complete either successfully or with error. ]*/
 /* Tests_SRS_THREADPOOL_WIN32_01_007: [ threadpool_dispose shall perform an implicit close if threadpool is OPEN. ]*/
+/* Tests_SRS_THREADPOOL_WIN32_01_030: [ threadpool_close shall wait for any executing callbacks by calling CloseThreadpoolCleanupGroupMembers, passing FALSE as fCancelPendingCallbacks. ]*/
```
dcristoloveanu, line 401: >threadpool_close [](http://example.com/codeflow?start=40&length=16)

This cannot be threadpool_close.

dcristoloveanu:
>     STRICT_EXPECTED_CALL(interlocked_decrement(IGNORED_ARG));

Sounds like this comment should be // destroy calls now.

---
Refers to: win32/tests/threadpool_win32_ut/threadpool_win32_ut.c:426 in cd9c0b2. [](commit_id = cd9c0b2945ab6c0e434767c836babc6f506e8bed, deletion_comment = False)

win32/devdoc/threadpool_win32.md
```diff
@@ -74,19 +74,13 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees all resources associated with threadpool.
 
-**SRS_THREADPOOL_WIN32_01_005: [** `threadpool_dispose` shall free all resources associated with `threadpool`. **]**
-
-**S_R_S_THREADPOOL_WIN32_01_016: [** If `threadpool` is `NULL`, `threadpool_dispose` shall return. **]**
-
-**SRS_THREADPOOL_WIN32_01_006: [** While `threadpool` is OPENING or CLOSING, `threadpool_dispose` shall wait for the open to complete either successfully or with error. **]**
-
 **SRS_THREADPOOL_WIN32_01_007: [** `threadpool_dispose` shall perform an implicit close. **]**
```
dcristoloveanu, line 77: This spec is obsolete now that close/open are not used any longer, so the spec should be removed. From specs/tests and code please.

dcristoloveanu (dismissed):
:clock1:

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: It is needed for the wait_work_function. The sleep prevent immediate execution of `SetEvent(wait_work_context.wait_event)` on line 633

nishudeshu: The wait_work_function timeout is 2000 ms and it needs to timeout for the test to pass.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: After the `threadpool_schedule_work` would be deprecated, this test will go away or will have only `threadpool_schedule_work_item` in which case the switch case block won't be needed.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Fixed.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Fixed.

win32/tests/threadpool_win32_int/threadpool_win32_int.c
nishudeshu: Fixed.

win32/tests/threadpool_win32_ut/threadpool_win32_ut.c
nishudeshu: Resolved.

win32/devdoc/threadpool_win32.md
nishudeshu: Okay, it said implicit close and the next 3 lines are subsection of this section. The close is happening but not in threadpool_close.

nishudeshu:
>     STRICT_EXPECTED_CALL(interlocked_decrement(IGNORED_ARG));

Fixed.

---
In reply to: [2486935318](https://github.com/Azure/c-pal/pull/408#issuecomment-2486935318) [](http://example.com/codeflow?ancestors=2486935318)

---
Refers to: win32/tests/threadpool_win32_ut/threadpool_win32_ut.c:426 in cd9c0b2. [](commit_id = cd9c0b2945ab6c0e434767c836babc6f506e8bed, deletion_comment = False)

win32/devdoc/threadpool_win32.md
dcristoloveanu: The change is to not have an open/close state any longer. In other words to help understand my thought process: what does "close" mean if we do not have an open/closed state for the module?

dcristoloveanu (changes requested):
:clock1:

win32/src/threadpool_win32.c
dcristoloveanu: The comment in iteration 4 refers to " result = 0;" not to the InterlockedDecrement.

win32/src/threadpool_win32.c
dcristoloveanu: I beg to differ. Knowing the history from the first line of the code being written in the project, I can 100% say that pending_api_calls exists as a reference count so that we do not switch the state to closed while the count is greater than 0.

Can you prove your statement that you can call SubmitThreadpoolWork from one thread and dispose is being executed on another thread?

win32/devdoc/threadpool_win32.md
nishudeshu: We have `CloseThreadpoolCleanupGroupMembers` and `CloseThreadpoolCleanupGroup` which essentially means we are closing the threadpool related items in this section

win32/src/threadpool_win32.c
nishudeshu: Okay, but in the previous code, the result = 0 happened before the InterlockedDecrement. I am not sure how does it matter?

win32/src/threadpool_win32.c
nishudeshu: Okay, I see what you are saying. But what is the harm in keeping them? Wouldn't it be useful if we want to bring back some feature like that in the future? Would it be helpful during debugs to see how many API calls were pending when some crash happened?

win32/src/threadpool_win32.c
dcristoloveanu: Fair enough, let me explain why it matters: setting the result in the middle of the code of a function and then doing more things is prone to being overlooked when reviewing. Thus as a convention for our team we set the result as the last thing (unless there is a reason not to do that).

This is not such a case, there is no functional reason why it has to be done before the decrement.
So, it should be as the last thing we do, as everyone in the team expects it.

A question for you is: what technical reason is there to have the assignment of result = 0 before the decrement? How would that help?

win32/devdoc/threadpool_win32.md
dcristoloveanu: Does the threadpool have an open state vs a closed state any longer? I don't think it does. If it would have a closed/opened state it would be modeled in a state variable (either owned by the threadpool object or delegated to an SM object).

The spec to perform an implicit close in the destroy is written for objects that have an open/closed state.

HTH.

win32/src/threadpool_win32.c
nishudeshu: Thanks for the clarification, I have updated the code.

win32/src/threadpool_win32.c
nishudeshu: Thanks for the clarification, I have updated the code.

win32/devdoc/threadpool_win32.md
nishudeshu: Removed the specs

dcristoloveanu (approved):
:shipit:

PR #409, opened 2024-11-12
c-pal code changes for gballoc_hl/ll_set_option

RohitG28:


win32/devdoc/gballoc_ll_jemalloc.md
```diff
@@ -165,4 +167,28 @@ static void jemalloc_print_stats_callback(void* context, const char* text)
...
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
+```
+
+`gballoc_ll_set_decay` sets the dirty and muzzy pages decay time to `decay_milliseconds` milliseconds.
+
+**SRS_GBALLOC_LL_JEMALLOC_28_001: [** `gballoc_ll_set_decay` shall advance jemalloc's internal epoch counter. **]**
```
anporumb, line 180: > shall advance jemalloc's internal epoch counter [](http://example.com/codeflow?start=60&length=48)

how does it do that?

My expectation here requirement-wise would be that an API is mentioned by name with enough information so that unit tests are craftable just by reading the spec.

anporumb: Basically this goes for all requirements written here as I'd have no idea what jemalloc API they call.

win32/devdoc/gballoc_ll_jemalloc.md
```diff
@@ -165,4 +167,28 @@ static void jemalloc_print_stats_callback(void* context, const char* text)
...
+
+**SRS_GBALLOC_LL_JEMALLOC_28_004: [** `gballoc_ll_set_decay` shall fetch the number of existing jemalloc arenas. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_005: [** For each existing arena: **]**
+
+- **SRS_GBALLOC_LL_JEMALLOC_28_006: [** `gballoc_ll_set_decay` shall set the dirty decay time for the arena to `decay_milliseconds` milliseconds. **]**
+
+- **SRS_GBALLOC_LL_JEMALLOC_28_007: [** `gballoc_ll_set_decay` shall set the muzzy decay time for the arena to `decay_milliseconds` milliseconds. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_008: [** If there are any errors, `gballoc_ll_set_decay` shall fail and return a non-zero value. **]**
```
anporumb, line 194: >If there are any errors [](http://example.com/codeflow?start=38&length=23)

This is a bit weird here, seeing as there's a "for each" in line 188 of this .md

What happens if there are 3 arenas... and... the 2nd one fails?

Will the third arena have its dirty/muzzy decay timer adjusted? Will it be skipped? Seeing as we'd already know the return value from the function ("error")

win32/devdoc/gballoc_ll_jemalloc.md
```diff
@@ -165,4 +167,28 @@ static void jemalloc_print_stats_callback(void* context, const char* text)
...
\ No newline at end of file
+**SRS_GBALLOC_LL_JEMALLOC_01_010: [** Otherwise, `jemalloc_print_stats_callback` shall print (log) `text`, breaking it does in chunks of `LOG_MAX_MESSAGE_LENGTH / 2`. **]**
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
+```
+
+`gballoc_ll_set_decay` sets the dirty and muzzy pages decay time to `decay_milliseconds` milliseconds.
```
anporumb, line 178: >sets the dirty and muzzy pages decay time to `decay_milliseconds` milliseconds. [](http://example.com/codeflow?start=23&length=79)

One would argue that decay_milliseconds should be 
1) part of "runtime" API (and that happens here)
2) part of "init" API (gballoc_ll_init takes a void* and this value should be behind that void*).

Because what happens with all the allocates between

gballoc_ll_init and
gballoc_ll_set_decay?

If `decay_milliseconds` exists at the time of gballoc_ll_init then it should be used by gballoc_ll_init.

win32/devdoc/gballoc_ll_mimalloc.md
```diff
@@ -152,3 +154,13 @@ MOCKABLE_FUNCTION(, void, gballoc_ll_print_stats);
...
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
+```
+
+`gballoc_ll_set_decay` does nothing and returns a non-zero value.
+
+**SRS_GBALLOC_LL_MIMALLOC_28_001: [** `gballoc_ll_set_decay` shall do nothing and return a non-zero value. **]**
```
anporumb, line 166: >return a non-zero value. [](http://example.com/codeflow?start=82&length=24)

I think it should succeed.

win32/devdoc/gballoc_ll_passthrough.md
```diff
@@ -150,3 +152,13 @@ MOCKABLE_FUNCTION(, void, gballoc_ll_print_stats);
...
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
+```
+
+`gballoc_ll_set_decay` does nothing and returns a non-zero value.
+
+**SRS_GBALLOC_LL_PASSTHROUGH_28_001: [** `gballoc_ll_set_decay` shall do nothing and return a non-zero value. **]**
```
anporumb, line 164: >return a non-zero value. [](http://example.com/codeflow?start=85&length=24)

I also think it should succeed.

win32/devdoc/gballoc_ll_win32heap.md
```diff
@@ -207,3 +209,13 @@ MOCKABLE_FUNCTION(, void, gballoc_ll_print_stats);
...
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
+```
+
+`gballoc_ll_set_decay` does nothing and returns a non-zero value.
+
+**SRS_GBALLOC_LL_WIN32HEAP_28_001: [** `gballoc_ll_set_decay` shall do nothing and return a non-zero value. **]**
```
anporumb, line 221: >and return a non-zero value. [](http://example.com/codeflow?start=79&length=28)

I think it should succeed.

In the event where win32heap is used - I think that if this fails - the whole program might fail for no good reason.

win32/devdoc/gballoc_ll_jemalloc.md
```diff
@@ -165,4 +167,28 @@ static void jemalloc_print_stats_callback(void* context, const char* text)
...
 **SRS_GBALLOC_LL_JEMALLOC_01_009: [** If `text` is NULL, `jemalloc_print_stats_callback` shall return. **]**
 
-**SRS_GBALLOC_LL_JEMALLOC_01_010: [** Otherwise, `jemalloc_print_stats_callback` shall print (log) `text`, breaking it does in chunks of `LOG_MAX_MESSAGE_LENGTH / 2`. **]**
\ No newline at end of file
+**SRS_GBALLOC_LL_JEMALLOC_01_010: [** Otherwise, `jemalloc_print_stats_callback` shall print (log) `text`, breaking it does in chunks of `LOG_MAX_MESSAGE_LENGTH / 2`. **]**
+
+### gballoc_ll_set_decay
+
+```c
+MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
```
anporumb, line 175: >decay_milliseconds [](http://example.com/codeflow?start=56&length=18)

Is -15 a valid argument value here?

win32/tests/gballoc_ll_jemalloc_ut/gballoc_ll_jemalloc_ut.c
```diff
@@ -628,4 +633,100 @@ TEST_FUNCTION(gballoc_ll_print_stats_calls_the_print_function_with_NULL_does_not
...
+    STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, NULL, 0))
+        .CaptureArgumentValue_name(&epoch)
+        .SetReturn(0);
+    STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, IGNORED_ARG, IGNORED_ARG))
+        .CaptureArgumentValue_name(&dirty_decay_ms)
+        .SetReturn(0);
+    STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, IGNORED_ARG, IGNORED_ARG))
+        .CaptureArgumentValue_name(&muzzy_decay_ms)
+        .SetReturn(0);
+    STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG))
```
anporumb, line 664: >IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG, IGNORED_ARG [](http://example.com/codeflow?start=41&length=63)

:(

can we do better in terms of not ignoring arguments?

win32/tests/gballoc_ll_jemalloc_ut/gballoc_ll_jemalloc_ut.c
```diff
@@ -628,4 +633,100 @@ TEST_FUNCTION(gballoc_ll_print_stats_calls_the_print_function_with_NULL_does_not
...
+    for (uint64_t i = 0; i < NUM_ARENAS; i++)
+    {
+        STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, IGNORED_ARG, IGNORED_ARG))
+            .SetReturn(0);
+        STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, IGNORED_ARG, IGNORED_ARG))
+            .SetReturn(0);
+    }
+
+    ///act
+    int result = gballoc_ll_set_decay(42);
```
anporumb, line 677: >42 [](http://example.com/codeflow?start=38&length=2)

this "42" here should be asserted that it is passed "somewhere" in the mallctl function calls.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
     /* Codes_SRS_GBALLOC_LL_JEMALLOC_01_008: [ gballoc_ll_print_stats shall call je_malloc_stats_print and pass to it jemalloc_print_stats_callback as print callback. ]*/
     je_malloc_stats_print(jemalloc_print_stats_callback, NULL, NULL);
 }
+
+int gballoc_ll_set_decay(int64_t decay_milliseconds)
+{
+    int result = 0;
+
+    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_001: [ gballoc_ll_set_decay shall advance jemalloc's internal epoch counter. ]*/
+    if (je_mallctl("epoch", NULL, NULL, NULL, 0) != 0)
```
anporumb, line 227: >je_mallctl("epoch", NULL, NULL, NULL, 0) [](http://example.com/codeflow?start=8&length=40)

errr... why do we need to set the epoch?

I've been reading the documentation and it seems to be that the epoch is used for reporting statistics.

`Take special note of the epoch mallctl, which controls refreshing of cached dynamic statistics.`

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+    {
+        /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+        LogError("failed to advance jemalloc epoch");
+        result = MU_FAILURE;
+    }
+    else
+    {
+        /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_002: [ gballoc_ll_set_decay shall set the dirty decay time for new arenas to decay_milliseconds milliseconds. ]*/
+        if (je_mallctl("arenas.dirty_decay_ms", NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
```
anporumb, line 236: >NULL, NULL [](http://example.com/codeflow?start=48&length=10)

would be nice to pass in here some non-`NULL` value so that the existing value is extracted from jemalloc.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+        {
+            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+            LogError("failed to set dirty decay time");
```
anporumb, line 239: >LogError("failed to set dirty decay time"); [](http://example.com/codeflow?start=12&length=43)

Should log all arguments with their values and if the previous value for `arenas.dirty_decay_ms` has been obtained from the call to `je_mallctl` then that should be logged too.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+            result = MU_FAILURE;
+        }
+        else
+        {
+            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_003: [ gballoc_ll_set_decay shall set the muzzy decay time for new arenas to decay_milliseconds milliseconds. ]*/
+            if (je_mallctl("arenas.muzzy_decay_ms", NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
```
anporumb, line 245: >NULL, NULL, [](http://example.com/codeflow?start=52&length=11)

pass here a non-`NULL` value which can be used to grab the existing set value for muzzy_decay_ms

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+            {
+                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                LogError("failed to set muzzy decay time");
```
anporumb, line 248: >LogError("failed to set muzzy decay time");  [](http://example.com/codeflow?start=16&length=44)

log all arguments with their values.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                LogError("failed to set muzzy decay time"); 
+                result = MU_FAILURE; 
+            }
+            else
+            {
+                uint64_t narenas;
+
+                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_004: [ gballoc_ll_set_decay shall fetch the number of existing jemalloc arenas. ]*/
+                if (je_mallctl("arenas.narenas", &narenas, &(size_t){sizeof(narenas)}, NULL, 0) != 0)
```
anporumb, line 256: >size_t [](http://example.com/codeflow?start=61&length=6)

what happens if you cast to `ssize_t` here?

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                {
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                    LogError("failed to get number of arenas");
```
anporumb, line 259: >LogError("failed to get number of arenas"); [](http://example.com/codeflow?start=20&length=43)

should log all arguments with their values.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+            result = MU_FAILURE;
+        }
+        else
+        {
+            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_003: [ gballoc_ll_set_decay shall set the muzzy decay time for new arenas to decay_milliseconds milliseconds. ]*/
+            if (je_mallctl("arenas.muzzy_decay_ms", NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+            {
+                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                LogError("failed to set muzzy decay time"); 
+                result = MU_FAILURE;
```
anporumb, line 249: >result [](http://example.com/codeflow?start=16&length=6)

assume line 249 executes.

Where are the changes done by line 236 undone?

This goes for all the changing APIs in the code here.

Leaving the module with "some" changes performed and "some" not performed is not ideal. 

I would recommend undo-ing (with a best effort approach, obviously) all the changes done to the config of jemalloc in the presence of failures along the way.

win32/src/gballoc_ll_jemalloc.c
anporumb: hint: somewhere the code shoudl assert at least that the sizeof(ssize_t) is the same as the sizeof(int64_t).

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_003: [ gballoc_ll_set_decay shall set the muzzy decay time for new arenas to decay_milliseconds milliseconds. ]*/
+            if (je_mallctl("arenas.muzzy_decay_ms", NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+            {
+                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                LogError("failed to set muzzy decay time"); 
+                result = MU_FAILURE; 
+            }
+            else
+            {
+                uint64_t narenas;
```
anporumb, line 253: >narenas [](http://example.com/codeflow?start=25&length=7)

according to documentation, "narenas" is "unsigned".

Code shoudl assert that sizeof(unsigned) is the same as sizeof("uint64_t).

Because what if they aren't.

Documentation here: `[https://jemalloc.net/jemalloc.3.html`](https://jemalloc.net/jemalloc.3.html`)

win32/src/gballoc_ll_jemalloc.c
anporumb, line 259: >LogError("failed to get number of arenas"); [](http://example.com/codeflow?start=20&length=43)

should log all arguments.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                    LogError("failed to get number of arenas");
+                    result = MU_FAILURE;
+                }
+                else
+                {
+                    char command[64];
+
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ For each existing arena: ]*/
+                    for (uint64_t i = 0; i < narenas; i++) {
```
anporumb, line 267: >{ [](http://example.com/codeflow?start=59&length=1)

keep `{` on the separate line

interfaces/inc/c_pal/gballoc_hl.h
```diff
@@ -63,6 +63,8 @@ extern "C" {
 
     MOCKABLE_FUNCTION(, void, gballoc_hl_print_stats);
 
+    MOCKABLE_FUNCTION(, int, gballoc_hl_set_decay, int64_t, decay_milliseconds);
```
dcristoloveanu, line 66: >gballoc_hl_set_decay [](http://example.com/codeflow?start=29&length=20)

I think I would like more a set_option API. I know it would be an API over an API (or hiding API calls into an API call).
This is probably one case where that would help, like in the case of IOCTL.

If we have a set_option, the generic API of gballoc does not need to change every time we add such an option. Also all the implementations do not need to change, just the one that handles the option.

interfaces/inc/c_pal/gballoc_ll.h
```diff
@@ -32,6 +32,8 @@ extern "C"
 
     MOCKABLE_FUNCTION(, void, gballoc_ll_print_stats);
 
+    MOCKABLE_FUNCTION(, int, gballoc_ll_set_decay, int64_t, decay_milliseconds);
```
dcristoloveanu, line 35: >decay [](http://example.com/codeflow?start=44&length=5)

decay is mushing together dirty and muzzy decay.
I would like it more if we'd keep both options and let the jemalloc implementation handle each one of them.

See my other comment about a "set_option".

win32/src/gballoc_ll_jemalloc.c
dcristoloveanu, line 236: >dirty_decay_ms [](http://example.com/codeflow?start=31&length=14)

As an uber comment, I really think we should not mush together options of jemalloc and we should handle 2 options.
What if next release we determine that it is better to turn off dirty and keep muzzy - just taking the example to an extreme.

Having these 2 options coalesced under one would not help.

win32/src/gballoc_ll_jemalloc.c
dcristoloveanu, line 227: >epoch [](http://example.com/codeflow?start=20&length=5)

Not sure why we need to mess with the epoch? I think maybe stating in the spec a little bit of the "why" would help.

Like "gballoc_ll_set_decay shall advance jemalloc's internal epoch counter in order to ... blah"

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                LogError("snprintf failed for arena %" PRIu64 "", i);
+                                result = MU_FAILURE;
+                            }
+                            else
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_007: [ gballoc_ll_set_decay shall set the muzzy decay time for the arena to decay_milliseconds milliseconds. ]*/
+                                if (je_mallctl(command, NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+                                {
```
dcristoloveanu, line 299: Seeing that the epoch is not used anywhere after it is "advanced", I am wondering why we need it? Some links to some jemalloc docs would also help probably.

dcristoloveanu (changes requested):
:clock1:

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                    char command[64];
+
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ For each existing arena: ]*/
+                    for (uint64_t i = 0; i < narenas; i++) {
+                        int snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu64 ".dirty_decay_ms", i);
+
+                        if (snprintf_result < 0)
+                        {
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                            LogError("snprintf failed for arena %" PRIu64 "", i);
```
anporumb, line 273: >LogError [](http://example.com/codeflow?start=28&length=8)

should log the complete command that failed.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                }
+                else
+                {
+                    char command[64];
+
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ For each existing arena: ]*/
+                    for (uint64_t i = 0; i < narenas; i++) {
+                        int snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu64 ".dirty_decay_ms", i);
+
+                        if (snprintf_result < 0)
```
anporumb, line 270: >snprintf_result < 0 [](http://example.com/codeflow?start=28&length=19)

What happens is line 264 is not so generous with the number of bytes and uses... hmmm.. 10 instead of 64?

C17 standard says `The snprintf function returns the number of characters that would have been written had n been
 sufficiently large, not counting the terminating null character, or a negative value if an encoding
 error occurred. Thus, the null-terminated output has been completely written if and only if the
 returned value is nonnegative and less than n.`

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            LogError("snprintf failed for arena %" PRIu64 "", i);
+                            result = MU_FAILURE;
+                        }
+                        else
+                        {
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_006: [ gballoc_ll_set_decay shall set the dirty decay time for the arena to decay_milliseconds milliseconds. ]*/
+                            if (je_mallctl(command, NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                LogError("failed to set dirty decay time for arena %" PRIu64 "", i);
```
anporumb, line 282: >LogError [](http://example.com/codeflow?start=32&length=8)

should log all parameters with their values.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                        if (snprintf_result < 0)
+                        {
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                            LogError("snprintf failed for arena %" PRIu64 "", i);
+                            result = MU_FAILURE;
+                        }
+                        else
+                        {
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_006: [ gballoc_ll_set_decay shall set the dirty decay time for the arena to decay_milliseconds milliseconds. ]*/
+                            if (je_mallctl(command, NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
```
anporumb, line 279: >NULL, NULL [](http://example.com/codeflow?start=52&length=10)

what happens if non-`NULL` is passed here and the read value is passed to logging?

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                LogError("failed to set dirty decay time for arena %" PRIu64 "", i);
+                                result = MU_FAILURE;
+                                break;
+                            }
+
+                            snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu64 ".muzzy_decay_ms", i);
+
+                            if (snprintf_result < 0)
```
anporumb, line 289: >snprintf_result [](http://example.com/codeflow?start=32&length=15)

same comment as in line 270 about what's the correct way of checking the result of snprintf.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                LogError("snprintf failed for arena %" PRIu64 "", i);
```
anporumb, line 292: >LogError("snprintf failed for arena %" PRIu64 "", i); [](http://example.com/codeflow?start=32&length=53)

should log the complete function with its arguments and their values.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                                result = MU_FAILURE;
+                            }
+                            else
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_007: [ gballoc_ll_set_decay shall set the muzzy decay time for the arena to decay_milliseconds milliseconds. ]*/
+                                if (je_mallctl(command, NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+                                {
+                                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                    LogError("failed to set muzzy decay time for arena %" PRIu64 "", i);
```
anporumb, line 301: >LogError("failed to set muzzy decay time for arena %" PRIu64 "", i); [](http://example.com/codeflow?start=36&length=68)

should log the complete failing function with its arguments and their values

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
     /* Codes_SRS_GBALLOC_LL_JEMALLOC_01_008: [ gballoc_ll_print_stats shall call je_malloc_stats_print and pass to it jemalloc_print_stats_callback as print callback. ]*/
     je_malloc_stats_print(jemalloc_print_stats_callback, NULL, NULL);
 }
+
+int gballoc_ll_set_decay(int64_t decay_milliseconds)
+{
+    int result = 0;
```
anporumb, line 224: >int result = 0; [](http://example.com/codeflow?start=4&length=15)

can the code here be re-written without optimistically initializing result to 0?

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+
+                    /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ For each existing arena: ]*/
+                    for (uint64_t i = 0; i < narenas; i++) {
+                        int snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu64 ".dirty_decay_ms", i);
+
+                        if (snprintf_result < 0)
+                        {
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                            LogError("snprintf failed for arena %" PRIu64 "", i);
+                            result = MU_FAILURE;
```
anporumb, line 274: >result = MU_FAILURE; [](http://example.com/codeflow?start=28&length=20)

why does line 284 have a `break` and here the same break doesn't exist?

The cases where the execution is allowed to continue and the cases where the execution is not allowed to continue are... non-similar.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_006: [ gballoc_ll_set_decay shall set the dirty decay time for the arena to decay_milliseconds milliseconds. ]*/
+                            if (je_mallctl(command, NULL, NULL, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+                            {
+                                /* Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ If there are any errors, gballoc_ll_set_decay shall fail and return a non-zero value. ]*/
+                                LogError("failed to set dirty decay time for arena %" PRIu64 "", i);
+                                result = MU_FAILURE;
+                                break;
+                            }
+
+                            snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu64 ".muzzy_decay_ms", i);
```
anporumb, line 287: >muzzy [](http://example.com/codeflow?start=100&length=5)

muzzy has the same number of characters as "dirty", maybe a complete call to snprintf can be easily avoided.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,97 @@ void gballoc_ll_print_stats(void)
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return result;
+}
```
anporumb, line 314: Because I am asking questions in the review, I need to followup with the answers, so I shall be waiting.

win32/src/gballoc_ll_jemalloc.c
RohitG28: I have added a separate je_mallctl call to read the old decay value. I did not find any documentation that says we can pass both values together(one to read and one to set) and looking at the code it looks like we might have to pass them in separate calls. I will look into the code further and try to run the changes by having both the parts in the same call on an INT cluster before finalizing this change.

win32/src/gballoc_ll_jemalloc.c
RohitG28: Fixed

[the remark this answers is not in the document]
RohitG28: We are now doing best effort undo in case any call fails.

win32/src/gballoc_ll_jemalloc.c
RohitG28: The unsigned size would be platform dependent. So the static assert might fail if the machine's not 64-bit. Do you think using unsigned directly in the code would be a good idea?

win32/devdoc/gballoc_ll_jemalloc.md
RohitG28: Done

win32/devdoc/gballoc_ll_jemalloc.md
RohitG28: we are now doing a best effort undo.

win32/devdoc/gballoc_ll_jemalloc.md
```diff
@@ -165,4 +167,28 @@ static void jemalloc_print_stats_callback(void* context, const char* text)
...
+
+**SRS_GBALLOC_LL_JEMALLOC_28_001: [** `gballoc_ll_set_decay` shall advance jemalloc's internal epoch counter. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_002: [** `gballoc_ll_set_decay` shall set the dirty decay time for new arenas to `decay_milliseconds` milliseconds. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_003: [** `gballoc_ll_set_decay` shall set the muzzy decay time for new arenas to `decay_milliseconds` milliseconds. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_004: [** `gballoc_ll_set_decay` shall fetch the number of existing jemalloc arenas. **]**
+
+**SRS_GBALLOC_LL_JEMALLOC_28_005: [** For each existing arena: **]**
```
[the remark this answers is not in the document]
RohitG28: From what I understand, arenas are never deleted. I am still looking for the documentation that can confirm this or maybe will try to dig into the code further. So in that case we would not need a locking mechanism since new arenas would already have their decay set, and existing arenas would be updated in this loop.

win32/devdoc/gballoc_ll_mimalloc.md
RohitG28: Done

win32/devdoc/gballoc_ll_win32heap.md
RohitG28: Done

win32/devdoc/gballoc_ll_jemalloc.md
RohitG28: I have added a check to fail if the passed decay value is less than -1.

win32/tests/gballoc_ll_jemalloc_ut/gballoc_ll_jemalloc_ut.c
```diff
@@ -628,4 +633,100 @@ TEST_FUNCTION(gballoc_ll_print_stats_calls_the_print_function_with_NULL_does_not
...
+{
+    ///arrange
+    uint64_t num_arenas = NUM_ARENAS;
+
+    char* epoch;
+    char* dirty_decay_ms;
+    char* muzzy_decay_ms;
+    char* narenas;
+
+    STRICT_EXPECTED_CALL(mock_je_mallctl(IGNORED_ARG, NULL, NULL, NULL, 0))
```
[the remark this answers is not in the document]
RohitG28: I have added the argument checks wherever possible in all the calls.

win32/tests/gballoc_ll_jemalloc_ut/gballoc_ll_jemalloc_ut.c
RohitG28: Yeah we are doing it using ValidateArgumentBuffer

win32/src/gballoc_ll_jemalloc.c
RohitG28: I have removed the epoch call since we would not be making use of the stats here.

win32/src/gballoc_ll_jemalloc.c
RohitG28: I have replied to your another comment for this. I would be investigating this further.

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

interfaces/inc/c_pal/gballoc_hl.h
RohitG28: Done

interfaces/inc/c_pal/gballoc_ll.h
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: I have removed the epoch call since we are not using the stats in the code.

win32/src/gballoc_ll_jemalloc.c
RohitG28: Removed

win32/src/gballoc_ll_jemalloc.c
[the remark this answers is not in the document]
RohitG28: dirty pages -> pages which still have data written and are held by the application
muzzy pages -> pages which have their data cleared and held by the application. Dirty pages are cleared by jemalloc and moved to the muzzy pool.

dirty_decay_ms -> how long a page stays in the dirty pool.
muzzy_decay_ms -> how long a page stays in the muzzy pool after its content has been cleared.

win32/src/gballoc_ll_jemalloc.c
RohitG28: I have added the extra check for snprintf. Do you want me to reduce the command array size to something lesser than 64? Since this won't be a frequent operation, I kept it a bit higher.

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: separated the calls

win32/src/gballoc_ll_jemalloc.c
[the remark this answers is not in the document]
RohitG28: Done

win32/devdoc/gballoc_ll_jemalloc.md
RohitG28: The allocations in between would happen as is but once the decay values are set they will just be held for longer in the dirty and muzzy pool rather than being released immediately. So decay values can actually be set before init as well since its an arena level parameter. However, if we plan to pass the decay value to gballoc_hl/ll_init, do you want me to have the same value passed to both dirty and muzzy decay options? We have now separated both the calls. Or should we stick to calling the set_option later?

win32/devdoc/gballoc_ll_passthrough.md
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
anporumb: I do not believe using "unsigned" directly in the code would be a good idea.

win32/src/gballoc_ll_jemalloc.c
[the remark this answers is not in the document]
anporumb: My question at no point contained "muzzy".

win32/src/gballoc_ll_jemalloc.c
RohitG28: Discussed offline

win32/src/gballoc_ll_jemalloc.c
RohitG28: As discussed offline, I have changed it to uin32_t and added a static assert.

win32/src/gballoc_ll_jemalloc.c
anporumb: I've briefly looked at iteration 6, and I cannot notice any INT tests that would confirm behavior.

I strongly advise to have INT tests before all this can be checked in.

RohitG28: Sure I will add the INT tests as well in this PR.

RohitG28:
Following have been covered in the changes:
1) Since we cannot lock the arenas while looping through them, there might be cases where some arenas might get deleted. In those cases, **mallctl** at those arena indices will return **EFAULT**. The code now handles that case and doesn't fail if an arena returns EFAULT.
2) Replaced **arenas.narenas** with **opt.narenas** to get the total number of jemalloc arenas. opt.narenas are the arenas automatically created by jemalloc and we should update their decay using mallct. There is one arena index reserved for huge arena at **opt.narenas + 1**, however jemalloc advises not to change the decay for that arena as huge allocations are usually not reused.

win32/src/gballoc_ll_jemalloc.c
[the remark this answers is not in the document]
RohitG28: opt.narenas = number of arenas automatically created by jemalloc
arenas.narenas = total number of arenas including the manual ones that the application might create using mallctl

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -3,10 +3,14 @@
 
 #include <stddef.h>
 #include <stdlib.h>
+#include <inttypes.h>
+#include <stdio.h>  // for snprintf
 
 #include "c_logging/logger.h"
 #include "c_pal/gballoc_ll.h"
 
+#include "macro_utils/macro_utils.h" // for MU_FAILURE
```
anporumb, line 12: should appear before other headers that use it.  IN this case, right between lines 8 and 9

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+    {
+        int64_t old_decay_milliseconds;
+        size_t old_decay_milliseconds_size = sizeof(old_decay_milliseconds);
+        MU_STATIC_ASSERT(sizeof(int64_t) == sizeof(size_t));
+
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ gballoc_ll_set_option shall retrieve the old dirty decay value and set the new dirty decay value to decay_milliseconds for new arenas by calling je_mallctl with arenas.dirty_decay_ms as the command. ]*/
+        if (je_mallctl("arenas.dirty_decay_ms", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+        {
+            /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+            LogError("je_mallctl(const char* name=arenas.dirty_decay_ms, void* oldp=%p, size_t* oldlenp=%p, void* newp=%p, size_t newlen=%zu) failed", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds));
```
anporumb, line 242: >const char* [](http://example.com/codeflow?start=33&length=11)

no need to specify the parameter types when logging an error from a called function.

Here and everywhere

anporumb, line 242: >oldp [](http://example.com/codeflow?start=79&length=4)

this is not the argument name. The function is called with "&old_decay_milliseconds".

Just not to repeat myself, change here and everywhere where logging has not been updated with the new argument names.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+    {
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_019: [ If decay_milliseconds is less than -1, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+        LogError("decay_milliseconds must be -1 or greater, decay_milliseconds=%" PRId64 "", decay_milliseconds);
+        result = MU_FAILURE;
+    }
+    else
+    {
+        int64_t old_decay_milliseconds;
+        size_t old_decay_milliseconds_size = sizeof(old_decay_milliseconds);
+        MU_STATIC_ASSERT(sizeof(int64_t) == sizeof(size_t));
```
anporumb, line 236: >MU_STATIC_ASSERT(sizeof(int64_t) == sizeof(size_t)); [](http://example.com/codeflow?start=8&length=52)

move all these asserts to somewhere near the top of the file, add relevant comments why we make these asserts.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+
+                /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_008: [ For each existing arena except last (since it is reserved for huge arena) ]*/
+                for (i = 0; i < narenas; i++)
+                {
+                    int snprintf_result = snprintf(command, sizeof(command), "arena.%" PRIu32 ".dirty_decay_ms", i);
+
+                    if (snprintf_result < 0 || (size_t)snprintf_result >= sizeof(command))
+                    {
+                        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+                        LogError("snprintf(const char* buffer=%p, size_t size=%zu, const char* format=arena.<i>.dirty_decay_ms, i=%" PRIu32 ") failed %s", command, sizeof(command), i, (snprintf_result < 0) ? " with encoding error" : " due to insufficient buffer size");
```
anporumb, line 271: >(snprintf_result < 0) ? " with encoding error" : " due to insufficient buffer size" [](http://example.com/codeflow?start=184&length=83)

this is nice!!

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+                        result = MU_FAILURE;
```
anporumb, line 272: > result = MU_FAILURE; [](http://example.com/codeflow?start=23&length=21)

probably not needed.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+                            {
+                                /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_020: [ If je_mallctl returns EFAULT, gballoc_ll_set_option shall continue without failing as this error is expected when the arena doesn't exist. ]*/
+                                LogError("je_mallctl(const char* name=%s, void* oldp=NULL, size_t* oldlenp=NULL, void* newp=%p, size_t newlen=%zu) failed, continuing without failure as the arena doesn't exist, old_decay_milliseconds=%" PRId64 "", command, &decay_milliseconds, sizeof(decay_milliseconds), old_decay_milliseconds);
+                            }
+                            else
+                            {
+                                /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+                                LogError("je_mallctl(const char* name=%s, void* oldp=NULL, size_t* oldlenp=NULL, void* newp=%p, size_t newlen=%zu) failed, old_decay_milliseconds=%" PRId64 "", command, &decay_milliseconds, sizeof(decay_milliseconds), old_decay_milliseconds);
+                                result = MU_FAILURE;
```
anporumb, line 290: >result = MU_FAILURE; [](http://example.com/codeflow?start=32&length=20)

probably not needed, gets overwritten in line 304 anyway

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+                        }
+                    }
+                }
+
+                if (i < narenas)
+                {
+                    for (uint32_t j = 0; j < i; j++)
+                    {
+                        (void)snprintf(command, sizeof(command), "arena.%" PRIu32 ".dirty_decay_ms", j);
+                        (void)je_mallctl(command, NULL, NULL, &old_decay_milliseconds, sizeof(old_decay_milliseconds));
```
anporumb, line 302: >je_mallctl [](http://example.com/codeflow?start=30&length=10)

undo-ing of previously performed actions should still log if it fails.

Here and everywhere.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+                    LogInfo("jemalloc muzzy_decay_ms set to %" PRId64 ", old value was %" PRId64 "", decay_milliseconds, old_decay_milliseconds);
+                    goto all_ok;
+                }
+            }
+            (void)je_mallctl("arenas.muzzy_decay_ms", NULL, NULL, &old_decay_milliseconds, sizeof(old_decay_milliseconds));
+        }
+    }
+all_ok:
+    return result;
+}
```
anporumb, line 417: all comments for "dirty" apply to "muzzy" too.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+{
+    int result;
+
+    if (/*Codes_SRS_GBALLOC_LL_JEMALLOC_28_001: [ If option_name is NULL, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+        option_name == NULL ||
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_002: [ If option_value is NULL, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+        option_value == NULL)
+    {
+        LogError("Invalid args: char* option_name = %p, void* option_value = %p",
+            option_name, option_value);
```
anporumb, line 429: >option_name [](http://example.com/codeflow?start=12&length=11)

use MU_P_OR_NULL

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+                }
+            }
+            (void)je_mallctl("arenas.muzzy_decay_ms", NULL, NULL, &old_decay_milliseconds, sizeof(old_decay_milliseconds));
+        }
+    }
+all_ok:
+    return result;
+}
+
+int gballoc_ll_set_option(char* option_name, void* option_value)
```
anporumb, line 419: >char [](http://example.com/codeflow?start=26&length=4)

why missing const?

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
```diff
@@ -260,4 +263,131 @@ TEST_FUNCTION(gballoc_ll_print_stats_works)
...
 }
 
+static void assert_set_option_works_for_decay(const char* option_name, int64_t decay_ms, uint32_t narenas)
+{
+    char command[32];
+    int snprintf_result;
+    int64_t verify_decay_ms;
+    size_t decay_ms_size = sizeof(decay_ms);
+
+    // Arena 1 does not exist, so je_mallctl should return EFAULT
```
anporumb, line 273: >Arena 1 does not exist [](http://example.com/codeflow?start=7&length=22)

isn't this the huge arena?

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
```diff
@@ -260,4 +263,131 @@ TEST_FUNCTION(gballoc_ll_print_stats_works)
...
+    // Initialize all the arenas except first(arena at index 1) after the default arena, so that we can verify that the decay is set for all the arenas except arena at index 1
+    uint32_t arena_id;
+    for (uint32_t i = 2; i < narenas; i++)
+    {
+        arena_id = i; // assign thread to arena i
+        ASSERT_ARE_EQUAL(int, 0, je_mallctl("thread.arena", NULL, NULL, &arena_id, sizeof(arena_id)));
+
+        ptr[i] = gballoc_ll_malloc(4); // this malloc will happen in the arena i
+        ASSERT_IS_NOT_NULL(ptr[i]);
+    }
```
anporumb, line 312: >} [](http://example.com/codeflow?start=4&length=1)

function should assign the thread back to its original arena

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
```diff
@@ -260,4 +263,131 @@ TEST_FUNCTION(gballoc_ll_print_stats_works)
...
+
+    // verify that all the arenas have the decay set except arena 1 and last arena(reserved for huge arena)
+    assert_set_option_works_for_decay("muzzy_decay", decay_ms, expected_narenas);
+
+    ///clean
+    for (uint32_t i = 0; i < expected_narenas; i++)
+    {
+        gballoc_ll_free(ptr[i]);
+    }
+}
```
anporumb, line 391: I think we're missing tests here, and it is better to have them now and here (in c-pal) rather than to discover that they don't work later and in production.

We should have a test that actually sees when memory is released to OS and depending on the config, it is released earlier/later/never at all.

I can imagine something like:
let's start like with a number of threads, all of them pretending to do memory work (alloc/free) for an amount of time.

Following this amount of time, some memory should still be used, some memory should be free(). Take a snapshot of the memory used by the process.

Wait past the decay timer (there's no other activity)

Take another snapshot of the memory used by the process.

The diff between the 2 snapshots and the decay parameter should be correlated as a condition to pass the test. With some tolerances, ofc.

anporumb: I still think this PR requires at least the above work to be completed, so I'll still be waiting. Rest looks good.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +218,240 @@ void gballoc_ll_print_stats(void)
...
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ gballoc_ll_set_option shall retrieve the old dirty decay value and set the new dirty decay value to decay_milliseconds for new arenas by calling je_mallctl with arenas.dirty_decay_ms as the command. ]*/
+        if (je_mallctl("arenas.dirty_decay_ms", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+        {
+            /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+            LogError("je_mallctl(const char* name=arenas.dirty_decay_ms, void* oldp=%p, size_t* oldlenp=%p, void* newp=%p, size_t newlen=%zu) failed", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds));
+            result = MU_FAILURE;
+        }
+        else
+        {
+            uint32_t narenas;
```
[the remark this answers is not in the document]
RohitG28: Done

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
RohitG28: no the test deliberately skips creating this arena 1. Huge arena is the last arena(arena 4)

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
RohitG28: Done

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
RohitG28: I have added new tests that check the dirty and muzzy page numbers after we decay the arenas. Hope these would be sufficient to test the changes?

win32/src/gballoc_ll_jemalloc.c
RohitG28: Added

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Done

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
anporumb: I looked at the tests, and while they do seem to test what I've mentioned here, it is a case of "testing jemalloc with jemalloc".

When I said before "Take a snapshot of the memory used by the process." I was referring to the working set size. So let's compare the working set "before" and "after" by calling Windows APIs and not use jemalloc's internals.

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
```diff
@@ -260,4 +273,456 @@ TEST_FUNCTION(gballoc_ll_print_stats_works)
...
+    {
+        result = gballoc_ll_set_option("dirty_decay", &decay_ms);
+    }
+    else
+    {
+        // set dirty decay to a low value so as to avoid retaining the dirty pages for long (this ensures that pages are moved to muzzy sooner)
+        int64_t low_dirty_decay_ms = 1;
+        result = gballoc_ll_set_option("dirty_decay", &low_dirty_decay_ms);
+        ASSERT_ARE_EQUAL(int, 0, result);
+        result = gballoc_ll_set_option("muzzy_decay", &decay_ms);
```
anporumb, line 437: >result [](http://example.com/codeflow?start=8&length=6)

should be asserted

anporumb: probably line 439 needs to become line 437.5

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +224,284 @@ void gballoc_ll_print_stats(void)
...
+    else
+    {
+        int64_t old_decay_milliseconds;
+        size_t old_decay_milliseconds_size = sizeof(old_decay_milliseconds);
+
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_005: [ gballoc_ll_set_option shall retrieve the old dirty decay value and set the new dirty decay value to decay_milliseconds for new arenas by calling je_mallctl with arenas.dirty_decay_ms as the command. ]*/
+        if (je_mallctl("arenas.dirty_decay_ms", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+        {
+            /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+            LogError("je_mallctl(arenas.dirty_decay_ms, &old_decay_milliseconds=%p, &old_decay_milliseconds_size=%p, &decay_milliseconds=%p, sizeof(decay_milliseconds)=%zu) failed", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds));
```
anporumb, line 247: >LogError [](http://example.com/codeflow?start=12&length=8)

this logerror should make an attempt at printing the value of `old_decay_milliseconds` even when je_mallctl failed. Probably initialize `old_decay_milliseconds` to some negative number just in case.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +224,284 @@ void gballoc_ll_print_stats(void)
...
+    else
+    {
+        int64_t old_decay_milliseconds;
+        size_t old_decay_milliseconds_size = sizeof(old_decay_milliseconds);
+
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_012: [ gballoc_ll_set_option shall retrieve the old muzzy decay value and set the new muzzy decay value to decay_milliseconds for new arenas by calling je_mallctl with arenas.muzzy_decay_ms as the command. ]*/
+        if (je_mallctl("arenas.muzzy_decay_ms", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds)) != 0)
+        {
+            /*Codes_SRS_GBALLOC_LL_JEMALLOC_28_018: [ If there are any errors, gballoc_ll_set_option shall fail and return a non-zero value. ]*/
+            LogError("je_mallctl(arenas.muzzy_decay_ms, ol&old_decay_millisecondsdp=%p, &old_decay_milliseconds_size=%p, &decay_milliseconds=%p, sizeof(decay_milliseconds)=%zu) failed", &old_decay_milliseconds, &old_decay_milliseconds_size, &decay_milliseconds, sizeof(decay_milliseconds));
```
anporumb, line 368: >ol&old_decay_millisecondsdp=%p [](http://example.com/codeflow?start=56&length=30)

looks like a wrong paste here was performed here.

win32/src/gballoc_ll_jemalloc.c
```diff
@@ -214,3 +224,284 @@ void gballoc_ll_print_stats(void)
...
+                LogError("je_mallctl(arenas.dirty_decay_ms, NULL, NULL, &old_decay_milliseconds=%p, sizeof(old_decay_milliseconds)=%zu) failed during cleanup", &old_decay_milliseconds, sizeof(old_decay_milliseconds));
+            }
+        }
+    }
+
+all_ok:
+    return result;
+}
+
+static int gballoc_ll_set_muzzy_decay(int64_t decay_milliseconds)
```
anporumb, line 349: >gballoc_ll_set_muzzy_decay [](http://example.com/codeflow?start=11&length=26)

muzzy and dirty functions are SAME except the spelling of muzzy/decay here and there.

Any chance we collapse the 2 functions in just 1? Halve the chances of a bug with half the code!

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
anporumb: Need to see the next version of the INT tests with Windows.

I understand if there are no INT tests for Linux at this time, but there should be some INT tests for Linux too using the parallel mechanisms existing for Linux.

RohitG28: Done

win32/src/gballoc_ll_jemalloc.c
RohitG28: Merged the two

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
```diff
@@ -260,4 +276,377 @@ TEST_FUNCTION(gballoc_ll_print_stats_works)
...
+{
+    PROCESS_MEMORY_COUNTERS memCounters;
+    if (GetProcessMemoryInfo(GetCurrentProcess(), &memCounters, sizeof(memCounters)))
+    {
+        *working_set_size = memCounters.WorkingSetSize;
+        return 0;
+    }
+    else
+    {
+        LogError("GetProcessMemoryInfo failed with error %d", GetLastError());
```
anporumb, line 416: >LogError [](http://example.com/codeflow?start=8&length=8)

Also known as "LogLastError"

anporumb (dismissed):
:shipit:

linux/src/gballoc_hl_passthrough.c
```diff
@@ -229,4 +229,10 @@ const GBALLOC_LATENCY_BUCKET_METADATA* gballoc_hl_get_latency_bucket_metadata(vo
     /*Codes_SRS_GBALLOC_HL_PASSTHROUGH_02_016: [ Each consecutive bucket shall be [1 << n, (1 << (n + 1)) - 1], where n starts at 9. ]*/
     return latency_buckets_metadata;
 }
+
+int gballoc_hl_set_option(const char* option_name, void* option_value)
+{
+    /* Codes_SRS_GBALLOC_HL_PASSTHROUGH_28_001: [ gballoc_hl_set_option shall call gballoc_ll_set_option with option_name and option_value as arguments. ]*/
+    return gballoc_ll_set_option(option_name, option_value);
```
dcristoloveanu, line 236: >gballoc_ll_set_option [](http://example.com/codeflow?start=11&length=21)

Nice that we got set option!

dcristoloveanu (approved):
:shipit:

win32/tests/gballoc_ll_jemalloc_int/gballoc_ll_jemalloc_int.c
RohitG28: Done

ipg2013 (approved):

PR #416, opened 2024-11-24
Linux: Removing threadpool_open and threadpool_close along with SM states

nishudeshu:
Removing Threadpool Open and Close and SM states.

### DIFF
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -429,10 +429,6 @@ MOCKABLE_FUNCTION(, void, threadpool_destroy_work_item, THANDLE(THREADPOOL), thr
 
 **SRS_THREADPOOL_LINUX_05_033: [** `threadpool_destroy_work_item` shall wait for all pending work items to finish execution. **]**
 
-  - **SRS_THREADPOOL_LINUX_05_034: [** When all pending work items finish execution, `threadpool_destroy_work_item` shall acquire the `SRW lock` in shared mode by calling `srw_lock_acquire_shared`. **]**
-
   - **SRS_THREADPOOL_LINUX_05_035: [** `threadpool_destroy_work_item` shall free the memory allocated to the work item of type `THREADPOOL_WORK_ITEM_HANDLE` created in `threadpool_create_work_item`. **]**
-
-  - **SRS_THREADPOOL_LINUX_05_036: [** `threadpool_destroy_work_item` shall release the shared `SRW lock`. **]**
 
   - **SRS_THREADPOOL_LINUX_05_037: [** If `InterlockedHL_WaitForValue` does not return `INTERLOCKED_HL_OK` then Log Message with severity `CRITICAL` and `terminate`. **]**
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -405,11 +405,11 @@ MOCKABLE_FUNCTION(, int, threadpool_schedule_work_item, THANDLE(THREADPOOL), thr
 
 - **SRS_THREADPOOL_LINUX_05_019: [** If reallocating the task array fails, `threadpool_schedule_work_item` shall fail and return a non-zero value. **]**
 
-- **SRS_THREADPOOL_LINUX_05_020: [** Otherwise, `threadpool_schedule_work_item` shall acquire the SRW lock in exclusive mode by calling srw_lock_acquire_exclusive. **]**
+- **SRS_THREADPOOL_LINUX_05_020: [** Otherwise, `threadpool_schedule_work_item` shall acquire the SRW lock in shared mode by calling srw_lock_acquire_exclusive. **]**
 
 - **SRS_THREADPOOL_LINUX_05_022: [** `threadpool_schedule_work_item` shall copy the `work_function` and `work_function_context` from `threadpool_work_item` into insert position in the task array. **]**
  
-- **SRS_THREADPOOL_LINUX_05_023: [** `threadpool_schedule_work_item` shall set the `task_state` to `TASK_WAITING` and then release the exclusive SRW lock by calling srw_lock_release_exclusive. **]**
+- **SRS_THREADPOOL_LINUX_05_023: [** `threadpool_schedule_work_item` shall set the `task_state` to `TASK_WAITING` and then release the shared SRW lock by calling srw_lock_release_exclusive. **]**
 
 - **SRS_THREADPOOL_LINUX_05_025: [** `threadpool_schedule_work_item` shall unblock the `threadpool` semaphore by calling `sem_post`. **]**
 
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -140,7 +140,7 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees the resouces associated with threadpool.
 
-**SRS_THREADPOOL_LINUX_07_089: [** `threadpool_dispose` shall signal all threads that the threadpool is closing by setting the flag to stop the threads to 1. **]**
+**SRS_THREADPOOL_LINUX_07_089: [** `threadpool_dispose` shall signal all threads to return. **]**
 
 **SRS_THREADPOOL_LINUX_07_027: [** `threadpool_dispose` shall join all threads in the `threadpool`. **]**
 
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -167,7 +167,7 @@ Note: `threadpool_open` will be deprecated and `threadpool_create` will perform 
 ```C
 MOCKABLE_FUNCTION(, void, threadpool_close, THANDLE(THREADPOOL), threadpool);
 ```
-Note: `threadpool_close` will be deprecated and threadpool_dispose will perform additional tasks of threadpool_close. This function will exist until all the libraries calling this API are modified to use only `threadpool_create`.
+Note: `threadpool_close` will be deprecated and `threadpool_dispose` will perform additional tasks of `threadpool_close`. This function will exist until all the libraries calling this API are modified to use only `threadpool_create`.
 `threadpool_close` closes the threadpool.
 
 **SRS_THREADPOOL_LINUX_07_025: [** If `threadpool` is `NULL`, `threadpool_close` shall fail and return. **]**
--- a/linux/devdoc/threadpool_linux_requirements.md
+++ b/linux/devdoc/threadpool_linux_requirements.md
@@ -140,7 +140,7 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees the resouces associated with threadpool.
 
-**SRS_THREADPOOL_LINUX_07_089: [** `threadpool_dispose` shall signal all threads `threadpool` is closing by calling `InterlockedHL_SetAndWakeAll`.  **]**
+**SRS_THREADPOOL_LINUX_07_089: [** `threadpool_dispose` shall signal all threads that the threadpool is closing by setting the flag to stop the threads to 1. **]**
 
 **SRS_THREADPOOL_LINUX_07_027: [** `threadpool_dispose` shall join all threads in the `threadpool`. **]**
 
@@ -158,9 +158,9 @@ MOCKABLE_FUNCTION(, int, threadpool_open, THANDLE(THREADPOOL), threadpool);
 Note: `threadpool_open` will be deprecated and `threadpool_create` will perform additional tasks of `threadpool_open`. This function will exist until all the libraries calling this API are modified to use only `threadpool_create`.
 `threadpool_open` opens the threadpool.
 
-**SRS_THREADPOOL_LINUX_05_047: [** If `threadpool` is `NULL`, `threadpool_open` shall fail and return a non-zero value. **]**
+**SRS_THREADPOOL_LINUX_07_017: [** If `threadpool` is `NULL`, `threadpool_open` shall fail and return a non-zero value. **]**
 
-**SRS_THREADPOOL_LINUX_05_050: [** Otherwise, `threadpool_open` shall succeed and return 0. **]**
+**SRS_THREADPOOL_LINUX_07_024: [** Otherwise, `threadpool_open` shall succeed and return 0. **]**
 
 ### threadpool_close
 
@@ -361,23 +361,7 @@ static int threadpool_work_func(void* param);
 
 **SRS_THREADPOOL_LINUX_07_084: [** If the work item function is not `NULL`, `threadpool_work_func` shall execute it with `work_function_ctx`. **]**
 
-- **SRS_THREADPOOL_LINUX_05_039: [** `threadpool_work_func` shall acquire the shared `SRW lock` by calling `srw_lock_acquire_shared`. **]**
-
-- **SRS_THREADPOOL_LINUX_05_040: [** `threadpool_work_func` shall save `pending_work_item_count_ptr` is not `NULL` in `is_pending_work_item_count_ptr_not_null`. **]**
-
-- **SRS_THREADPOOL_LINUX_05_041: [** `threadpool_work_func` shall release the shared `SRW lock` by calling `srw_lock_release_shared`. **]**
-
-- **SRS_THREADPOOL_LINUX_05_042: [** If the `is_pending_work_item_count_ptr_not_null` is `TRUE` then: **]**
-
-  - **SRS_THREADPOOL_LINUX_05_043: [** `threadpool_work_func` shall acquire the exclusive `SRW lock` by calling `srw_lock_acquire_exclusive`. **]**
- 
-  - **SRS_THREADPOOL_LINUX_05_044: [** `threadpool_work_func` shall decrement the `pending_work_item_count_ptr` by calling `interlocked_decrement`. **]**
-    
-  - **SRS_THREADPOOL_LINUX_05_045: [** `threadpool_work_func` shall send wake up signal to single listener for the address in `pending_work_item_count_ptr`. **]**
-
-  - **SRS_THREADPOOL_LINUX_05_046: [** `threadpool_work_func` shall release the shared `SRW lock` by calling `srw_lock_release_exclusive`. **]**
-
-**SRS_THREADPOOL_LINUX_07_085: [** `threadpool_work_func` shall loop until `threadpool_close` or `threadpool_destroy` is called. **]**
+**SRS_THREADPOOL_LINUX_07_085: [** `threadpool_work_func` shall loop until the flag to stop the threads is not set to 1. **]**
 
 ### threadpool_create_work_item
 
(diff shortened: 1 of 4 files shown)


linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -143,16 +140,14 @@ static void threadpool_dispose(THREADPOOL* threadpool)
 
 `threadpool_dispose` frees the resouces associated with threadpool.
 
-**SRS_THREADPOOL_LINUX_07_013: [** `threadpool_dispose` shall perform an implicit close if `threadpool` is open. **]**
+**SRS_THREADPOOL_LINUX_07_089: [** `threadpool_dispose` shall signal all threads that the threadpool is closing by setting the flag to stop the threads to 1. **]**
```
dcristoloveanu, line 143: >closing [](http://example.com/codeflow?start=104&length=7)

If we say "shall signal all threads to return" we don't have to mention closing (since we do not have close as a state any longer)

dcristoloveanu:
> #include "c_pal/sm.h"

This is not needed any longer

---
Refers to: linux/src/threadpool_linux.c:32 in a679d72. [](commit_id = a679d7201516a537460b55b0bd17b8762d887dd1, deletion_comment = False)

linux/devdoc/threadpool_linux_requirements.md
nishudeshu: Thanks. Was unable to find right word to replace threadpool closing. Will do this.

jebrando: stopping would work here.

nishudeshu:
> #include "c_pal/sm.h"

Removed.

---
In reply to: [2512829554](https://github.com/Azure/c-pal/pull/416#issuecomment-2512829554) [](http://example.com/codeflow?ancestors=2512829554)

---
Refers to: linux/src/threadpool_linux.c:32 in a679d72. [](commit_id = a679d7201516a537460b55b0bd17b8762d887dd1, deletion_comment = False)

linux/src/threadpool_linux.c
```diff
@@ -217,29 +195,23 @@ static int threadpool_work_func(void* param)
...
+                        {                            
                             srw_lock_acquire_exclusive(threadpool->srw_lock);
-                            /* Codes_SRS_THREADPOOL_LINUX_05_044: [ threadpool_work_func shall decrement the pending_work_item_count_ptr by calling interlocked_decrement. ]*/
-                            interlocked_decrement(threadpool->task_array[current_index].pending_work_item_count_ptr);
-                            /* Codes_SRS_THREADPOOL_LINUX_05_045: [ threadpool_work_func shall send wake up signal to single listener for the address in pending_work_item_count_ptr. ]*/
-                            wake_by_address_single(threadpool->task_array[current_index].pending_work_item_count_ptr);
-                            /* Codes_SRS_THREADPOOL_LINUX_05_046: [ threadpool_work_func shall release the shared SRW lock by calling srw_lock_release_exclusive. ]*/
+                            (void)interlocked_decrement(threadpool->task_array[current_index].pending_work_item_count_ptr);
+                            wake_by_address_single(threadpool->task_array[current_index].pending_work_item_count_ptr);
```
jebrando, line 207: Since you're only waiting up if the value is zero, then you need to have an if on the decrement as to only wait on zero, it makes it more efficient.

if (interlocked_decrement(...) == 0)
{
     wake_by_address_single(...);
}

jebrando:
> int threadpool_open(THANDLE(THREADPOOL) threadpool)

Maybe add a comment saying that the open and close will go away soon.

---
Refers to: linux/src/threadpool_linux.c:446 in a679d72. [](commit_id = a679d7201516a537460b55b0bd17b8762d887dd1, deletion_comment = False)

jebrando (dismissed):
:shipit:

linux/src/threadpool_linux.c
```diff
@@ -325,18 +297,26 @@ static int reallocate_threadpool_array(THREADPOOL* threadpool)
...
+    /* Codes_SRS_THREADPOOL_LINUX_07_089: [ threadpool_dispose shall signal all threads that the threadpool is closing by setting the flag to stop the threads to 1. ]*/
+    (void)InterlockedHL_SetAndWakeAll(&threadpool->stop_thread, 1);
+    for (int32_t index = 0; index < threadpool->used_thread_count; index++)
+    {
+        int dont_care;
+        /* Codes_SRS_THREADPOOL_LINUX_07_027: [ threadpool_dispose shall join all threads in the threadpool. ]*/
+        if (ThreadAPI_Join(threadpool->thread_handle_array[index], &dont_care) != THREADAPI_OK)
+        {
+            LogError("Failure joining thread number %" PRId32 "", index);
+        }
```
dcristoloveanu, line 308: >} [](http://example.com/codeflow?start=8&length=1)

We usually have the other branch in the code even when emoty. We put a comment in the empty branch like "// everything ok". That is only to make it nicer and explicit to developers not to wonder that a branch was missed.

linux/src/threadpool_linux.c
```diff
@@ -359,84 +339,101 @@ THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
...
-                            /* Codes_SRS_THREADPOOL_LINUX_07_009: [ threadpool_create shall create a shared semaphore with initialized value zero. ]*/
-                            if (sem_init(&result->semaphore, 0 , 0) != 0)
+                            (void)interlocked_exchange(&result->stop_thread, 0);
+                            (void)interlocked_exchange(&result->task_count, 0);
+
+                            /* Codes_SRS_THREADPOOL_LINUX_07_010: [ insert_idx and consume_idx for the task array shall be initialized to 0. ]*/
+                            (void)interlocked_exchange_64(&result->insert_idx, 0);
+                            (void)interlocked_exchange_64(&result->consume_idx, 0);
+
+                            int32_t index;
```
dcristoloveanu, line 401: >int32_t [](http://example.com/codeflow?start=28&length=7)

uint32_t

linux/src/threadpool_linux.c
```diff
@@ -868,76 +776,61 @@ int threadpool_schedule_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WOR
...
+                task_item->work_function_ctx = threadpool_work_item->work_function_ctx;
+                task_item->work_function = threadpool_work_item->work_function;
 
-                    /* Codes_SRS_THREADPOOL_LINUX_05_024: [ threadpool_schedule_work_item shall notify a single thread that is waiting for update of this value by a wake signal. ]*/
-                    wake_by_address_single(task_item->pending_work_item_count_ptr);
+                /* Codes_SRS_THREADPOOL_LINUX_05_023: [ threadpool_schedule_work_item shall set the task_state to TASK_WAITING and then release the exclusive SRW lock by calling srw_lock_release_exclusive. ]*/
+                (void)interlocked_exchange(&task_item->task_state, TASK_WAITING);
 
-                    srw_lock_release_exclusive(threadpool_ptr->srw_lock);
+                wake_by_address_single(task_item->pending_work_item_count_ptr);
```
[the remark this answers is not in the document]
nishudeshu: I thought may be during debug or crash dump, the place of call for Interlocked HL Wait for value should get the updated value.

linux/src/threadpool_linux.c
nishudeshu: Okay, generally if the code changes or expands, and we add more listeners to this address; like do something when the pending work item count becomes 100, wouldn't this constraint create an issue and cause major debug issue? I know in this case we are waiting on 0 but in general would it be a big over head to send these signals?

linux/src/threadpool_linux.c
```diff
@@ -359,84 +339,101 @@ THANDLE(THREADPOOL) threadpool_create(EXECUTION_ENGINE_HANDLE execution_engine)
...
+                                }
+                            }
+
+                            /* Codes_SRS_THREADPOOL_LINUX_07_022: [ If one of the thread creation fails, threadpool_create shall fail and return a non-zero value, terminate all threads already created. ]*/
+                            if (index < result->used_thread_count)
                             {
-                                /* Codes_SRS_THREADPOOL_LINUX_07_011: [ If any error occurs, threadpool_create shall fail and return NULL. ]*/
-                                LogError("Failure creating sem_init");
+                                (void)interlocked_exchange(&result->stop_thread, 1);
+                                for (int32_t inner = 0; inner < index; inner++)
```
dcristoloveanu, line 417: >int32_t [](http://example.com/codeflow?start=37&length=7)

uint32_t

linux/src/threadpool_linux.c
nishudeshu: It was a moved code. Fixed.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
```diff
@@ -103,6 +104,7 @@ static void threadpool_long_task(void* context)
     WRAP_DATA* data = context;
     ASSERT_ARE_EQUAL(int, 0, strcmp(data->mem, "READY"));
     strcpy(data->mem, "DONE");
+    ThreadAPI_Sleep(1);
```
dcristoloveanu, line 107: >ThreadAPI_Sleep(1); [](http://example.com/codeflow?start=4&length=19)

Why is this Sleep needed?

linux/src/threadpool_linux.c
nishudeshu: Added an else block

linux/src/threadpool_linux.c
```diff
@@ -325,18 +297,26 @@ static int reallocate_threadpool_array(THREADPOOL* threadpool)
 
 static void threadpool_dispose(THREADPOOL* threadpool)
 {
-    /* Codes_SRS_THREADPOOL_LINUX_07_013: [ threadpool_destroy shall perform an implicit close if threadpool is open. ]*/
-    internal_close(threadpool);
+    /* Codes_SRS_THREADPOOL_LINUX_07_089: [ threadpool_dispose shall signal all threads that the threadpool is closing by setting the flag to stop the threads to 1. ]*/
+    (void)InterlockedHL_SetAndWakeAll(&threadpool->stop_thread, 1);
+    for (int32_t index = 0; index < threadpool->used_thread_count; index++)
```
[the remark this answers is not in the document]
nishudeshu: Fixed.

linux/tests/threadpool_linux_ut/threadpool_linux_ut.c
```diff
@@ -371,63 +354,105 @@ TEST_FUNCTION(creating_2_threadpool_succeeds)
...
+/* Tests_SRS_THREADPOOL_LINUX_07_016: [ threadpool_dispose shall free the memory allocated in threadpool_create. ]*/
+/* Tests_SRS_THREADPOOL_LINUX_07_014: [ threadpool_dispose shall destroy the semphore by calling sem_destroy. ]*/
+/* Tests_SRS_THREADPOOL_LINUX_07_015: [ threadpool_dispose shall destroy the SRW lock by calling srw_lock_destroy. ]*/
+TEST_FUNCTION(threadpool_dispose_performs_an_implicit_close)
 {
     // arrange
-    THANDLE(THREADPOOL) threadpool = threadpool_create(test_execution_engine);
+    THANDLE(THREADPOOL) threadpool = test_create_threadpool();
     ASSERT_ARE_EQUAL(int, 0, threadpool_open(threadpool));
     umock_c_reset_all_calls();
```
dcristoloveanu, line 396: Not needed anymore.

linux/tests/threadpool_linux_int/threadpool_linux_int.c
nishudeshu: It was creating a raise condition and the counter would stop 4998. Similar to threadpool_long_task_v2, it is solving the problem for this test.

linux/src/threadpool_linux.c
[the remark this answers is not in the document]
nishudeshu: Fixed

linux/src/threadpool_linux.c
nishudeshu: Fixed

linux/tests/threadpool_linux_int/threadpool_linux_int.c
nishudeshu: Resolved by changing Sleep to Shared locks.

linux/src/threadpool_linux.c
```diff
@@ -217,29 +194,23 @@ static int threadpool_work_func(void* param)
...
+                            (void)interlocked_decrement(threadpool->task_array[current_index].pending_work_item_count_ptr);
+                            wake_by_address_single(threadpool->task_array[current_index].pending_work_item_count_ptr);                         
                             srw_lock_release_exclusive(threadpool->srw_lock);
                         }
                     }
                 }
             }
-        /* Codes_SRS_THREADPOOL_LINUX_07_085: [ threadpool_work_func shall loop until threadpool_close or threadpool_destroy is called. ]*/
-        } while (interlocked_add(&threadpool->run_thread, 0) != 1);
+        /* Codes_SRS_THREADPOOL_LINUX_07_085: [ threadpool_work_func shall loop until the flag to stop the threadss is not set to 1. ]*/
```
dcristoloveanu, line 214: >threadss [](http://example.com/codeflow?start=107&length=8)

threads

linux/src/threadpool_linux.c
```diff
@@ -217,29 +194,25 @@ static int threadpool_work_func(void* param)
...
                         bool is_pending_work_item_count_ptr_not_null = (NULL != threadpool->task_array[current_index].pending_work_item_count_ptr);
-                        /* Codes_SRS_THREADPOOL_LINUX_05_041: [ threadpool_work_func shall release the shared SRW lock by calling srw_lock_release_shared. ]*/
                         srw_lock_release_shared(threadpool->srw_lock);
-                        /* Codes_SRS_THREADPOOL_LINUX_05_042: [ If the is_pending_work_item_count_ptr_not_null is TRUE then: ]*/
+
                         if (is_pending_work_item_count_ptr_not_null)
-                        {
-                            /* Codes_SRS_THREADPOOL_LINUX_05_043: [ threadpool_work_func shall acquire the exclusive SRW lock by calling srw_lock_acquire_exclusive. ]*/
+                        {                            
                             srw_lock_acquire_exclusive(threadpool->srw_lock);
```
dcristoloveanu, line 204: >srw_lock_acquire_exclusive(threadpool->srw_lock); [](http://example.com/codeflow?start=28&length=49)

Why are we taking the lock exclusively here?
This lock was meant to be acquired exclusively only when we resize the task array.
And it is meant to be acquired shared when any operation is performed on the items in the array.

linux/src/threadpool_linux.c
```diff
@@ -955,33 +857,21 @@ void threadpool_destroy_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WOR
     else
     {
         THREADPOOL* threadpool_ptr = THANDLE_GET_T(THREADPOOL)(threadpool);
-
-        /* Codes_SRS_THREADPOOL_LINUX_05_031: [ threadpool_destroy_work_item shall call sm_exec_begin ]*/
-        SM_RESULT sm_result = sm_exec_begin(threadpool_ptr->sm);
-        if (sm_result != SM_EXEC_GRANTED)
+        
+        /* Codes_SRS_THREADPOOL_LINUX_05_033: [ threadpool_destroy_work_item shall wait for all pending work items to finish execution. ]*/
+        if (InterlockedHL_WaitForValue(&threadpool_work_item->pending_work_item_count, 0, UINT32_MAX) == INTERLOCKED_HL_OK)
```
dcristoloveanu, line 862: >pending_work_item_count [](http://example.com/codeflow?start=62&length=23)

Overall I think using this pending_work_item_count is wrong. Because it tried to act as a reference count, but it is very bad at it.

For example, assume this:

A work item is created
A thread 1 schedules the work item to execute => pending_work_item_count = 1
Callback executes, => pending_work_item_count = 0
Thread 1 calls destroy_work_item, succesfully executes ln 862.
Thread 2 comes and schedules the work item => pending_work_item_count = 1
Thread 1 frees the work item.

There is no safety against this kind of race.

I believe the correct way to handle this is proper ref counting (either sm or THANDLE).

Note that the previous API we had (which would alloc/schedule/free) does not really have this problem because the lifetime of the work item is handled inside the threadpool.

I think the new API should have the THANDLE(WORK_ITEM) done before we propagate it everywhere. Otherwise, I sense that we would have to revisit again. I think this will remove these race issues we are discussing about.

I understand how changing to a THANDLE needs to change the header of threadpool and the win32 and linux implementation, so I would assume you want to do the change to THANDLE as a separate task - which means that the code here will change again anyway.

linux/src/threadpool_linux.c
```diff
@@ -955,33 +857,21 @@ void threadpool_destroy_work_item(THANDLE(THREADPOOL) threadpool, THREADPOOL_WOR
...
         {
-            /* Codes_SRS_THREADPOOL_LINUX_05_032: [ If sm_exec_begin returns SM_EXEC_REFUSED, threadpool_destroy_work_item shall fail. ]*/
-            LogError("sm_exec_begin failed : %" PRI_MU_ENUM, MU_ENUM_VALUE(SM_RESULT, sm_result));
+            /* Codes_SRS_THREADPOOL_LINUX_05_034: [ When all pending work items finish execution, threadpool_destroy_work_item shall acquire the SRW lock in shared mode by calling srw_lock_acquire_shared. ]*/
+            srw_lock_acquire_shared(threadpool_ptr->srw_lock);
```
dcristoloveanu, line 865: >srw_lock_acquire_shared(threadpool_ptr->srw_lock); [](http://example.com/codeflow?start=12&length=50)

Having a lock over a free does nothing, so let's remove this acquire/release

linux/src/threadpool_linux.c
```diff
@@ -217,29 +194,25 @@ static int threadpool_work_func(void* param)
...
-                            interlocked_decrement(threadpool->task_array[current_index].pending_work_item_count_ptr);
-                            /* Codes_SRS_THREADPOOL_LINUX_05_045: [ threadpool_work_func shall send wake up signal to single listener for the address in pending_work_item_count_ptr. ]*/
-                            wake_by_address_single(threadpool->task_array[current_index].pending_work_item_count_ptr);
-                            /* Codes_SRS_THREADPOOL_LINUX_05_046: [ threadpool_work_func shall release the shared SRW lock by calling srw_lock_release_exclusive. ]*/
+                            if (interlocked_decrement(threadpool->task_array[current_index].pending_work_item_count_ptr) == 0)
+                            {
+                                wake_by_address_single(threadpool->task_array[current_index].pending_work_item_count_ptr);                         
+                            }
                             srw_lock_release_exclusive(threadpool->srw_lock);
                         }
```
dcristoloveanu, line 210: I think this acquire shared/release shared and then acquire exclusive and release exclusive is useless.

In line 199 we evaluate if threadpool->task_array[current_index].pending_work_item_count_ptr is NULL, but after the lock is released that can easily change.

I think this whole dance of pending_work_item_count_ptr is just a poor substitute for having a field here with a THANDLE of the WORK_ITEM. After all that is what we aim at - making sure that a reference is kept to the THANDLE(WORK_ITEM) while it still has to execute.

dcristoloveanu (dismissed):
:shipit:

linux/src/threadpool_linux.c
nishudeshu: We had shared lock and Helgrind was complaining, changing it to exclusive solved the issue since pending_work_item_count_ptr is also initialized in other threads at the index of thread_array for every threadpool_schedule_work_item.

linux/src/threadpool_linux.c
nishudeshu: Hi Dan, yes, it'll be changed later in other task.

linux/src/threadpool_linux.c
nishudeshu: Hi Dan, it won't change in this case since it is assigned in threadpool_schedule_work_item for an index in thread_array and hence not NULL for a given 'index' and won't change after the shared lock is released. When threadpool_schedule_work is used, it will be NULL for the given 'index' and won't change after shared lock is released.

This is just a code switch until threadpool_schedule_,work goes away

linux/src/threadpool_linux.c
nishudeshu: Fixed

linux/src/threadpool_linux.c
nishudeshu: Fixed.

jebrando (approved):
:shipit:

PR #421, opened 2024-12-11
Have the same renaming for threadpool work item reals for Linux as we…

dcristoloveanu:
… have for Windows


mattdurak (approved):
:shipit:

PR #424, opened 2024-12-23
added GBALLOC_LL_TYPE_{ALLOCATOR_TYPE} definition

RohitG28:
added definition for GBALLOC_LL_TYPE_{ALLOCATOR_TYPE} in CMakeLists.txt that will be used in main.cpp to decide if we should enable jemalloc specific logic or not.


dcristoloveanu (approved):
:shipit:

PR #427, opened 2025-01-08
C-PAL: Removal of threadpool_open and threadpool_close

nishudeshu:
Final Removal of threadpool_open and threadpool_close


dcristoloveanu (approved):
:shipit:

PR #431, opened 2025-02-13
add_threadpool_reals

M-iceberg:


linux/src/threadpool_linux.c
```diff
@@ -503,7 +503,7 @@ static void threadpool_timer_dispose(THREADPOOL_TIMER * timer)
     // the callback to use it if we hit this window.
 
     /* Codes_SRS_THREADPOOL_LINUX_07_096: [ threadpool_timer_dispose shall call ThreadAPI_Sleep to allow timer resources to clean up. ]*/
-    ThreadAPI_Sleep(30);
+    ThreadAPI_Sleep(35);
```
jebrando, line 506: >ThreadAPI_Sleep(35); [](http://example.com/codeflow?start=4&length=20)

This is interesting.  I'm guessing it was done because of a failure in one of the int test, which means that sooner or later this test will fail no matter what time you have here.  I'm guessing that you will be asked at some point to change this to something that is deterministic and passes consistently.  You probably should create a task for it if I'm correct.  If I'm wrong then you can just ignore this.

jebrando (dismissed):
:shipit:

linux/src/threadpool_linux.c
dcristoloveanu: Jelani you are correct, this has drawn soooo much attention already :-).

dcristoloveanu, line 506: >35 [](http://example.com/codeflow?start=20&length=2)

I find it very interesting that we had to change from 30 to 35. Did some test fail?

dcristoloveanu: Then why not 40?
Or 45?
Or ...

dcristoloveanu (approved):
:shipit:

linux/src/threadpool_linux.c
darobs: See the comment below, We may have a task for this already.

darobs (approved):

linux/src/threadpool_linux.c
M-iceberg: added a task here for later fix https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

nishudeshu:
Hi @M-iceberg also consider creating real_threadpool_timer_thandle.h and real_threadpool_timer_thandle.c similar to real_threadpool_thandle.h and real_threadpool_thandle.c with this commit.

There is a chance, the tests might fail because of not correctly mocking the THANDLE'ed structures.

dcristoloveanu (approved):
:shipit:

PR #435, opened 2025-02-18
Fix the threadpool_int test

nishudeshu:
Fix threadpool_int merge of WIN32 and LINUX versions, fix chaos_knight_test, add a new test


common/tests/threadpool_int/threadpool_int.c
```diff
@@ -143,6 +147,19 @@ static void wait_work_function(void* context)
...
 
+static int schedule_work_multiple_threads(void* context)
+{
+    THANDLE(THREADPOOL) threadpool = context;
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, InterlockedHL_WaitForValue(&g_start, 1, UINT32_MAX));
+
+    for (size_t i = 0; i < WORK_ITEM_COUNT; i++)
+    {
+        ASSERT_ARE_EQUAL(int, 0, threadpool_schedule_work(threadpool, work_function, (void *)&g_call_count));
+        ThreadAPI_Sleep(1);
```
parth21999, line 158: >ThreadAPI_Sleep(1); [](http://example.com/codeflow?start=8&length=19)

The test I shared with you didn't have this sleep. 

What is the reason behind adding this sleep?

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -1367,61 +1386,61 @@ static int chaos_thread_with_timers_no_lock_and_null_work_item_func(void* contex
 }
 
 //the following 2 tests passed on windows now, but need some fix on linux side for schedule_work_item https://msazure.visualstudio.com/One/_workitems/edit/30570880
```
parth21999, line 1388: >//the following 2 tests passed on windows now, but need some fix on linux side for schedule_work_item [](http://example.com/codeflow?start=0&length=101)

This comment seems to suggest that this test was not working on linux earlier and was commented out for that reason. 

Have we fixed the underlying issue and verified the fix before we re-enable the test?

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -45,6 +46,8 @@ typedef struct WRAP_DATA_TAG
 #define XTEST_FUNCTION(A) void A(void)
```
dcristoloveanu, line 46: This can be removed now, we have DISABLED_TEST_FUNCTION. Not related to the change, but just noticed it and should be easy to remove.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -1100,17 +1117,19 @@ static int chaos_thread_func(void* context)
         {
         case 0:
             //perform a schedule item
-            if (threadpool_schedule_work(chaos_test_data->threadpool, threadpool_task_wait_60_millisec, (void*)&chaos_test_data->executed_work_functions) == 0)
+            if (threadpool_schedule_work(chaos_test_data->threadpool, work_function, (void*)&chaos_test_data->executed_work_functions) == 0)
```
dcristoloveanu, line 1120: >work_function [](http://example.com/codeflow?start=70&length=13)

The previous function seems to also have waited 60 ms.
Is the change to not wait on purpose? Generally I think we don't need not wait when chaos testing, but this makes me wonder if it has some helgrind/valgrind implications.

dcristoloveanu:
> // {

I understand that you did not comment it out originally,  but should this test also be uncommented?

---
Refers to: common/tests/threadpool_int/threadpool_int.c:1447 in 6158340. [](commit_id = 6158340408255ef5e138a5352d6806021a375ea7, deletion_comment = False)

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -1622,4 +1641,40 @@ TEST_FUNCTION(close_while_items_are_scheduled_still_executes_all_items_v2)
...
+
+    THREAD_HANDLE thread_handles[THREAD_COUNT];
+
+    (void)interlocked_exchange(&g_start, 0);
+    (void)interlocked_exchange_64(&g_call_count, 0);
+
+    // act
+    for (size_t i = 0; i < THREAD_COUNT; i++)
+    {
+        ThreadAPI_Create(&thread_handles[i], schedule_work_multiple_threads, (void*)threadpool);
```
dcristoloveanu, line 1661: >ThreadAPI_Create [](http://example.com/codeflow?start=8&length=16)

This should assert

dcristoloveanu (dismissed):
:shipit:

common/tests/threadpool_int/threadpool_int.c
nishudeshu: The issue is for the Linux side of threadpool where the semaphore cannot cope with so fast scheduling in multiple threads since `sem_init` is initialized with the last parameter as 0. Usually, we should pass it the correct number of threads spawned or scheduled for the threadpool. But it is hard to tell on how man threads execution engine will allocate. So the last parameter is initialized to 0. Because of fast scheduling, all the threads are executing critical section of the code and no one is listening to `sem_timedwait()` when `sem_post()` is fired by the scheduler. Since the `value` or the last parameter of `sem_init` is initalized to 0, this `sem_post` is missed resulting in lesser tasks being scheduled on the threadpool. Hence a small `Sleep` or any non trivial operation can slow down this a little bit and the `sem_post` are not missed.
I have added a work around to this which was already there but not used by using the `task_count` field.
@dcristoloveanu @mdurak @darobs @jebrando @anporumb  may want to review this closely.

common/tests/threadpool_int/threadpool_int.c
nishudeshu: Yes, it is fixed. The migration changed the code a little bit that I have fixed.

common/tests/threadpool_int/threadpool_int.c
nishudeshu: Will Remove in the next iteration

common/tests/threadpool_int/threadpool_int.c
nishudeshu: The original `chaos_knight_test` used `work_function`. I think @M-iceberg  tried with `threadpool_task_wait_60_millisec` since `chaos_knight_test` test was failing on Linux. This code was commented, so appears to be trial and error code.

nishudeshu:
> // {

I think @bingyiyang is working on it since its related to timer

---
In reply to: [2667734150](https://github.com/Azure/c-pal/pull/435#issuecomment-2667734150) [](http://example.com/codeflow?ancestors=2667734150)

---
Refers to: common/tests/threadpool_int/threadpool_int.c:1447 in 6158340. [](commit_id = 6158340408255ef5e138a5352d6806021a375ea7, deletion_comment = False)

nishudeshu:
> // {

I may not have full context on whats going on and if this test is even needed.

---
In reply to: [2668368212](https://github.com/Azure/c-pal/pull/435#issuecomment-2668368212) [](http://example.com/codeflow?ancestors=2668368212,2667734150)

---
Refers to: common/tests/threadpool_int/threadpool_int.c:1447 in 6158340. [](commit_id = 6158340408255ef5e138a5352d6806021a375ea7, deletion_comment = False)

common/tests/threadpool_int/threadpool_int.c
parth21999: I think the purpose of this test was to expose this issue with timing/synchronization.

If the way forward is to use `threadpool_work_item_create` and `threadpool_work_item_schedule`, maybe we should update the test to use those APIs and demonstrate that the linux implementation can handle it. 

Or maybe we should disable the test and put a comment that the test should be enabled once we have a fix for the timing issue. 

Putting a sleep here depowers the test and then the test fails to bring attention to an actual issue in the linux implementation.

nishudeshu: True, Linux Threadpool has some issues with its implementation and we know there is a bug somewhere. This is a good test to find a bug. 
But the Linux implementation is not live and we are using only Win32 implementation in Production.
So, our aim to use Linux is to find issues in the code using threadpool. Since Linux doesn't have a Threadpool implementation, we have implemented our own.
The Linux threadpool tries to mimic Win32 threadpool. Although, the code written is very good but it has some bugs.

I feel, as long as a Sleep solves the issue on the Linux side, perhaps Sleep only when the platform is Linux and do not sleep when its Win32. This way, the test need not be disabled.

Hopefully, the current implementation solves the race condition.

parth21999: > Sleep only when the platform is Linux and do not sleep when its Win32

I like this. We should do this and add a comment that the sleep can be removed when the linked work item is complete

nishudeshu: @parth21999 So, I have removed the Sleep all together and fixed the Semaphore issue (I hope). Only for Linux Helgrind and Valgrind, we are adding Sleep(0) for Yielding the thread and avoid timeout issues we have reduced the number of items and number of threads into half. We yield few times in the loop of each thread. On Linux, when the code runs without Helgrind and Valgrind, it runs with same parameters as Win32.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -143,6 +152,44 @@ static void wait_work_function(void* context)
     wake_by_address_single_64(&wait_work_context->call_count);
 }
 
+static size_t get_work_per_thread()
+{
+    size_t work_per_thread = WORK_PER_THREAD;
+#ifdef USE_VALGRIND
+    // Check if we are being invoked from Valgrind to ensure that
+    // the non-Valgrind and Helgrind run do not Yield
```
dcristoloveanu, line 160: >Yield [](http://example.com/codeflow?start=48&length=5)

There is no yielding in this function, sorry, but this is a misleading comment.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -143,6 +152,44 @@ static void wait_work_function(void* context)
...
+    if (RUNNING_ON_VALGRIND)
+    {
+        // Yield
```
dcristoloveanu, line 163: I am not convinced we get a lot of value from checking whether we are running under valgrind/helgrind vs relying on the USE_VALGRIND compile switch.
Feels like we just make the check a little bit more complex.

dcristoloveanu: I can live with having the check also, just noticed that we're adding these checks for not so much value.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -143,6 +152,44 @@ static void wait_work_function(void* context)
...
+        work_per_thread = WORK_PER_THREAD / 10;
+    }
+#endif
+    return work_per_thread;
+}
+
+static int schedule_work_multiple_threads(void* context)
```
dcristoloveanu, line 170: >schedule_work_multiple_threads [](http://example.com/codeflow?start=11&length=30)

I don't understand what does this function have to do with multiple threads.
It is called from one thread (the main thread that executes the test).
It schedules a number of items to execute.
So where is the multiple threads?

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -143,6 +152,44 @@ static void wait_work_function(void* context)
...
+
+#ifdef USE_VALGRIND
+        // Check if we are being invoked from Valgrind to ensure that
+        // the non-Valgrind and Helgrind run do not Yield
+        if (RUNNING_ON_VALGRIND)
+        {
+            // Yield
+            ThreadAPI_Sleep(0);
+        }
+#endif
```
dcristoloveanu, line 188: The way I read the test it fires a number of items that are scheduled to be executed in the threadpool.
Like schedule 1M items.

That is all.
So why is the yield actually needed?

I am somewhat confused here

dcristoloveanu: Ok, I misread the test initially, I see we're starting some threads and we are scheduling N items from these threads.

I guess yielding might be needed in order to get to execute the threads that are used by the threadpool.

dcristoloveanu (changes requested):
:clock1:

common/tests/threadpool_int/threadpool_int.c
nishudeshu: This was typo or remnant from copy paste from below. There was no intention to mislead.

common/tests/threadpool_int/threadpool_int.c
nishudeshu: I found that the USE_VALGRIND applies to non Valgrind/Helgrind normal runs also. Hence I added this check of RUNNING_ON_VALGRIND. USE_VALGRIND only shields in Windows compiled code.

linux/src/threadpool_linux.c
```diff
@@ -140,7 +140,7 @@ static int threadpool_work_func(void* param)
                 ts.tv_nsec += TP_SEMAPHORE_TIMEOUT_MS;
 
                 /* Codes_SRS_THREADPOOL_LINUX_07_075: [ threadpool_work_func shall wait on the semaphore with a time limit. ]*/
-                if (sem_timedwait(&threadpool->semaphore, &ts) != 0)
+                if (interlocked_add_64(&threadpool->task_count, 0) == 0 && sem_timedwait(&threadpool->semaphore, &ts) != 0)
```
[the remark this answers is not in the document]
nishudeshu: If it's `TRUE`, it goes to `sem_timedwait` and waits for 100 ms and loops back if `sem_timedwait` times out. If `sem_timedwait` returns `zero`, we have work and jump to `else` to do the work.
If it's `FALSE', it means we have data to process and goes into the `else` and does the work.
So, I don't see a spin here.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -128,11 +136,12 @@ static void threadpool_long_task_v2(void* context)
 
 static void work_function(void* context)
 {
-    (void) context;
-    (void)interlocked_increment_64(&g_call_count);
-    wake_by_address_single_64(&g_call_count);
+    volatile_atomic int64_t* call_count = (volatile_atomic int64_t*)context;
+    (void)interlocked_increment_64(call_count);
+    wake_by_address_single_64((void *)call_count);
```
parth21999, line 141: >(void *) [](http://example.com/codeflow?start=30&length=8)

Is this cast needed?

nishudeshu: I just restored the old code

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -1622,4 +1668,46 @@ TEST_FUNCTION(close_while_items_are_scheduled_still_executes_all_items_v2)
...
+
+    ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, InterlockedHL_SetAndWakeAll(&g_start, 1));
+
+    for (size_t i = 0; i < THREAD_COUNT; i++)
+    {
+        int thread_result;
+        ASSERT_ARE_EQUAL(THREADAPI_RESULT, THREADAPI_OK, ThreadAPI_Join(thread_handles[i], &thread_result));
+        ASSERT_ARE_EQUAL(int, 0, thread_result);
+    }
+    LogInfo("Call Count before comparison %" PRId64 "", interlocked_add_64(&multi_thread_call_count, 0));
```
parth21999, line 1703: >LogInfo("Call Count before comparison %" PRId64 "", interlocked_add_64(&multi_thread_call_count, 0)); [](http://example.com/codeflow?start=4&length=101)

I checked out to your branch, and ran the test using shell script many times.

The only change I made to the script was change the timeout on line 1704 to `45000`.

linux/src/threadpool_linux.c
parth21999, line 143: >interlocked_add_64(&threadpool->task_count, 0) [](http://example.com/codeflow?start=20&length=46)

Shouldn't this change lead to spec+test changes?

common/tests/threadpool_int/threadpool_int.c
nishudeshu: It's a known issue and I have spoken to @jebrando about it and it has something to do with reallocation of queue array when number of tasks exceeds initial buffer allocation. 

So you can create a new story for this. It may happen due to the parallel scheduling and the reallocation logic may sometimes give this issue.

linux/src/threadpool_linux.c
nishudeshu: It might become too much implementation detailed

common/tests/threadpool_int/threadpool_int.c
nishudeshu: Disabling the test and fixing the original story of uncommenting the tests.

CMakeLists.txt
```diff
@@ -81,13 +84,19 @@ set(original_run_int_tests ${run_int_tests})
...
+set(original_run_helgrind ${run_helgrind})
+set(original_run_drd ${run_drd})
 
 set(run_e2e_tests OFF)
 set(run_unittests OFF)
 set(run_int_tests OFF)
 set(run_traceability OFF)
 set(run_perf_tests OFF)
 set(run_reals_check OFF)
+set(run_run_valgrind OFF)
```
dcristoloveanu, line 97: >run_run_valgrind [](http://example.com/codeflow?start=4&length=16)

I bet this does not work :-)

CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ option(run_perf_tests "set run_perf_tests to ON to run perf tests (default is OF
 option(use_cppunittest "set use_cppunittest to ON to build CppUnitTest tests on Windows (default is OFF)" OFF)
 option(run_traceability "run traceability tool (default is ON)" ON)
 option(run_reals_check "set run_reals_check to ON to run reals check (default is OFF)." OFF)
+option(run_valgrind "set run_valgrind to ON when running the program under Valgrind (Turn ON under Linux only) (default is OFF)." OFF)
+option(run_helgrind "set run_helgrind to ON when running the program under Helgrind (Turn ON under Linux only) (default is OFF)." OFF)
+option(run_drd "set run_drd to ON when running the program under DRD (Turn ON under Linux only) (default is OFF)." OFF)
```
dcristoloveanu, line 32: Why do we want to add these?

dcristoloveanu (approved):
:shipit:

CMakeLists.txt
nishudeshu: Fixed

CMakeLists.txt
nishudeshu: Just to set the default to OFF so that windows do not detect valgrind.h

dcristoloveanu (approved):
:shipit:

nishudeshu:
https://msazure.visualstudio.com/One/_workitems/edit/30570880

PR #437, opened 2025-02-20
Update devops_gated.yml for Azure Pipelines

nishudeshu:
WIN build for some reason is still seeing the USE_VALGRIND option. Disabling use of Valgrind for WIN runs by using -Drun_valgrind:bool=OFF -Drun_helgrind:bool=OFF -Drun_drd:bool=OFF


build/devops_gated.yml
```diff
@@ -12,7 +12,7 @@ resources:
 jobs:
 - template: /pipeline_templates/build_all_flavors.yml@c_build_tools
   parameters:
-    cmake_options: -Drun_unittests:BOOL=ON -Drun_e2e_tests:BOOL=ON -Drun_int_tests:BOOL=ON -Drun_perf_tests:BOOL=ON -Drun_traceability:BOOL=ON -Duse_cppunittest:BOOL=ON -Drun_reals_check:BOOL=ON
+    cmake_options: -Drun_unittests:BOOL=ON -Drun_e2e_tests:BOOL=ON -Drun_int_tests:BOOL=ON -Drun_perf_tests:BOOL=ON -Drun_traceability:BOOL=ON -Duse_cppunittest:BOOL=ON -Drun_reals_check:BOOL=ON -Drun_valgrind:bool=OFF -Drun_helgrind:bool=OFF -Drun_drd:bool=OFF
```
dcristoloveanu, line 15: >-Drun_valgrind:bool=OFF -Drun_helgrind:bool=OFF -Drun_drd:bool=OFF [](http://example.com/codeflow?start=195&length=66)

We don't use these flags on Windows, so why would it fail on Windows?

dcristoloveanu (approved):
:shipit:

build/devops_gated.yml
nishudeshu: The `USE_VALGRIND` flag was set to 1 for some reason and it appears that all these flags passed to `cmake` should be turned `OFF` at the startup. I have fixed this by making the default of these flags as `OFF`. It gets enabled in `build.sh` of Linux Build.

dcristoloveanu:
@nishudeshu I believe this PR is no longer needed, closing.

PR #441, opened 2025-02-28
add code fix timer dispose race

M-iceberg:


common/tests/threadpool_int/threadpool_int.c
```diff
@@ -409,7 +410,7 @@ TEST_FUNCTION(MU_C3(scheduling_, N_WORK_ITEMS, _work_with_pool_threads))
     execution_engine_dec_ref(execution_engine);
 }
 
-#define N_THREADPOOL_TIMERS 100
+#define N_THREADPOOL_TIMERS 1
```
jebrando: Do we want to have this as 1?  It seems like we'd want a least a few threads here.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -409,7 +410,7 @@ TEST_FUNCTION(MU_C3(scheduling_, N_WORK_ITEMS, _work_with_pool_threads))
...
 
 TEST_FUNCTION(MU_C3(starting_, N_THREADPOOL_TIMERS, _timer_start_runs_once))
```
jebrando: >starting_, N_THREADPOOL_TIMERS, _timer_start_runs_once) [](http://example.com/codeflow?start=20&length=55)

I know you didn't do this, but it seems that this name is misleading, it should just say something about 1 timer to run once.

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -843,7 +845,7 @@ TEST_FUNCTION(timer_cancel_restart_works_runs_periodically)
     ASSERT_IS_NOT_NULL(timer);
 
     // Timer should run 4 times in about 2.1 seconds
```
jebrando: >// Timer should run 4 times in about 2.1 seconds [](http://example.com/codeflow?start=4&length=48)

These comments should be fixed.  Here and on line 859

linux/src/threadpool_linux.c
```diff
@@ -97,6 +98,31 @@ typedef struct THREADPOOL_TAG
 
 THANDLE_TYPE_DEFINE(THREADPOOL);
 
+#define TIMER_TABLE_SIZE 100
```
jebrando, line 101: Looking at the changes it seems like we can only have 100 times at any one time.  What happens when we go over the 100 timer mark (which can happen really quick in the software)?  Do we have any failure mechanism as in do we fail the timer create?  Should we?

nishudeshu:
Branch might need to be updated with the latest changes from @dcristoloveanu

linux/src/threadpool_linux.c
```diff
@@ -109,8 +136,18 @@ static void on_timer_callback(sigval_t timer_data)
...
-        timer_instance->work_function(timer_instance->work_function_ctx);
+        /* Codes_SRS_THREADPOOL_LINUX_07_100: [ on_timer_callback shall acquire the timer table exclusive lock.]*/
+        srw_lock_ll_acquire_exclusive(&timer_table_lock);
+        for(uint32_t i = 0; i < TIMER_TABLE_SIZE; i++)
+        {
+            if(timer_table[i] == timer_instance)
+            {
+                /* Codes_SRS_THREADPOOL_LINUX_45_004: [ If timer exists, on_timer_callback shall call the timer's work_function with work_function_ctx. ]*/
+                timer_instance->work_function(timer_instance->work_function_ctx);
+            }
```
nishudeshu, line 147: Should we do `time_table[i] = NULL` between these two lines?

`timer_instance->work_function(timer_instance->work_function_ctx);`
`time_table[i] = NULL`

linux/src/threadpool_linux.c
```diff
@@ -109,8 +136,18 @@ static void on_timer_callback(sigval_t timer_data)
     }
     else
     {
-        /* Codes_SRS_THREADPOOL_LINUX_45_004: [ on_timer_callback shall call the timer's work_function with work_function_ctx. ]*/
-        timer_instance->work_function(timer_instance->work_function_ctx);
+        /* Codes_SRS_THREADPOOL_LINUX_07_100: [ on_timer_callback shall acquire the timer table exclusive lock.]*/
+        srw_lock_ll_acquire_exclusive(&timer_table_lock);
+        for(uint32_t i = 0; i < TIMER_TABLE_SIZE; i++)
```
nishudeshu, line 141: >for( [](http://example.com/codeflow?start=8&length=4)

Add space after `for`

nishudeshu:
>     if (timer_delete(timer->time_id) != 0)

Add a `timer == NULL` check in case someone accidentally calls this directly without using THANDLE malloc and pass this function as a callback.

---
Refers to: linux/src/threadpool_linux.c:530 in fe3aacd. [](commit_id = fe3aacdfda7c89f1c95dc59e54b3367d410f239f, deletion_comment = False)

linux/src/threadpool_linux.c
```diff
@@ -498,17 +535,19 @@ static void threadpool_timer_dispose(THREADPOOL_TIMER * timer)
...
-    // To reproduce this problem, create a large number of threads that delete the timer at the same time it is due to expire and remove the pause
-    // above.
+
+    /* Codes_SRS_THREADPOOL_LINUX_07_102: [ threadpool_timer_dispose shall acquire the timer table exclusive lock. ]*/
+    srw_lock_ll_acquire_exclusive(&timer_table_lock);
+    for(uint32_t i = 0; i < TIMER_TABLE_SIZE; i++)
+    {
+        if(timer_table[i] == timer)
+        {
+            timer_table[i] = NULL;
```
nishudeshu, line 545: I think this should be done in `on_timer_callback` also

linux/src/threadpool_linux.c
```diff
@@ -543,7 +582,13 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
             if (srw_lock_ll_init(&result->timer_lock) != 0)
```
nishudeshu, line 582: >srw_lock_ll_init [](http://example.com/codeflow?start=16&length=16)

Might be useful to print the return value of the init

linux/src/threadpool_linux.c
```diff
@@ -543,7 +582,13 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
...
             {
                 /* Codes_SRS_THREADPOOL_LINUX_07_060: [ If any error occurs, threadpool_timer_start shall fail and return NULL. ]*/
-                LogError("srw_lock_ll_init failed");
+                LogError("srw_lock_ll_init failed, result->timer_lock = %p.", &result->timer_lock);
+            }
+            /* Codes_SRS_THREADPOOL_LINUX_07_096: [ threadpool_timer_start shall call lazy_init with do_init as initialization function.  ]*/
+            else if(lazy_init(&g_lazy, do_init, (void*)&timer_table_lock) != LAZY_INIT_OK)
```
nishudeshu, line 588: >lazy_init [](http://example.com/codeflow?start=20&length=9)

Try to capture and log the return value

linux/src/threadpool_linux.c
```diff
@@ -580,7 +625,22 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
                         /* Codes_SRS_THREADPOOL_LINUX_07_061: [ threadpool_timer_start shall return and allocated handle in timer_handle. ]*/
                         /* Codes_SRS_THREADPOOL_LINUX_07_062: [ threadpool_timer_start shall succeed and return a non-NULL handle. ]*/
                         result->time_id = time_id;
-                        goto all_ok;
+
+                        srw_lock_ll_acquire_exclusive(&timer_table_lock);
+                        for(uint32_t i = 0; i < TIMER_TABLE_SIZE; i++)
```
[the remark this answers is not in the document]
M-iceberg: modified

common/tests/threadpool_int/threadpool_int.c
M-iceberg: done

common/tests/threadpool_int/threadpool_int.c
M-iceberg: done

linux/src/threadpool_linux.c
M-iceberg: done

linux/src/threadpool_linux.c
M-iceberg: the timer is not deleted in here yet.

linux/src/threadpool_linux.c
M-iceberg: on_timer_callback doesn't delete the timer from the table

linux/src/threadpool_linux.c
M-iceberg: checked all the cases in the repo and I think check for if return value is not 0 is enough here

linux/src/threadpool_linux.c
M-iceberg: lazy_init can only return LAZY_INIT_OK or LAZY_INIT_ERROR, no need to print the return result if it's not equal to LAZY_INIT_OK here

M-iceberg:
>     if (timer_delete(timer->time_id) != 0)

dispose should not be used directly in that way 

---
In reply to: [2691704576](https://github.com/Azure/c-pal/pull/441#issuecomment-2691704576) [](http://example.com/codeflow?ancestors=2691704576)

---
Refers to: linux/src/threadpool_linux.c:530 in fe3aacd. [](commit_id = fe3aacdfda7c89f1c95dc59e54b3367d410f239f, deletion_comment = False)

linux/src/threadpool_linux.c
M-iceberg: done increased the size of timer table and added retry for 2 times as discussed : )

linux/src/threadpool_linux.c
```diff
@@ -580,7 +626,31 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
...
+                                timer_table[i] = result;
+                                srw_lock_ll_release_exclusive(&timer_table_lock);
+                                goto all_ok;
+                            }
+                            else if (i == TIMER_TABLE_SIZE - 1)
+                            {
+                                if (retry_times < MAX_TIMER_START_RETRY)
+                                {
+                                    ThreadAPI_Sleep(10);
+                                    i = 0;
```
dcristoloveanu, line 645: This retry does not really solve anything, does it?
If we're out of timers, we're out of timers, we're probably pushing the system to what it can do and if we are pushing it it will fail after 1 or 2 retries also.

I feel having retries actually muddies the implementation because  it tries to hide some errors but in a rather unpredictable way ...

linux/src/threadpool_linux.c
```diff
@@ -580,7 +626,31 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
...
+                        {
+                            if (timer_table[i] == NULL)
+                            {
+                                timer_table[i] = result;
+                                srw_lock_ll_release_exclusive(&timer_table_lock);
+                                goto all_ok;
+                            }
+                            else if (i == TIMER_TABLE_SIZE - 1)
+                            {
+                                if (retry_times < MAX_TIMER_START_RETRY)
```
dcristoloveanu, line 642: All this code has no specs

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -842,7 +843,7 @@ TEST_FUNCTION(timer_cancel_restart_works_runs_periodically)
     THANDLE(THREADPOOL_TIMER) timer = threadpool_timer_start(threadpool, 100, 500, work_function, (void*)&g_call_count);
     ASSERT_IS_NOT_NULL(timer);
 
-    // Timer should run 4 times in about 2.1 seconds
+    // Timer should run 4 times in about 3 seconds
     ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, InterlockedHL_WaitForValue64(&g_call_count, 4, 3000));
     LogInfo("Timer completed 4 times");
```
dcristoloveanu, line 848: I have to say I don't like this kind of test.
It is brittle. Something will delay the system at some point and lo and behold, next we have someone debugging a flakiness.

Let's create a task under things that never get done for fixing it...

dcristoloveanu (changes requested):
:clock1:

dcristoloveanu:
I kind of like this change :-). Since it kills that horrible Sleep we had.
I have 2 slightly bigger points:
- the retry is not needed
- there are no specs for some of the added code

common/tests/threadpool_int/threadpool_int.c
M-iceberg: created the task : [https://msazure.visualstudio.com/One/_workitems/edit/31672710](https://msazure.visualstudio.com/One/_workitems/edit/31672710)

linux/src/threadpool_linux.c
M-iceberg: removed the retry

linux/src/threadpool_linux.c
M-iceberg: added

linux/devdoc/threadpool_linux_requirements.md
```diff
@@ -213,6 +213,14 @@ MOCKABLE_FUNCTION(, THANDLE(THREADPOOL_TIMER), threadpool_timer_start, THANDLE(T
 
 **SRS_THREADPOOL_LINUX_07_091: [** `threadpool_timer_start` shall initialize the lock guarding the timer state by calling `srw_lock_ll_init`. **]**
 
+**SRS_THREADPOOL_LINUX_07_096: [** threadpool_timer_start shall call `lazy_init` with `do_init` as initialization function.  **]**
```
dcristoloveanu, line 216: >`do_init` [](http://example.com/codeflow?start=86&length=9)

Consider putting the do_init signature right after this function (ln 239)
And having the specs for do_init placed there too.

/consider

M-iceberg: done

jebrando (dismissed):
:shipit:

common/tests/threadpool_int/threadpool_int.c
```diff
@@ -411,7 +411,7 @@ TEST_FUNCTION(MU_C3(scheduling_, N_WORK_ITEMS, _work_with_pool_threads))
 
 #define N_THREADPOOL_TIMERS 100
 
-TEST_FUNCTION(MU_C3(starting_, N_THREADPOOL_TIMERS, _timer_start_runs_once))
+TEST_FUNCTION(starting_one_timer_runs_once)
```
dcristoloveanu, line 414: >starting_one_timer_runs_once [](http://example.com/codeflow?start=14&length=28)

The test we had was a bit stronger, right?

dcristoloveanu (dismissed):
:shipit:

common/tests/threadpool_int/threadpool_int.c
M-iceberg: done

nishudeshu:
> // TEST_FUNCTION(chaos_knight_test_with_timers_no_lock)

Did you try uncommenting this one?

---
Refers to: common/tests/threadpool_int/threadpool_int.c:1408 in dd4baf2. [](commit_id = dd4baf2582f312e790fc1746b2e30201c7239117, deletion_comment = False)

linux/src/threadpool_linux.c
```diff
@@ -580,7 +625,25 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
...
+                                srw_lock_ll_release_exclusive(&timer_table_lock);
+                                goto all_ok;
+                            }
+                            else if (i == TIMER_TABLE_SIZE - 1)
+                            {
+                                /* Codes_SRS_THREADPOOL_LINUX_07_060: [ If any error occurs, threadpool_timer_start shall fail and return NULL. ]*/
+                                LogError("No timers available");
+                                srw_lock_ll_release_exclusive(&timer_table_lock);
+                            }
+                        }
```
nishudeshu, line 646: You can modify this block as:
```
srw_lock_ll_acquire_exclusive(&timer_table_lock);
for (uint32_t i = 0; i < TIMER_TABLE_SIZE; i++)
{
        if (timer_table[i] == NULL)
         {
                /* Codes_SRS_THREADPOOL_LINUX_07_106:[ threadpool_timer_start shall add the new timer to the timer table and release the exclusive lock. ]*/
                      timer_table[i] = result;
                       srw_lock_ll_release_exclusive(&timer_table_lock);
                       goto all_ok;
                      }
          }
          else
          {
                 LogError("No timers available");
          }
}
/* Codes_SRS_THREADPOOL_LINUX_07_060: [ If any error occurs, threadpool_timer_start shall fail and return NULL. ]*/
srw_lock_ll_release_exclusive(&timer_table_lock);
```

linux/src/threadpool_linux.c
```diff
@@ -580,7 +625,25 @@ THANDLE(THREADPOOL_TIMER) threadpool_timer_start(THANDLE(THREADPOOL) threadpool,
                         /* Codes_SRS_THREADPOOL_LINUX_07_061: [ threadpool_timer_start shall return and allocated handle in timer_handle. ]*/
                         /* Codes_SRS_THREADPOOL_LINUX_07_062: [ threadpool_timer_start shall succeed and return a non-NULL handle. ]*/
                         result->time_id = time_id;
-                        goto all_ok;
+
+                        /* Codes_SRS_THREADPOOL_LINUX_07_105: [ threadpool_timer_start shall acquire the exclusive lock for the timer table. ]*/
```
nishudeshu, line 629: > Codes_SRS_THREADPOOL_L [](http://example.com/codeflow?start=26&length=23)

Better to implement timer_table as a circular buffer or use tqueue

dcristoloveanu (approved):
:shipit:

PR #444, opened 2025-04-02
Nisdeshmukh/uncomment threadpool test

nishudeshu:
Uncommenting chaos_knight_test

### DIFF
--- a/common/tests/threadpool_int/threadpool_int.c
+++ b/common/tests/threadpool_int/threadpool_int.c
@@ -128,9 +128,9 @@ static void threadpool_long_task_v2(void* context)
 
 static void work_function(void* context)
 {
-    (void) context;
-    (void)interlocked_increment_64(&g_call_count);
-    wake_by_address_single_64(&g_call_count);
+    volatile_atomic int64_t* call_count = (volatile_atomic int64_t*)context;
+    (void)interlocked_increment_64(call_count);
+    wake_by_address_single_64((void*)call_count);
 }
 
 static void wait_work_function(void* context)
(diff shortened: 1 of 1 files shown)


common/tests/threadpool_int/threadpool_int.c
```diff
@@ -1347,62 +1349,64 @@ static int chaos_thread_with_timers_no_lock_and_null_work_item_func(void* contex
     return 0;
 }
 
-//the following 2 tests passed on windows now, but need some fix on linux side for schedule_work_item https://msazure.visualstudio.com/One/_workitems/edit/30570880
-// TEST_FUNCTION(chaos_knight_test)
```
mattdurak, line 1351: >//  [](http://example.com/codeflow?start=0&length=3)

Just a general note, in the future let's comment out tests with pre-processor "if 0" to make the diff nicer when we add the test back

mattdurak (approved):
:shipit:

common/tests/threadpool_int/threadpool_int.c
jebrando: Or you can use the DISABLED_TEST_FUNCTION macro, then the test will still get compiled, just not run

jebrando (approved):
:shipit:

common/tests/threadpool_int/threadpool_int.c
nishudeshu: @mattdurak  @jebrando Thanks, I had used `DISABLED_TEST_FUNCTION` elsewhere in the past. This was commented by Bingyi when she combined WIN32 and LINUX integ tests.

PR #455, opened 2025-04-29
Task 32555218 - object helper spec and header changes

shrjadhav-msft:
Updated object_helper spec and header files to include new functions for creating and getting job objects by name and a function for assigning process handle to job object.


win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
 **SRS_JOB_OBJECT_HELPER_18_019: [** `job_object_helper_create` shall succeed and return the `JOB_OBJECT_HELPER` object.  **]**
 
 
+## job_object_helper_create_with_name
+```c
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create_with_name, const char*, job_name);
+```
+``job_object_helper_create_with_name`` is used to create a Job Object with the specified name.
+
+**SRS_JOB_OBJECT_HELPER_19_001: [** `job_object_helper_create_with_name` shall log the error and return `NULL` if the passed `job_name` is `NULL` or empty. **]**
```
anporumb, line 42: >shall log the error an [](http://example.com/codeflow?start=73&length=22)

Generally there's no expectation to have "shall log the error".

This is because it would imply writing a STRICT_EXPECTED_CALL(whaverLogFunction...) and we'd be then overtesting the logging part.

Another reason is logging is used itself in the mocking framework and you'd have problems due to "this is the real logging done by the code under test" and "this is the logging due to printing the test name on the screen". In our stack logging lives just on top the macro layer and way below testing.

Even assuming we'd have some way of testing the logging (note: we do test other telemetry components, such as metrics) in general the assumption is "logging can fail at any time for whatever reason" and then testing it would be pretty low ROI.

Another reason is that in the established folklore of the people who will be doing the code reviews there's an established "hunt for missing logs" mind set which will not allow a PR to be aprroved it there can more logging be done.

So here and everywhere: remove "shall log".

dcristoloveanu, line 42: >log the error [](http://example.com/codeflow?start=79&length=13)

We generally do not state explicitly the fact that we log errors/info/etc. We thought it would be ... too much if we do that

dcristoloveanu, line 42: >if the passed `job_name` is `NULL` or empty [](http://example.com/codeflow?start=111&length=43)

Just a nitpick as you settle in the style of writing these .mds, we usually state the if first, like
"If job_name is NULL then ..."

It makes it easier to follow when reading it IMO.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
 **SRS_JOB_OBJECT_HELPER_18_019: [** `job_object_helper_create` shall succeed and return the `JOB_OBJECT_HELPER` object.  **]**
 
 
+## job_object_helper_create_with_name
+```c
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create_with_name, const char*, job_name);
+```
+``job_object_helper_create_with_name`` is used to create a Job Object with the specified name.
```
dcristoloveanu, line 40: >`` [](http://example.com/codeflow?start=0&length=2)

Probably only one backtick needed here

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+**SRS_JOB_OBJECT_HELPER_19_001: [** `job_object_helper_create_with_name` shall log the error and return `NULL` if the passed `job_name` is `NULL` or empty. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_002: [** `job_object_helper_create_with_name` shall allocate `JOB_OBJECT_HELPER` object. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_016: [** `job_object_helper_create_with_name` shall call `mbs_to_wcs` to convert job_name to `wchar_t*`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_003: [** `job_object_helper_create_with_name` shall call `CreateJobObject` to create a new job object passing `NULL` for `lpJobAttributes` and `job_name` to `lpName`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_004: [** `job_object_helper_create_with_name` shall log the error and return `NULL` if the `CreateJobObject` fail. **]**
```
dcristoloveanu, line 50: >`job_object_helper_create_with_name` shall log the error and return `NULL` if the `CreateJobObject` fail. [](http://example.com/codeflow?start=36&length=105)

We usually write one catch all spec like 
"If any error occurrs, then blaahblah shall fail and return NULL"

Makes specs easier to read.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+**SRS_JOB_OBJECT_HELPER_19_011: [** `job_object_helper_assign_process` shall call `OpenJobObject` with Permission `JOB_OBJECT_ASSIGN_PROCESS`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_018: [** `job_object_helper_assign_process` shall log the error and return `MU_FAILURE` if `OpenJobObject` call fails. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_013: [** `job_object_helper_assign_process` shall call `AssignProcessToJobObject` to assign the current process to the job object. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_014: [** `job_object_helper_assign_process` shall log error and return `MU_FAILURE_` if the `AssignProcessToJobObject` fails. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_015: [** `job_object_helper_assign_process` shall return `0` if the `AssignProcessToJobObject` succeeds. **]**
```
dcristoloveanu, line 92: >`job_object_helper_assign_process` shall return `0` if the `AssignProcessToJobObject` succeeds. [](http://example.com/codeflow?start=36&length=95)

Consider phrasing like "job_object_helper_assign_process shall succeed and return 0." Then you don't have to change the spec if one more step is added at a later point between AssignProcessToJobObject and returning.

win32/devdoc/job_object_helper.md
anporumb, line 50: >if the `CreateJobObject` fail [](http://example.com/codeflow?start=111&length=29)

Usually we tend to have a "catch all" clause for error. "if anything goes bad, return error". This is because in the greatest majority of time our API calls are transactional: either they happened fully (and some action has been performed/started, some object is created, some transition happened) or the system behaves as if the call was not performed.

For example line 31 or 120 of this file.

It would also get tiring very fast to write after every thing that could fail "If blablablabla call fails then return a non-zero value". And then 3 lines of text after, the same "if balbalbalba2 call fails then return a non-zero value". 

For example, between lines 44 and 46 there should have been a req that would read like: "if allocating the JOB_OBJECT_HELPER object fails then fail and return NULL". => Too many of these specs would be written.

This is also why we write the "catch all clause" with "shall fail and return a non-zero value" because later, in some other module, we can rely on "if any failure ... then fail".

==

So here "shall fail" is missing.

Also change the requirement to read "if there are any failures..."

==

Note: We have amazing support to test "if any failure..." in umock_c so there's that as a help. Also if we'd have individual req for every failure... then we'd need to have individual tests for every requirement. Boring and low ROI.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+**SRS_JOB_OBJECT_HELPER_19_006: [** `job_object_helper_get` shall log the error and return `NULL` if the passed `job_name` is `NULL` or empty. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_017: [** `job_object_helper_get` shall call `mbs_to_wcs` to convert the `job_name` to `wchar_t*` type. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_007: [** `job_object_helper_get` shall call `OpenJobObject` with name parameter set to `job_name`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_008: [** `job_object_helper_get` shall log the error and return `NULL` if the `job object` with the name does not exist. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_009: [** `job_object_helper_get`shall return the `JOB_OBJECT_HELPER` object when the job object is opened successfully. **]**
```
anporumb, line 69: >`JOB_OBJECT_HELPER` object when the job object is opened successfully. [](http://example.com/codeflow?start=76&length=70)

This is very weird and I don't think it can work.

So assume 2 calls are previously done:
1) `job_object_helper_create_with_name("IONESCU")`
2) `job_object_helper_create_with_name("POPESCU")`

Now a call to `job_object_helper_get` is made with "IONESCU". 

Windows will return the `HANDLE` for IONESCU, but "our" object (THANDLE(JOB_OBJECT_HELPER))... how is it retrieved? How does the software know how to look for HANDLE(of IONESCU) => THANDLE(JOB_OBJECT_HELPER) in the absence of some global data... tracking all the strings<===>THANDLE(JOB_OBJECT_HELPER)?

==

So either I don't understand something basic, or there's missing requirements, or it cannot work.

anporumb: Probably: more explanations are required.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+**SRS_JOB_OBJECT_HELPER_19_003: [** `job_object_helper_create_with_name` shall call `CreateJobObject` to create a new job object passing `NULL` for `lpJobAttributes` and `job_name` to `lpName`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_004: [** `job_object_helper_create_with_name` shall log the error and return `NULL` if the `CreateJobObject` fail. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_005: [** `job_object_helper_create_with_name` shall return the `JOB_OBJECT_HELPER` object when the job object is created successfully. **]**
+
+
+## job_object_helper_get
+```c
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_get, const char*, job_name);
```
anporumb, line 57: >job_object_helper_get [](http://example.com/codeflow?start=48&length=21)

is this function intended to return some of the handles created by `job_object_helper_create` (note: the version without _name").

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+```
+``job_object_helper_get`` is used for getting a `JOB_OBJECT_HELPER` created with the name `job_name`.
+
+**SRS_JOB_OBJECT_HELPER_19_006: [** `job_object_helper_get` shall log the error and return `NULL` if the passed `job_name` is `NULL` or empty. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_017: [** `job_object_helper_get` shall call `mbs_to_wcs` to convert the `job_name` to `wchar_t*` type. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_007: [** `job_object_helper_get` shall call `OpenJobObject` with name parameter set to `job_name`. **]**
```
anporumb, line 65: >name [](http://example.com/codeflow?start=92&length=4)

`name` parameter is actually called `lpName` in `[https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-openjobobjecta`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-openjobobjecta`).

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+```
+``job_object_helper_create_with_name`` is used to create a Job Object with the specified name.
+
+**SRS_JOB_OBJECT_HELPER_19_001: [** `job_object_helper_create_with_name` shall log the error and return `NULL` if the passed `job_name` is `NULL` or empty. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_002: [** `job_object_helper_create_with_name` shall allocate `JOB_OBJECT_HELPER` object. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_016: [** `job_object_helper_create_with_name` shall call `mbs_to_wcs` to convert job_name to `wchar_t*`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_003: [** `job_object_helper_create_with_name` shall call `CreateJobObject` to create a new job object passing `NULL` for `lpJobAttributes` and `job_name` to `lpName`. **]**
```
anporumb, line 48: >CreateJobObject [](http://example.com/codeflow?start=85&length=15)

I am going to add here that CReateObject is either CreateObjectA or CreateObjectW, depending on how our strings come to us.

In this case, it looks like either
1) a case of CreateObjectA and then `mbs_to_wcs` is not needed.
2) a case of CreateObjectW and then `mbs_to_wcs` is needed.

In most of the cases... cough... we've been looking the other way about wchar_t* "stuff" and the strings used internally in the project are all nice narrow strings.

anporumb: Same observation goes for OpenJobObject and all APIs that take a "name".

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+**SRS_JOB_OBJECT_HELPER_19_009: [** `job_object_helper_get`shall return the `JOB_OBJECT_HELPER` object when the job object is opened successfully. **]**
+
+
+## job_object_helper_assign_process
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_assign_process, THANDLE(JOB_OBJECT_HELPER), job_object_helper, THANDLE(PROCESS_HANDLE), process_hndl);
+```
+``job_object_helper_assign_process`` assigns the process to the Job Object.
+
+**SRS_JOB_OBJECT_HELPER_19_010: [** `job_object_helper_assign_process` shall log the error and return `MU_FAILURE` if the `job_object_helper` is `NULL`. **]**
```
anporumb, line 78: > return `MU_FAILURE [](http://example.com/codeflow?start=94&length=19)

return non-zero.

Reason is ... MU_FAILURE gets expanded to `__LINE__`.

Now, when writing unit tests, it is impossible to know what `__LINE__` was at the time when the error happened. Also writing `result = API_CALL(...)` and then doing `ASSERT_ARE_EQUAL(int, MU_FAILURE, result)` will always fail because MU_FAILURE is expanded on the current line of the test code.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+**SRS_JOB_OBJECT_HELPER_19_020: [** If the `process_hndl` is null, then `job_object_helper_assign_process` shall set it to the `GetCurrentProcess()` output. **]**
```
anporumb, line 80: >null [](http://example.com/codeflow?start=61&length=4)

`NULL`

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +33,65 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+**SRS_JOB_OBJECT_HELPER_19_019: [** `job_object_helper_assign_process` shall log the error and return `MU_FAILURE` if `GetCurrentProcess` call fails. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_011: [** `job_object_helper_assign_process` shall call `OpenJobObject` with Permission `JOB_OBJECT_ASSIGN_PROCESS`. **]**
```
anporumb, line 84: >`OpenJobObject` [](http://example.com/codeflow?start=82&length=15)

But... if `job_object_helper` exists iand it is not `NULL` why can't it store directly the handle we have to open it again?

win32/devdoc/job_object_helper.md
anporumb, line 92: >shall return [](http://example.com/codeflow?start=71&length=12)

the usual idiom here "shall succeed and return 0".

This is to disambiguate between `"shall fail" and to establish the success return value.

win32/devdoc/job_object_helper.md
anporumb: This is reason to be "Waiting for author" for now

anporumb (changes requested):
:clock1:

anporumb:
> `job_object_helper` is a module that wraps the Windows JobObject API, primarily for the purpose of limiting the amount of memory and CPU the current process can consume.

Dan is whispering the final intention is to have "something" that limits across multiple processes.

I am not sure this PR is achieving that goal. Or the whisper is too early :)

---
Refers to: win32/devdoc/job_object_helper.md:3 in e564658. [](commit_id = e5646584ca9a7d493d4981716fae8cce0d70c462, deletion_comment = False)

win32/devdoc/job_object_helper.md
```diff
@@ -5,6 +5,9 @@
 ## Exposed API
 ```c
 MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create_with_name, const char*, job_name);
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_get, const char*, job_name);
+MOCKABLE_FUNCTION(, int, job_object_helper_assign_process, THANDLE(JOB_OBJECT_HELPER), job_object_helper, THANDLE(PROCESS_HANDLE), process_hndl);
```
mattdurak, line 10: >PROCESS_HANDLE [](http://example.com/codeflow?start=114&length=14)

How does one get a `THANDLE(PROCESS_HANDLE)`?

win32/devdoc/job_object_helper.md
```diff
@@ -5,6 +5,9 @@
 ## Exposed API
 ```c
```
mattdurak, line 6: Typically, the `Exposed API` also shows the typedefs like:

```
typedef struct JOB_OBJECT_HELPER_TAG JOB_OBJECT_HELPER;
THANDLE_TYPE_DECLARE(JOB_OBJECT_HELPER);

typedef struct PROCESS_HANDLE_TAG PROCESS_HANDLE;
THANDLE_TYPE_DECLARE(PROCESS_HANDLE);
```

(but that was partially missing before these changes)

win32/devdoc/job_object_helper.md
mattdurak, line 10: >process_hndl [](http://example.com/codeflow?start=131&length=12)

nit/poetry: consider `process_handle`

win32/devdoc/job_object_helper.md
mattdurak: Likely copied from line 19 which has double backticks as well (and should be fixed while we're here)

win32/devdoc/job_object_helper.md
shrjadhav-msft: Makes sense. Thanks for the details, it helped me. Have handled these suggestions.

shrjadhav-msft: Yeah, makes sense. handled this.

shrjadhav-msft: Agree, Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Agree, Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done. Thanks for detailed explanation.

win32/devdoc/job_object_helper.md
shrjadhav-msft: No, We can't get the unnamed jobs from windows,

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Yeah, based on whether #ifdef UNICODE it decides which one to use. 
Btw, do you meant, we could just go ahead with directly using CreateJobObjectA and OpenJobObjectA and getting rid off mbs_to_wcs call?

win32/devdoc/job_object_helper.md
shrjadhav-msft: Agree, Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done.

shrjadhav-msft:
> > `job_object_helper` is a module that wraps the Windows JobObject API, primarily for the purpose of limiting the amount of memory and CPU the current process can consume.
> 
> Dan is whispering the final intention is to have "something" that limits across multiple processes.
> 
> I am not sure this PR is achieving that goal. Or the whisper is too early :)
> 
> Refers to: win32/devdoc/job_object_helper.md:3 in [e564658](https://github.com/Azure/c-pal/commit/e5646584ca9a7d493d4981716fae8cce0d70c462). [](commit_id = [e564658](https://github.com/Azure/c-pal/commit/e5646584ca9a7d493d4981716fae8cce0d70c462), deletion_comment = False)

yeah, these are partial changes, other changes would be in the user of this `job_object_helper`..

Every Process which wants to share the resource limits(e.g. ebs replicas running in a node), will perform following actions on startup

```
job_object = job_object_helper_get("unique_job_name")
if job_object == NULL then
	job_object = job_object_helper_create_with_name("unique_job_name);
	job_object_helper_limit_cpu(job_object, cpu_rate);
	job_object_helper_limit_memory(job_object, memory_max);
job_object_helper_assign_process(job_object, GetCurrentProcess());

```

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done

win32/devdoc/job_object_helper.md
shrjadhav-msft: Ok. Added this.

win32/devdoc/job_object_helper.md
shrjadhav-msft: I suppose, this would be same using THANDLE_MALLOC(PROCESS_HANDLE) and then initializing it's process_handle member with the HANDLE of the process.
I was first confused about whether to use HANDLE or THANDLE, all over the codebase i did not find any occurrence of raw HANDLE's being used, may be THANDLE is using some sort of ref counting, so thought of using the same. Let me know, if i am missing something here.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Agree, not required to open again. removed this.

win32/devdoc/job_object_helper.md
anporumb: Probably "will allocate" from your comment sounds like a requirement. (code needs to do some malloc somewhere)

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +40,59 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+**SRS_JOB_OBJECT_HELPER_19_004: [** If there are any failures then `job_object_helper_create_with_name` shall fail and return `NULL`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_005: [** `job_object_helper_create_with_name` shall succeed and return the `JOB_OBJECT_HELPER` object. **]**
+
+
+## job_object_helper_get
+```c
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_get, const char*, job_name);
+```
+`job_object_helper_get` is used for getting a `JOB_OBJECT_HELPER` created with the name `job_name`.
```
anporumb, line 66: >for getting a `JOB_OBJECT_HELPER` created with the name `job_name`. [](http://example.com/codeflow?start=32&length=67)

Poetry: when I read this, it sounds to me like "oh, we have used `job_object_helper_create_with_name` to create 100 objects... now I need the object previously created by some string".

The API being called "_get" - that's what is makes me believe.

But reading the answers here this is more like "there's some existing Windows Job Object... and... we need it wrapped into a THANDLE".

Maybe poetry, but this does create a new THANDLE and it wraps inside an existing Windows HANDLE if it exists.

So I am asking here for a better name.

win32/devdoc/job_object_helper.md
anporumb: That's exactly what I mean.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +40,59 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+
+## job_object_helper_assign_process
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_assign_process, THANDLE(JOB_OBJECT_HELPER), job_object_helper, THANDLE(PROCESS_HANDLE), process_handle);
+```
+`job_object_helper_assign_process` assigns the process to the Job Object.
+
+**SRS_JOB_OBJECT_HELPER_19_010: [** If the `job_object_helper` is `NULL` then `job_object_helper_assign_process` shall fail and return a non-zero value. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_020: [** If the `process_handle` is null, then `job_object_helper_assign_process` shall set it to the `GetCurrentProcess()` output. **]**
```
anporumb, line 87: >null [](http://example.com/codeflow?start=63&length=4)

`NULL`

anporumb:
> #include "jobapi.h"

probably jobapi2

---
Refers to: win32/src/job_object_helper.c:8 in 57f8d72. [](commit_id = 57f8d727967caa7b74a28e6e41b4d203605cb2ce, deletion_comment = False)

win32/devdoc/job_object_helper.md
```diff
@@ -4,7 +4,17 @@
...
+
+typedef struct JOB_OBJECT_HELPER_TAG JOB_OBJECT_HELPER;
+THANDLE_TYPE_DECLARE(JOB_OBJECT_HELPER);
+
+typedef struct PROCESS_HANDLE_TAG PROCESS_HANDLE;
+THANDLE_TYPE_DECLARE(PROCESS_HANDLE);
+
 MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create_with_name, const char*, job_name);
+MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_get, const char*, job_name);
```
anporumb, line 16: Windows file HANDLES have a "OPEN_ALWAYS" mode which either creates a new file, or opens an existing one.

Borrowing from that idea: would replacing these 3 APIs with `job_object_helper_create_always" help for our purposes here?

anporumb: `CreateJobObject` seems to already have the OPEN_ALWAYS sematics:

From [https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createjobobjecta](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createjobobjecta) :

`If the object existed before the function call, the function returns a handle to the existing job object and GetLastError returns ERROR_ALREADY_EXISTS.`

anporumb:
Here we need to extra cautious about sharing across processes issues. 

Probably you're going to want to have the jobs created actually in the global namespace.

If the module can only be used as the code written above, why wouldn't the module provide just 1x API, which would be called job_object_helper_limit_cpu_and_memory_for_all_sibling_processes(cpu_limit, memory_limit) which either fails or succeeds? [poetry is bad with this name...]

That way, the user doesn't need to care about the lifetime management of the THANDLE(job_object)... because it doesn't exist. It also follows a pattern of transactional usage: either it fails, either it succeeds.

Consider.





---
In reply to: [2841327857](https://github.com/Azure/c-pal/pull/455#issuecomment-2841327857) [](http://example.com/codeflow?ancestors=2841327857)

win32/devdoc/job_object_helper.md
shrjadhav-msft: yeah makes sense. Initially, I was thinking of adding only `job_object_helper_get_or_create` instead of two separate functions `job_object_helper_create_with_name` and `job_object_helper_get`.. but later I ended up adding 2 methods, thinking it will add more modularity.. but yeah, looks like having a single method is more apt here..  
btw the existing `job_object_helper_create` also assigns the current process to the created job, and its getting used in geo_replication, so I thought of keeping it as it is. wdyt?

anporumb:
Still waiting because global namespace creation of objects needs to happen. Also would like to see more of the lifting done by the module (it is called "helper"... it should do more).

---
In reply to: [2842420276](https://github.com/Azure/c-pal/pull/455#issuecomment-2842420276) [](http://example.com/codeflow?ancestors=2842420276,2841327857)

win32/devdoc/job_object_helper.md
mattdurak: Ah, I see.

So `HANDLE` is a win32 concept. We don't use `HANDLE` directly, other than in implementations of the PAL.

All of our modules expose "opaque handles", as in we have a typedef to a struct pointer and the definition of the struct is in the .c file (so the caller of the module including the header only knows the type, but can't access its fields). This gives us encapsulation and some level of type safety for C.

In case we have objects which have shared ownership semantics, we used to have a basic "refcount" helper which would have some "inc_ref" and "dec_ref" functions. But that was a bit error prone, to make sure there were always "inc_ref" calls when storing a new reference to the object and "dec_ref" when it is out of scope/cleaned up.

So a new iteration of refcount was born called "THANDLE". The idea with `THANDLE` is that you can't write this code (without casts):

```c
THANDLE(MY_TYPE) t = NULL;
t = other; // won't compile
```

Instead, you must write:

```c
THANDLE(MY_TYPE) t = NULL;
THANDLE_ASSIGN(MY_TYPE)(&t, other);
```

Which does the reference counting inside a special "assign" function. You can think of it like overloading `operator=` in c++.

But, importantly for this case, THANDLE has an API that only exposes "assign", "move", "initialize", "initialize_move" (basically all assignment operations) as part of `THANDLE_TYPE_DECLARE`. The implementation to actually _create_ a `THANDLE` comes from `THANDLE_TYPE_DEFINE` which belongs in the .c file. Which means it's not part of the API and callers cannot call things like `THANDLE_MALLOC`.

The idea being, `THANDLE` works like our other opaque handles in that the callers can't just "create" one on their own and generally (not always) don't have access to the struct fields (and shouldn't know/care about those fields). They can just use the `THANDLE_ASSIGN` and related functions to manage the handle lifetime, and must pass it into other functions to do anything meaningful.

So whichever module defines `THANDLE(PROCESS_HANDLE)` should provide the functions to _create_ it and _use_ it. Which means if you need to get a `THANDLE(PROCESS_HANDLE)` to pass as argument to `job_object_helper_assign_process`, then there needs to be some `process_handle_create` function that takes whatever necessary arguments. I assume that means there would be maybe `process_handle_create_for_current_process` (which does `GetCurrentProcess();` as you described) and maybe some other `process_handle_create` which takes some other argument(s) to get a different process handle.

mattdurak: Ping me if you want to have a call to go over this

win32/devdoc/job_object_helper.md
mattdurak, line 87: >hall set it to the `GetCurrentProcess()` output. [](http://example.com/codeflow?start=110&length=48)

So is `process_handle` like an "in/out" argument? It may be provided and internally contain a win32 `HANDLE` which is used? Or it may be `NULL` and we use `GetCurrentProcess()` **and** we store it in `process_handle` as an out argument to the caller?

I think this adds a bit more ambiguity to the function and it feels a bit less SRP (single responsibility principle).

It would probably be better to have a separate `process_handle_create_for_current_process()` (which does `GetCurrentProcess()` and `process_handle_create(uint32_t process_id)` which does `OpenProcess(..., dwProcessId)`.

Then this `process_handle` argument is always an input only and must not be `NULL`.

win32/devdoc/job_object_helper.md
```diff
@@ -30,6 +40,59 @@ MOCKABLE_FUNCTION(, THANDLE(JOB_OBJECT_HELPER), job_object_helper_create);
...
+**SRS_JOB_OBJECT_HELPER_19_020: [** If the `process_handle` is null, then `job_object_helper_assign_process` shall set it to the `GetCurrentProcess()` output. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_013: [** `job_object_helper_assign_process` shall call `AssignProcessToJobObject` to assign the `process_handle` to the job object. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_014: [** If there are any failures then `job_object_helper_assign_process` shall fail and return a non-zero value. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_015: [** `job_object_helper_assign_process` shall succeed and return `0`. **]**
+
+
 ## job_object_helper_dispose
```
mattdurak, line 96: >job_object_helper_dispose [](http://example.com/codeflow?start=3&length=25)

will need specs for `process_handle_dispose` too

win32/devdoc/job_object_helper.md
mattdurak: Also, typically the THANDLE definition and other functions for `process_handle` would be in a separate "module" (.md/.h/.c/test files)

mattdurak:
> #include "jobapi.h"

Is this needed at all?

Looking at [AssignProcessToJobObject](https://learn.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject) it says to include Windows.h

>  jobapi2.h (include Windows.h)

---
In reply to: [2842296440](https://github.com/Azure/c-pal/pull/455#issuecomment-2842296440) [](http://example.com/codeflow?ancestors=2842296440)

---
Refers to: win32/src/job_object_helper.c:8 in 57f8d72. [](commit_id = 57f8d727967caa7b74a28e6e41b4d203605cb2ce, deletion_comment = False)

win32/devdoc/job_object_helper.md
shrjadhav-msft: ok. got your point. yes `process_handle` is an input only argument. I will update this to remove this ambiguity. thanks.

win32/devdoc/job_object_helper.md
anporumb: I think we are going to end up with 1x API in the whole module. This is what I think.

shrjadhav-msft:
> Here we need to extra cautious about sharing across processes issues.
> 
> Probably you're going to want to have the jobs created actually in the global namespace.
> 
> If the module can only be used as the code written above, why wouldn't the module provide just 1x API, which would be called job_object_helper_limit_cpu_and_memory_for_all_sibling_processes(cpu_limit, memory_limit) which either fails or succeeds? [poetry is bad with this name...]
> 
> That way, the user doesn't need to care about the lifetime management of the THANDLE(job_object)... because it doesn't exist. It also follows a pattern of transactional usage: either it fails, either it succeeds.
> 
> Consider.
> 
> In reply to: [2841327857](https://github.com/Azure/c-pal/pull/455#issuecomment-2841327857)

Agree that caller needs to be more cautious and will have to handle partial failures when say cpu limit set works and memory limit set fails etc. This could be done by retrying the failed operations again... 
I thought. adding separate API functions provides the caller more flexibility to use this helper as per need. Also given this helper is used in multiple repo's, there could be different requirements related to job object and these functions could be reused. Having only one function which does all create/assign current process/resource limit set will fix the same behavior for all callers.

mattdurak:
Yes, this is used in "multiple repos", but I think that's literally just Geo and Elastic Log. I can't speak for Geo and I don't know what future uses we have planned, but from what I've seen, I can say that Andrei's suggestion would be preferred. Our needs in Elastic Log are just "limit the CPU and/or memory for the current process".

When David pulled `job_object_helper` into Elastic Log, I had comments asking why it's even `THANDLE` when we just release it right away. We are creating this "job object helper" calling a few things on it to limit CPU and/or memory, then releasing it. Why have a handle that doesn't really need a lifetime beyond 2 function calls? And if we really want to have separate create/set limit calls, then I still say "why THANDLE"? This helper doesn't need to live for very long.

So maybe we can check with Bert on how they use this in Geo, but for Elastic Log, if we just had 1 API to set the CPU and memory limit for the current process, that would be ideal (with something like 0 meaning no limit).

For reference, we actually did just wrap it that way for Elastic Log, see: [https://msazure.visualstudio.com/One/_git/Azure-Messaging-ElasticLog?path=/el_sf_service/src/elastic_log_service_limits.c&_a=contents&version=GBmaster](https://msazure.visualstudio.com/One/_git/Azure-Messaging-ElasticLog?path=/el_sf_service/src/elastic_log_service_limits.c&_a=contents&version=GBmaster)

In fact, it looks like Geo does the same (without tests!) [https://msazure.visualstudio.com/One/_git/Azure-Messaging-GeoReplication?path=/geo_interop/src/geo_replicator_factory.c&_a=contents&version=GBmaster](https://msazure.visualstudio.com/One/_git/Azure-Messaging-GeoReplication?path=/geo_interop/src/geo_replicator_factory.c&_a=contents&version=GBmaster)

---
In reply to: [2843012199](https://github.com/Azure/c-pal/pull/455#issuecomment-2843012199) [](http://example.com/codeflow?ancestors=2843012199)

shrjadhav-msft:
> Yes, this is used in "multiple repos", but I think that's literally just Geo and Elastic Log. I can't speak for Geo and I don't know what future uses we have planned, but from what I've seen, I can say that Andrei's suggestion would be preferred. Our needs in Elastic Log are just "limit the CPU and/or memory for the current process".
> 
> When David pulled `job_object_helper` into Elastic Log, I had comments asking why it's even `THANDLE` when we just release it right away. We are creating this "job object helper" calling a few things on it to limit CPU and/or memory, then releasing it. Why have a handle that doesn't really need a lifetime beyond 2 function calls? And if we really want to have separate create/set limit calls, then I still say "why THANDLE"? This helper doesn't need to live for very long.
> 
> So maybe we can check with Bert on how they use this in Geo, but for Elastic Log, if we just had 1 API to set the CPU and memory limit for the current process, that would be ideal (with something like 0 meaning no limit).
> 
> For reference, we actually did just wrap it that way for Elastic Log, see: https://msazure.visualstudio.com/One/_git/Azure-Messaging-ElasticLog?path=/el_sf_service/src/elastic_log_service_limits.c&_a=contents&version=GBmaster
> 
> In fact, it looks like Geo does the same (without tests!) https://msazure.visualstudio.com/One/_git/Azure-Messaging-GeoReplication?path=/geo_interop/src/geo_replicator_factory.c&_a=contents&version=GBmaster
> 
> In reply to: [2843012199](https://github.com/Azure/c-pal/pull/455#issuecomment-2843012199)

Cool, Thanks for clarifications.

@anporumb @mattdurak Let me know if following looks fine then, 
1. Remove all existing THANDLE stuff from job_object_helper
2. have a single API `job_object_helper_set_cpu_memory_limits(global_job_name, cpu_limit, memory_limit)` ; this would then internally check if global job object exists, if not create the same, assign the current process to it and then set the limits.
3. update the existing code in geo-replication/elastic log to use this updated function.

anporumb:
I am fine with having just 1 API.

---
In reply to: [2843960318](https://github.com/Azure/c-pal/pull/455#issuecomment-2843960318) [](http://example.com/codeflow?ancestors=2843960318)

shrjadhav-msft:
> I am fine with having just 1 API.
> 
> In reply to: [2843960318](https://github.com/Azure/c-pal/pull/455#issuecomment-2843960318)

@anporumb @mattdurak 

have added a single API function which Creates the job object, if one doesn't exist and apply the limits and then assign the current process to the job object. 

For now, I haven't removed existing exposed API functions from this module. As those are being used in replicator/elastic log.. I am thinking, We can change the code in replicator/elastic log to use this new function first and then remove all THANDLE and existing functions from the API.  Let me know if you are ok with the approach? I will create the tasks for these and work on them.  thanks.

win32/devdoc/job_object_helper.md
shrjadhav-msft: removed this function. so resolving this.

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,30 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_set_job_limits_to_current_process, const char*, job_name, uint32_t, percent_cpu, uint32_t, percent_physical_memory);
+```
+`job_object_helper_set_job_limits_to_current_process` Creates the Job Object with limits if the job object doesn't exists and set limits on the current process by assigning it to the Job object.
+
+**SRS_JOB_OBJECT_HELPER_19_001: [** if `job_name` is not `NULL` or empty, `job_object_helper_set_job_limits_to_current_process` shall call `OpenJobObjectA` passing `job_name` for `lpName`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_002: [** Otherwise `job_object_helper_set_job_limits_to_current_process` shall call `CreateJobObjectA` passing `job_name` for `lpName` and `NULL` for `lpJobAttributes`. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_003: [** `job_object_helper_set_job_limits_to_current_process` shall call `SetInformationJobObject` passing `JobObjectCpuRateControlInformation` and a  `JOBOBJECT_CPU_RATE_CONTROL_INFORMATION` object with `JOB_OBJECT_CPU_RATE_CONTROL_ENABLE` and `JOB_OBJECT_CPU_RATE_CONTROL_HARD_CAP` set, and `CpuRate` set to `percent_cpu` times `100`. **]**
```
mattdurak, line 97: >percent_cpu [](http://example.com/codeflow?start=339&length=11)

This requirement should be prefixed with "If `percent_cpu` is not 0".

Same for the physical memory parts.

In ELS we want to allow restricting just CPU or just memory in some cases so we don't want to set the other limit at all if only one applies.

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,30 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+
+**SRS_JOB_OBJECT_HELPER_19_004: [** `job_object_helper_set_job_limits_to_current_process` shall call `GlobalMemoryStatusEx` to get the total amount of physical memory in kb. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_005: [** `job_object_helper_set_job_limits_to_current_process` shall call `SetInformationJobObject`, passing `JobObjectExtendedLimitInformation` and a `JOBOBJECT_EXTENDED_LIMIT_INFORMATION` object with `JOB_OBJECT_LIMIT_JOB_MEMORY` set and `JobMemoryLimit` set to the `percent_physical_memory` percent of the physical memory in bytes. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_006: [** `job_object_helper_set_job_limits_to_current_process` shall call `GetCurrentProcess` to get the current process handle. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_007: [** `job_object_helper_set_job_limits_to_current_process` shall call `AssignProcessToJobObject` to assign the current process to the new job object. **]**
+
+**SRS_JOB_OBJECT_HELPER_19_008: [** `job_object_helper_set_job_limits_to_current_process` shall call `CloseHandle` to close the handle of the current process. **]**
```
mattdurak, line 107: >CloseHandle [](http://example.com/codeflow?start=102&length=11)

We also need a spec for closing the job object handle

mattdurak:
Sounds like the right approach. Add the new function, update the other repos to use it, cleanup the old ones. Less disruption when propagating these changes.

---
In reply to: [2846975494](https://github.com/Azure/c-pal/pull/455#issuecomment-2846975494) [](http://example.com/codeflow?ancestors=2846975494)

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Agree, Done.

mattdurak (dismissed):
:shipit:

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,36 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_set_job_limits_to_current_process, const char*, job_name, uint32_t, percent_cpu, uint32_t, percent_physical_memory);
+```
+`job_object_helper_set_job_limits_to_current_process` Creates the Job Object with limits if the job object doesn't exists and set limits on the current process by assigning it to the Job object.
+
+**S_RS_JOB_OBJECT_HELPER_19_001: [** if `job_name` is not `NULL` or empty, `job_object_helper_set_job_limits_to_current_process` shall call `OpenJobObjectA` passing `job_name` for `lpName`. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_002: [** Otherwise `job_object_helper_set_job_limits_to_current_process` shall call `CreateJobObjectA` passing `job_name` for `lpName` and `NULL` for `lpJobAttributes`. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_013: [** if `percent_cpu` is greater than `100` then `job_object_helper_set_job_limits_to_current_process` should fail. **]**
```
anporumb, line 97: >should [](http://example.com/codeflow?start=135&length=6)

`should` is weak, it implies "might or might not".

We shall only use `shall`!

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,36 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+
+## job_object_helper_set_job_limits_to_current_process
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_set_job_limits_to_current_process, const char*, job_name, uint32_t, percent_cpu, uint32_t, percent_physical_memory);
+```
+`job_object_helper_set_job_limits_to_current_process` Creates the Job Object with limits if the job object doesn't exists and set limits on the current process by assigning it to the Job object.
+
+**S_RS_JOB_OBJECT_HELPER_19_001: [** if `job_name` is not `NULL` or empty, `job_object_helper_set_job_limits_to_current_process` shall call `OpenJobObjectA` passing `job_name` for `lpName`. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_002: [** Otherwise `job_object_helper_set_job_limits_to_current_process` shall call `CreateJobObjectA` passing `job_name` for `lpName` and `NULL` for `lpJobAttributes`. **]**
```
anporumb, line 95: >Otherwise [](http://example.com/codeflow?start=37&length=9)

Unclear to what this 'otherwise' binds to. It seems to bind to the case when the job_name is NULL for example.

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,36 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
 **SRS_JOB_OBJECT_HELPER_18_050: [** `job_object_helper_limit_cpu` shall succeed and return `0`. **]**
+
+
+## job_object_helper_set_job_limits_to_current_process
+```c
+MOCKABLE_FUNCTION(, int, job_object_helper_set_job_limits_to_current_process, const char*, job_name, uint32_t, percent_cpu, uint32_t, percent_physical_memory);
+```
+`job_object_helper_set_job_limits_to_current_process` Creates the Job Object with limits if the job object doesn't exists and set limits on the current process by assigning it to the Job object.
+
+**S_RS_JOB_OBJECT_HELPER_19_001: [** if `job_name` is not `NULL` or empty, `job_object_helper_set_job_limits_to_current_process` shall call `OpenJobObjectA` passing `job_name` for `lpName`. **]**
```
anporumb, line 93: >** [](http://example.com/codeflow?start=0&length=2)

all parameters with all values are valid for this function?

win32/devdoc/job_object_helper.md
anporumb, line 97: >i [](http://example.com/codeflow?start=37&length=1)

Capital letter for beginning of phrase. Here and everywhere

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,36 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+
+**S_RS_JOB_OBJECT_HELPER_19_002: [** Otherwise `job_object_helper_set_job_limits_to_current_process` shall call `CreateJobObjectA` passing `job_name` for `lpName` and `NULL` for `lpJobAttributes`. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_013: [** if `percent_cpu` is greater than `100` then `job_object_helper_set_job_limits_to_current_process` should fail. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_003: [** If `percent_cpu` is not `0` then `job_object_helper_set_job_limits_to_current_process` shall call `SetInformationJobObject` passing `JobObjectCpuRateControlInformation` and a  `JOBOBJECT_CPU_RATE_CONTROL_INFORMATION` object with `JOB_OBJECT_CPU_RATE_CONTROL_ENABLE` and `JOB_OBJECT_CPU_RATE_CONTROL_HARD_CAP` set, and `CpuRate` set to `percent_cpu` times `100`. **]**
```
anporumb, line 99: >   [](http://example.com/codeflow?start=211&length=2)

2 spaces.

win32/devdoc/job_object_helper.md
```diff
@@ -77,3 +82,36 @@ MOCKABLE_FUNCTION(, int, job_object_helper_limit_cpu, THANDLE(JOB_OBJECT_HELPER)
...
+
+**S_RS_JOB_OBJECT_HELPER_19_006: [** `job_object_helper_set_job_limits_to_current_process` shall call `GetCurrentProcess` to get the current process handle. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_007: [** `job_object_helper_set_job_limits_to_current_process` shall call `AssignProcessToJobObject` to assign the current process to the new job object. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_008: [** `job_object_helper_set_job_limits_to_current_process` shall call `CloseHandle` to close the handle of the current process. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_011: [** `job_object_helper_set_job_limits_to_current_process` shall call `CloseHandle` to close the handle of the Job object. **]**
+
+**S_RS_JOB_OBJECT_HELPER_19_009: [** If there are any failures, `job_object_helper_set_job_limits_to_current_process` shall fail and return a non-zero value. **]**
```
anporumb, line 115: >and return a non-zero value. [](http://example.com/codeflow?start=129&length=28)

"and return a non-zero value."

win32/devdoc/job_object_helper.md
anporumb, line 93: >OpenJobObjectA [](http://example.com/codeflow?start=141&length=14)

Why does it need to do OpenJob? Why is CreateJobObject not enough?

anporumb: Still waiting.

mattdurak (dismissed):
:shipit:

win32/devdoc/job_object_helper.md
shrjadhav-msft: If job-name is null then it will create a non-named job object. for percent_cpu and percent_memory_limit have handled cases .

win32/devdoc/job_object_helper.md
shrjadhav-msft: Updated.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Agree. Done.

win32/devdoc/job_object_helper.md
shrjadhav-msft: Done.

win32/reals/real_job_object_helper_renames.h
```diff
@@ -1,8 +1,9 @@
...
+#define JOB_OBJECT_HELPER                                   real_JOB_OBJECT_HELPER
 
-#define job_object_helper_create        real_job_object_helper_create 
-#define job_object_helper_limit_memory  real_job_object_helper_limit_memory
-#define job_object_helper_limit_cpu     real_job_object_helper_limit_cpu
\ No newline at end of file
+#define job_object_helper_create                            real_job_object_helper_create
+#define job_object_helper_limit_memory                      real_job_object_helper_limit_memory
+#define job_object_helper_limit_cpu                         real_job_object_helper_limit_cpu
+#define job_object_helper_set_job_limits_to_current_process real_job_object_helper_set_job_limits_to_current_process
```
anporumb, line 9: >s [](http://example.com/codeflow?start=115&length=1)

file needs to end with an empty line.

here and everywhere.

shrjadhav-msft: Done

shrjadhav-msft:
@microsoft-github-policy-service agree [company="{Microsoft}"]


---
In reply to: [2838553342](https://github.com/Azure/c-pal/pull/455#issuecomment-2838553342) [](http://example.com/codeflow?ancestors=2838553342)

shrjadhav-msft:
@microsoft-github-policy-service agree company="Microsoft"

---
In reply to: [2863888464](https://github.com/Azure/c-pal/pull/455#issuecomment-2863888464) [](http://example.com/codeflow?ancestors=2863888464)

shrjadhav-msft:
Done

---
In reply to: [2804935412](https://github.com/Azure/c-pal/pull/455#issuecomment-2804935412) [](http://example.com/codeflow?ancestors=2804935412)

shrjadhav-msft:
Thanks

---
In reply to: [2815356104](https://github.com/Azure/c-pal/pull/455#issuecomment-2815356104) [](http://example.com/codeflow?ancestors=2815356104)

shrjadhav-msft:
Thanks

---
In reply to: [2822584843](https://github.com/Azure/c-pal/pull/455#issuecomment-2822584843) [](http://example.com/codeflow?ancestors=2822584843)

shrjadhav-msft:
\azp run

mattdurak (approved):
:shipit:

anporumb (approved):
:shipit:

PR #456, opened 2025-04-29
Remove lazy_init from gballoc_hl_passthrough

mattdurak:
It was observed in some CPU profiling that every call to malloc does an `interlocked_compare_exchange`.
This is because gballoc has a lazy initialization.

Most of the gballoc implementations do **not** have an init/deinit implementation and they just no-op. However, `gballoc_hl_passthrough` was doing a lazy init _just in case_ it needed to lazy initialize the `gballoc_ll`. This is not necessary as the only `gballoc_ll` which does initialization (`gballoc_ll_win32heap`) also does `lazy_init` on its calls. Therefore, we can leave it to the LL to `lazy_init` when needed and avoid doing so for most of the LL that don't do any initialization (passthrough, jemalloc, mimalloc).

This is a small perf improvement, but low-hanging fruit and can add up when there are many allocations happening.

### DIFF
--- a/linux/src/gballoc_hl_passthrough.c
+++ b/linux/src/gballoc_hl_passthrough.c
@@ -16,9 +16,7 @@ int gballoc_hl_init(void* gballoc_hl_init_params, void* gballoc_ll_init_params)
     (void)gballoc_hl_init_params; /*are ignored, this is "passthrough*/
 
     /*Codes_SRS_GBALLOC_HL_PASSTHROUGH_42_001: [ gballoc_hl_init shall call gballoc_ll_init as function to execute and gballoc_ll_init_params as parameter and return the result. ]*/
-    int result = gballoc_ll_init(gballoc_ll_init_params);
-
-    return result;
+    return gballoc_ll_init(gballoc_ll_init_params);
 }
 
 void gballoc_hl_deinit(void)
@@ -71,12 +69,8 @@ void gballoc_hl_free(void* ptr)
 
 size_t gballoc_hl_size(void* ptr)
 {
-    size_t result;
-
     /* Codes_SRS_GBALLOC_HL_PASSTHROUGH_01_003: [ Otherwise, gballoc_hl_size shall call gballoc_ll_size with ptr as argument and return the result of gballoc_ll_size. ]*/
-    result = gballoc_ll_size(ptr);
-
-    return result;
+    return gballoc_ll_size(ptr);
 }
 
 void* gballoc_hl_calloc(size_t nmemb, size_t size)
--- a/linux/src/gballoc_hl_passthrough.c
+++ b/linux/src/gballoc_hl_passthrough.c
@@ -11,27 +11,12 @@
 
 #include "c_pal/gballoc_hl.h"
 
-static int wasInitialized = 0;
-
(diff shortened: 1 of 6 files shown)


linux/src/gballoc_hl_passthrough.c
```diff
@@ -11,39 +11,20 @@
...
-    else
-    {
-        /*Codes_SRS_GBALLOC_HL_PASSTHROUGH_02_002: [ gballoc_hl_init shall succeed and return 0. ]*/
-        result = 0;
-        wasInitialized = 1;
-    }
+    /*Codes_SRS_GBALLOC_HL_PASSTHROUGH_42_001: [ gballoc_hl_init shall call gballoc_ll_init as function to execute and gballoc_ll_init_params as parameter and return the result. ]*/
+    int result = gballoc_ll_init(gballoc_ll_init_params);
 
     return result;
```
dcristoloveanu, line 21: Should this just be a one liner?

linux/src/gballoc_hl_passthrough.c
```diff
@@ -92,16 +73,9 @@ size_t gballoc_hl_size(void* ptr)
...
-        /* Codes_SRS_GBALLOC_HL_PASSTHROUGH_01_002: [ If the module was not initialized, gballoc_hl_size shall return 0. ]*/
-        result = 0;
-    }
-    else
-    {
-        /* Codes_SRS_GBALLOC_HL_PASSTHROUGH_01_003: [ Otherwise, gballoc_hl_size shall call gballoc_ll_size with ptr as argument and return the result of gballoc_ll_size. ]*/
-        result = gballoc_ll_size(ptr);
-    }
+    /* Codes_SRS_GBALLOC_HL_PASSTHROUGH_01_003: [ Otherwise, gballoc_hl_size shall call gballoc_ll_size with ptr as argument and return the result of gballoc_ll_size. ]*/
+    result = gballoc_ll_size(ptr);
```
dcristoloveanu, line 77: Also smells like a one liner

dcristoloveanu:
I think we should migrate towards not doing the lazy init for Xalloc modules at all. We should probably make sure we call init/deinit once in the main of the exes we have - maybe by failing somehow very horribly in Debug for example. Because technically there is no need for us to do an extra interlocked for every malloc/free call, since we are in full control of our system.

dcristoloveanu (approved):
:shipit:

mattdurak:
Agreed!
However this feels very low priority now because there are only 2 modules that do any initialization at all: gballoc_hl_metrics, and gballoc_ll_win32heap. Since we are really only using gballoc_hl_passthrough and gballoc_ll_passthrough/gballoc_ll_jemalloc.

---
In reply to: [2840300451](https://github.com/Azure/c-pal/pull/456#issuecomment-2840300451) [](http://example.com/codeflow?ancestors=2840300451)

linux/src/gballoc_hl_passthrough.c
mattdurak: changed!

linux/src/gballoc_hl_passthrough.c
mattdurak: changed!

anporumb (approved):
:shipit:

PR #460, opened 2025-05-14
Task-325555241 added integration tests

shrjadhav-msft:
This PR adds integration tests for job_object_helper.
For launching multiple processes and calling job_object_helper functions from it, this PR adds a new binary job_object_helper_tester under tools directory. Integration tests uses CreateProcess to launch multiple processes calling job_object_helper function with the same job name.


shrjadhav-msft:
Added integration tests in https://github.com/Azure/c-pal/pull/458

PR #463, opened 2025-06-10
Skipping the iterations for higher thread counts to avoid timeout

flash-sinx:
Skipping the iterations for higher thread counts where non barrier to barrier calls ratio is higher

### DIFF
--- a/common/tests/sm_int/sm_int.c
+++ b/common/tests/sm_int/sm_int.c
@@ -1097,6 +1097,9 @@ TEST_FUNCTION(sm_does_not_block)
     {
         for(uint32_t n_barrier_threads=0; n_barrier_threads<=nthreads; n_barrier_threads+=4)
         {
+            // This check is added to avoid excessive barrier threads when the total thread count is high.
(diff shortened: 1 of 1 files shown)


common/tests/sm_int/sm_int.c
```diff
@@ -1097,6 +1097,10 @@ TEST_FUNCTION(sm_does_not_block)
     {
         for(uint32_t n_barrier_threads=0; n_barrier_threads<=nthreads; n_barrier_threads+=4)
         {
+            if (nthreads >= 64 && n_barrier_threads > (nthreads * 3 / 5)) // 3/5 = 60%
+            {
+                continue; // Skip this iteration
```
anporumb, line 1102: >// Skip this iteration [](http://example.com/codeflow?start=26&length=22)

comment should explain why we are skipping this iteration.

Otherwise in 1 year nobody's going to remember why.

anporumb (dismissed):
:shipit:

common/tests/sm_int/sm_int.c
flash-sinx: Makes sense, added the comment.

ACKPaul (approved):

PR #473, opened 2025-07-21
Fix linux threadpool reals

BertKleewein:


linux/linux_reals/real_threadpool.c
```diff
@@ -11,4 +11,12 @@
 
 #include "real_srw_lock_ll_renames.h"
 
+#include "real_threadapi_renames.h"
+
+#include "real_interlocked_hl_renames.h"
```
darobs, line 16: You don't need real_intrerlocked_renames.h?

anporumb (dismissed):
:shipit:

linux/linux_reals/real_threadpool.c
BertKleewein: Yeah, good question.  I think that nobody is using the linux threadpool_reals, otherwise this would have been caught much earlier.  Honestly, I'm not sure why I needed to make this change at all.  There's too much linker magic going on in this code.

BertKleewein: added it

PR #561, opened 2026-03-28
Fix socket handle leak: always call closesocket/close after shutdown in disconnect

dcristoloveanu:
## Summary

Fixes socket handle leaks in \socket_transport_disconnect()\ detected by VLD in integration tests.

**Root cause:** When \shutdown()\ failed on non-binding sockets, \closesocket()\/\close()\ was skipped due to an \if/else\ structure, leaking the socket handle and associated Winsock internal allocations (\WahInsertHandleContext\, 256 bytes each).

**Fix:** Ensure \closesocket()\/\close()\ is always called regardless of \shutdown()\ result. Per Windows/POSIX docs, \shutdown()\ is for graceful connection termination while \closesocket()\/\close()\ releases the socket descriptor and must always be called.

## Changes
- **win32/src/socket_transport_win32.c**: Restructured disconnect to always call \closesocket()\
- **linux/src/socket_transport_linux.c**: Restructured disconnect to always call \close()\
- **win32/devdoc/socket_transport_win32_requirements.md**: Updated SRS_SOCKET_TRANSPORT_WIN32_09_083
- **linux/devdoc/socket_transport_linux_requirements.md**: Updated SRS_SOCKET_TRANSPORT_LINUX_11_026
- **win32/tests/socket_transport_win32_ut**: Updated unit test expectations
- **linux/tests/socket_transport_linux_ut**: Updated unit test expectations

## Testing
- Γ£à Unit tests pass (socket_transport_win32_ut, socket_transport_linux_ut)
- Γ£à Integration tests pass (socket_transport_int)

Fixes AB#37142411

---
Session ID: `62cfb822-ac74-4d9a-96e3-7bded6ab46d2`

### DIFF
--- a/linux/src/socket_transport_linux.c
+++ b/linux/src/socket_transport_linux.c
@@ -340,16 +340,13 @@ void socket_transport_disconnect(SOCKET_TRANSPORT_HANDLE socket_transport)
             // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_025: [ socket_transport_disconnect shall call shutdown to stop both the transmit and reception of the connected socket. ]
             if (shutdown(socket_transport->socket, 2) != 0)
             {
-                // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_026: [ If shutdown does not return 0, the socket is not valid therefore socket_transport_disconnect shall not call 'close' ]
                 LogErrorNo("shutdown failed on socket: %" PRI_SOCKET "", socket_transport->socket);
             }
-            else
+
+            // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_023: [ socket_transport_disconnect shall call close to disconnect the connected socket. ]
+            if (close(socket_transport->socket) != 0)
             {
-                // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_023: [ socket_transport_disconnect shall call close to disconnect the connected socket. ]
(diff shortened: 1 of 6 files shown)


linux/src/socket_transport_linux.c
```diff
@@ -340,16 +340,14 @@ void socket_transport_disconnect(SOCKET_TRANSPORT_HANDLE socket_transport)
             // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_025: [ socket_transport_disconnect shall call shutdown to stop both the transmit and reception of the connected socket. ]
             if (shutdown(socket_transport->socket, 2) != 0)
             {
-                // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_026: [ If shutdown does not return 0, the socket is not valid therefore socket_transport_disconnect shall not call 'close' ]
+                // Codes_SRS_SOCKET_TRANSPORT_LINUX_11_026: [ If shutdown does not return 0, socket_transport_disconnect shall log the error. ]
```
dcristoloveanu: >SRS_SOCKET_TRANSPORT_LINUX_11_026 [](http://example.com/codeflow?start=25&length=33)

There is no point in having this spec, we do not spec just logging errors ... Fix this for windows and linux #Resolved

parth21999 (approved):

PR #587, opened 2026-05-07
Fix InterlockedHL_WaitForXXX timeout accounting

flash-sinx:
## Summary
All 4 `InterlockedHL_WaitForXXX` functions passed the full original `timeout_ms` to `wait_on_address` on every loop iteration. After a spurious wakeup (value changed but not to the target), the loop restarted with a fresh timeout, making the total wait potentially unbounded.

## Fix
Track elapsed time using `timer_global_get_elapsed_ms()` and pass only the remaining time to each `wait_on_address` call. If elapsed time exceeds `timeout_ms` before calling `wait_on_address`, return `INTERLOCKED_HL_TIMEOUT` immediately.

## Files changed (5)
- `common/devdoc/interlocked_hl_requirements.md` - 12 new SRS requirements (27_001-27_012)
- `common/src/interlocked_hl.c` - Timer-based timeout tracking in all 4 WaitForXXX functions
- `common/reals/real_interlocked_hl.c` - Added real_timer_renames.h include
- `common/tests/interlocked_hl_ut/interlocked_hl_ut_pch.h` - Added timer.h mock
- `common/tests/interlocked_hl_ut/interlocked_hl_ut.c` - Updated existing tests + 4 new timeout accounting tests

## Validation
- Build: clean, zero warnings
- Tests: 58/58 passed
- SRS consistency, uniqueness, AAA comments, spec tags: all pass

AB#37772988

### DIFF
--- a/common/devdoc/interlocked_hl_requirements.md
+++ b/common/devdoc/interlocked_hl_requirements.md
@@ -68,11 +68,21 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
 
 **SRS_INTERLOCKED_HL_01_003: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_OK`. **]**
 
-**SRS_INTERLOCKED_HL_01_004: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall wait until the value at `address_to_check` changes using `wait_on_address`. **]**
+**SRS_INTERLOCKED_HL_27_001: [** `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` to determine the start time of the wait. **]**
+
+**SRS_INTERLOCKED_HL_27_017: [** If `timeout_ms` is `UINT32_MAX`, `InterlockedHL_WaitForValue` shall pass `UINT32_MAX` to `wait_on_address`. **]**
+
+**SRS_INTERLOCKED_HL_27_002: [** `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start. **]**
+
+**SRS_INTERLOCKED_HL_27_021: [** If the elapsed time is negative, `InterlockedHL_WaitForValue` shall treat it as zero. **]**
+
+**SRS_INTERLOCKED_HL_27_013: [** If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_TIMEOUT`. **]**
+
+**SRS_INTERLOCKED_HL_01_004: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall wait until the value at `address_to_check` changes using `wait_on_address` with a timeout equal to `timeout_ms` minus the elapsed time. **]**
 
 **SRS_INTERLOCKED_HL_01_007: [** When `wait_on_address` succeeds, `InterlockedHL_WaitForValue` shall again compare the value at `address_to_check` with `value_to_wait`. **]**
 
-**SRS_INTERLOCKED_HL_11_001: [** If `wait_on_address` hits the timeout specified in `timeout_ms`, `InterlockedHL_WaitForValue` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
+**SRS_INTERLOCKED_HL_11_001: [** If `wait_on_address` hits the timeout, `InterlockedHL_WaitForValue` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
 
 **SRS_INTERLOCKED_HL_01_006: [** If `wait_on_address` fails, `InterlockedHL_WaitForValue` shall fail and return `INTERLOCKED_HL_ERROR`. **]**
 
@@ -88,11 +98,21 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
 
 **SRS_INTERLOCKED_HL_05_002: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForValue64` shall return `INTERLOCKED_HL_OK`. **]**
 
-**SRS_INTERLOCKED_HL_05_003: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue64` shall wait until the value at `address_to_check` changes using `wait_on_address_64`. **]**
+**SRS_INTERLOCKED_HL_27_004: [** `InterlockedHL_WaitForValue64` shall call `timer_global_get_elapsed_ms` to determine the start time of the wait. **]**
+
+**SRS_INTERLOCKED_HL_27_018: [** If `timeout_ms` is `UINT32_MAX`, `InterlockedHL_WaitForValue64` shall pass `UINT32_MAX` to `wait_on_address_64`. **]**
+
+**SRS_INTERLOCKED_HL_27_005: [** `InterlockedHL_WaitForValue64` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start. **]**
+
+**SRS_INTERLOCKED_HL_27_022: [** If the elapsed time is negative, `InterlockedHL_WaitForValue64` shall treat it as zero. **]**
+
+**SRS_INTERLOCKED_HL_27_014: [** If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForValue64` shall return `INTERLOCKED_HL_TIMEOUT`. **]**
+
+**SRS_INTERLOCKED_HL_05_003: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue64` shall wait until the value at `address_to_check` changes using `wait_on_address_64` with a timeout equal to `timeout_ms` minus the elapsed time. **]**
 
 **SRS_INTERLOCKED_HL_05_004: [** When `wait_on_address_64` succeeds, `InterlockedHL_WaitForValue64` shall again compare the value at `address_to_check` with `value_to_wait`. **]**
 
-**SRS_INTERLOCKED_HL_05_005: [** If `wait_on_address_64` hits the timeout specified in timeout_ms, `InterlockedHL_WaitForValue64` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
+**SRS_INTERLOCKED_HL_05_005: [** If `wait_on_address_64` hits the timeout, `InterlockedHL_WaitForValue64` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
 
 **SRS_INTERLOCKED_HL_05_006: [** If `wait_on_address_64` fails, `InterlockedHL_WaitForValue64` shall fail and return `INTERLOCKED_HL_ERROR`. **]**
 
@@ -109,11 +129,21 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForNot
 
 **SRS_INTERLOCKED_HL_42_002: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForNotValue` shall return `INTERLOCKED_HL_OK`. **]**
 
-**SRS_INTERLOCKED_HL_42_003: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForNotValue` shall wait until the value at `address_to_check` changes by using `wait_on_address`. **]**
+**SRS_INTERLOCKED_HL_27_007: [** `InterlockedHL_WaitForNotValue` shall call `timer_global_get_elapsed_ms` to determine the start time of the wait. **]**
+
+**SRS_INTERLOCKED_HL_27_019: [** If `timeout_ms` is `UINT32_MAX`, `InterlockedHL_WaitForNotValue` shall pass `UINT32_MAX` to `wait_on_address`. **]**
+
+**SRS_INTERLOCKED_HL_27_008: [** `InterlockedHL_WaitForNotValue` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start. **]**
+
+**SRS_INTERLOCKED_HL_27_023: [** If the elapsed time is negative, `InterlockedHL_WaitForNotValue` shall treat it as zero. **]**
+
+**SRS_INTERLOCKED_HL_27_015: [** If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForNotValue` shall return `INTERLOCKED_HL_TIMEOUT`. **]**
+
+**SRS_INTERLOCKED_HL_42_003: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForNotValue` shall wait until the value at `address_to_check` changes by using `wait_on_address` with a timeout equal to `timeout_ms` minus the elapsed time. **]**
 
 **SRS_INTERLOCKED_HL_42_005: [** When `wait_on_address` succeeds, `InterlockedHL_WaitForNotValue` shall again compare the value at `address_to_check` with `value_to_wait`. **]**
 
-**SRS_INTERLOCKED_HL_11_002: [** If `wait_on_address` hits the timeout specified in timeout_ms, `InterlockedHL_WaitForNotValue` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
+**SRS_INTERLOCKED_HL_11_002: [** If `wait_on_address` hits the timeout, `InterlockedHL_WaitForNotValue` shall fail and return `INTERLOCKED_HL_TIMEOUT`. **]**
 
 **SRS_INTERLOCKED_HL_42_007: [** If `wait_on_address` fails, `InterlockedHL_WaitForNotValue` shall fail and return `INTERLOCKED_HL_ERROR`. **]**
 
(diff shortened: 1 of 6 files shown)


common/devdoc/interlocked_hl_requirements.md
```diff
@@ -68,8 +68,14 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
 
 **SRS_INTERLOCKED_HL_01_003: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_OK`. **]**
 
+**SRS_INTERLOCKED_HL_27_001: [** `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` to determine the start time of the wait. **]**
+
+**SRS_INTERLOCKED_HL_27_002: [** Before each call to `wait_on_address`, `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start. If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_TIMEOUT`. **]**
```
RohitG28: we can break this down into two specs.
- `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start
- If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_TIMEOUT`.

common/devdoc/interlocked_hl_requirements.md
```diff
@@ -68,8 +68,14 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
...
+
 **SRS_INTERLOCKED_HL_01_004: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall wait until the value at `address_to_check` changes using `wait_on_address`. **]**
 
+**SRS_INTERLOCKED_HL_27_003: [** `InterlockedHL_WaitForValue` shall call `wait_on_address` with a timeout equal to `timeout_ms` minus the elapsed time. **]**
+
```
RohitG28: we can merge these two specs into one

RohitG28 (dismissed):

common/devdoc/interlocked_hl_requirements.md
flash-sinx: Done. Split 27_002 into two specs: 27_002 (compute elapsed) and 27_013 (timeout check). Applied the same pattern to 27_005/014, 27_008/015, 27_011/016 for the other three functions.

common/devdoc/interlocked_hl_requirements.md
flash-sinx: Done. Merged 01_004 + 27_003 into a single 01_004 with updated wording: '...using wait_on_address with a timeout equal to timeout_ms minus the elapsed time.' Applied same merge for 05_003+27_006, 42_003+27_009, 05_009+27_012.

common/src/interlocked_hl.c
```diff
@@ -104,8 +108,18 @@ INTERLOCKED_HL_RESULT InterlockedHL_WaitForValue(int32_t volatile_atomic* addres
...
+            /* Codes_SRS_INTERLOCKED_HL_27_013: [ If the elapsed time is greater than or equal to timeout_ms, InterlockedHL_WaitForValue shall return INTERLOCKED_HL_TIMEOUT. ] */
+            if (elapsed_ms >= (double)timeout_ms)
+            {
+                result = INTERLOCKED_HL_TIMEOUT;
+                break;
+            }
+
+            /* Codes_SRS_INTERLOCKED_HL_01_004: [ If the value at address_to_check is not equal to value_to_wait, InterlockedHL_WaitForValue shall wait until the value at address_to_check changes using wait_on_address with a timeout equal to timeout_ms minus the elapsed time. ]*/
+            uint32_t remaining_ms = timeout_ms - (uint32_t)elapsed_ms;
+            WAIT_ON_ADDRESS_RESULT wait_result = wait_on_address(address_to_check, current_value, remaining_ms);
```
dcristoloveanu, line 139: [MrBot] **[major | correctness]** `UINT32_MAX` no longer means "wait indefinitely"

`SRS_SYNC_43_008` defines that `wait_on_address` waits indefinitely when `timeout_ms == UINT32_MAX`. After the first spurious wakeup, this code passes `remaining_ms = UINT32_MAX - elapsed_ms`, which is finite (~49 days). Callers that pass `UINT32_MAX` as "infinite" silently lose that semantic — and several production callers do exactly that:
- `common/src/sm.c:247` (`sm_close_begin_internal`)
- `common/src/sm.c:521` (`sm_barrier_begin`)
- `linux/src/completion_port_linux.c:467`
- `linux/src/threadpool_linux.c:444`

**Fix:** special-case `timeout_ms == UINT32_MAX` and pass it straight through, skipping elapsed accounting. The same issue applies in all 4 `InterlockedHL_WaitFor*` functions.

common/src/interlocked_hl.c
```diff
@@ -104,8 +108,18 @@ INTERLOCKED_HL_RESULT InterlockedHL_WaitForValue(int32_t volatile_atomic* addres
...
+            double elapsed_ms = timer_global_get_elapsed_ms() - start_time_ms;
+            /* Codes_SRS_INTERLOCKED_HL_27_013: [ If the elapsed time is greater than or equal to timeout_ms, InterlockedHL_WaitForValue shall return INTERLOCKED_HL_TIMEOUT. ] */
+            if (elapsed_ms >= (double)timeout_ms)
+            {
+                result = INTERLOCKED_HL_TIMEOUT;
+                break;
+            }
+
+            /* Codes_SRS_INTERLOCKED_HL_01_004: [ If the value at address_to_check is not equal to value_to_wait, InterlockedHL_WaitForValue shall wait until the value at address_to_check changes using wait_on_address with a timeout equal to timeout_ms minus the elapsed time. ]*/
+            uint32_t remaining_ms = timeout_ms - (uint32_t)elapsed_ms;
```
dcristoloveanu: [MrBot] **[major | correctness]** Cast `(uint32_t)elapsed_ms` is undefined behavior if `elapsed_ms` is negative

`elapsed_ms` can become negative if another thread calls `global_timer_state_reset()` mid-wait (it's a process-wide global), or if the timer reports an error as `-1.0`. Casting a negative `double` to `uint32_t` is **undefined behavior** in C (C11 §6.3.1.4¶1: "If the value of the integral part cannot be represented by the integer type, the behavior is undefined"). In practice it produces a huge `remaining_ms`, leading to a very long unintended wait.

Note the `elapsed_ms >= timeout_ms` check above evaluates to `false` for negative values, so it does NOT catch this case before the cast.

**Fix:** clamp before the cast — `if (elapsed_ms < 0.0) elapsed_ms = 0.0;`. Same issue in all 4 `InterlockedHL_WaitFor*` functions.

dcristoloveanu: [MrBot] **[minor | correctness]** Truncation of `elapsed_ms` can extend total wait by up to ~1 ms past `timeout_ms`

`(uint32_t)elapsed_ms` truncates toward zero. If `elapsed_ms = 999.4` and `timeout_ms = 1000`, then `remaining_ms = 1`, and the function can wait ~1 ms past the requested deadline. To be strictly conservative, round up: `(uint32_t)ceil(elapsed_ms)`. Probably acceptable, but worth deciding intentionally.

common/src/interlocked_hl.c
```diff
@@ -93,6 +94,9 @@ INTERLOCKED_HL_RESULT InterlockedHL_WaitForValue(int32_t volatile_atomic* addres
     {
         int32_t current_value;
 
+        /* Codes_SRS_INTERLOCKED_HL_27_001: [ InterlockedHL_WaitForValue shall call timer_global_get_elapsed_ms to determine the start time of the wait. ] */
+        double start_time_ms = timer_global_get_elapsed_ms();
```
dcristoloveanu, line 98: [MrBot] **[minor | maintainability]** Timeout-accounting block copy-pasted across 4 functions

The same start-time capture + elapsed compute + remaining compute + budget check appears in all four `InterlockedHL_WaitFor*` functions. Future fixes (e.g., the `UINT32_MAX` issue or the negative-elapsed clamp) will need to be replicated 4 times. Pre-existing `LogError` typos in this file already show that copy-paste mistakes here are a real risk.

**Suggestion:** extract a small helper, e.g.:
```c
static bool compute_remaining_timeout_ms(double start_time_ms, uint32_t timeout_ms, uint32_t* remaining_ms);
```
returning `false` when the budget is exhausted. Each function would then have a single call site.

dcristoloveanu, line 98: [MrBot] **[minor | performance]** Fast-path overhead: 1 wasted `timer_global_get_elapsed_ms` call when value matches immediately

`timer_global_get_elapsed_ms()` is invoked unconditionally before the loop, even when the very first `interlocked_add` shows the value already matches and no wait is needed. Cost per call:
- **Windows**: `InterlockedCompareExchange` on a process-wide global (`g_timer_state`) + `QueryPerformanceCounter` (~50–150 ns). The atomic on a single shared cache line can be observed by other timer callers process-wide as cross-core contention.
- **Linux**: `clock_gettime(CLOCK_MONOTONIC)` via vDSO (~30–80 ns).

The fast-path waste affects every caller, including hot-ish production paths (`sm.c:247`, `sm.c:521`, `completion_port_linux.c:467`, `threadpool_linux.c:444`).

**Suggested fix:** lazy-initialize `start_time_ms` inside the loop after the first value mismatch. This also (incidentally) preserves `UINT32_MAX` ⇒ "wait indefinitely" semantics on the first wait.

**Validation:** there is no perf microbenchmark for `InterlockedHL_*` (none under `interfaces/tests/gballoc_hl_perf` or similar). Recommend adding one (fast-path: value matches immediately; slow-path: many spurious wakeups) so future regressions are measurable.

common/tests/interlocked_hl_int/interlocked_hl_int.c
```diff
@@ -403,6 +403,74 @@ TEST_FUNCTION(interlocked_hl_wait_for_value_timest)
     // cleanup
 }
 
+/*
+Tests:
+InterlockedHL_WaitForValue timeout with non-matching value returns TIMEOUT and does not hang
+*/
+TEST_FUNCTION(interlocked_hl_wait_for_value_timeout_does_not_hang)
```
dcristoloveanu, line 412: [MrBot] **[major | testing]** New integration tests don't actually exercise the bug being fixed

These 4 tests just verify a single 500 ms timeout returns `INTERLOCKED_HL_TIMEOUT`. The pre-PR code would **also pass all 4 tests**, because no spurious wakeups occur — the address is never woken. The actual bug being fixed concerns *repeated spurious wakeups* extending the total wait beyond `timeout_ms`.

**Suggested fix:** add a test where a helper thread calls `wake_by_address_single` repeatedly (e.g., every 50 ms for 1 s) without changing the value to the target. Assert that `InterlockedHL_WaitForValue` returns `TIMEOUT` and the total wall-clock wait stays bounded (e.g., < 2× `timeout_ms`). That test would FAIL on master and PASS here, providing real regression coverage.

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -413,6 +426,34 @@ TEST_FUNCTION(when_the_wait_on_address_timesout_InterlockedHL_WaitForValue_also_
     ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_TIMEOUT, result);
 }
 
+/* Tests_SRS_INTERLOCKED_HL_27_002: [ InterlockedHL_WaitForValue shall call timer_global_get_elapsed_ms and compute the elapsed time since the start. ] */
+/* Tests_SRS_INTERLOCKED_HL_27_013: [ If the elapsed time is greater than or equal to timeout_ms, InterlockedHL_WaitForValue shall return INTERLOCKED_HL_TIMEOUT. ] */
+/* Tests_SRS_INTERLOCKED_HL_01_004: [ If the value at address_to_check is not equal to value_to_wait, InterlockedHL_WaitForValue shall wait until the value at address_to_check changes using wait_on_address with a timeout equal to timeout_ms minus the elapsed time. ] */
+TEST_FUNCTION(InterlockedHL_WaitForValue_timesout_when_elapsed_exceeds_timeout_after_wakeup)
```
dcristoloveanu, line 429: [MrBot] **[minor | testing]** Missing edge-case unit tests for the new timeout arithmetic

No tests for the inputs most likely to break the new code:
- `timeout_ms == 0` (should return `TIMEOUT` immediately on value mismatch)
- `timeout_ms == UINT32_MAX` (would surface the lost "wait indefinitely" semantic — see major comment in `interlocked_hl.c`)
- `elapsed_ms` exactly equal to `timeout_ms` (boundary)
- Timer mocked to return `-1.0` or to go backwards (would surface the UB cast — see major comment in `interlocked_hl.c`)

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
```diff
@@ -325,16 +327,20 @@ TEST_FUNCTION(when_the_value_equals_target_value_InterlockedHL_WaitForValue_retu
     ASSERT_ARE_EQUAL(INTERLOCKED_HL_RESULT, INTERLOCKED_HL_OK, result);
 }
 
-/* Tests_SRS_INTERLOCKED_HL_01_004: [ If the value at address_to_check is not equal to value_to_wait, InterlockedHL_WaitForValue shall wait until the value at address_to_check changes using wait_on_address. ] */
+/* Tests_SRS_INTERLOCKED_HL_01_004: [ If the value at address_to_check is not equal to value_to_wait, InterlockedHL_WaitForValue shall wait until the value at address_to_check changes using wait_on_address with a timeout equal to timeout_ms minus the elapsed time. ] */
 /* Tests_SRS_INTERLOCKED_HL_01_007: [ When wait_on_address succeeds, InterlockedHL_WaitForValue shall again compare the value at address_to_check with value_to_wait. ] */
+/* Tests_SRS_INTERLOCKED_HL_27_002: [ InterlockedHL_WaitForValue shall call timer_global_get_elapsed_ms and compute the elapsed time since the start. ] */
+/* Tests_SRS_INTERLOCKED_HL_27_013: [ If the elapsed time is greater than or equal to timeout_ms, InterlockedHL_WaitForValue shall return INTERLOCKED_HL_TIMEOUT. ] */
```
dcristoloveanu: [MrBot] **[minor | testing]** Misleading `Tests_SRS_27_013` tag on a success-path test

This `Tests_SRS_INTERLOCKED_HL_27_013` annotation claims coverage of "if elapsed >= timeout return `INTERLOCKED_HL_TIMEOUT`", but the test below (`when_the_value_equals_target_value_after_waiting_..._returns_OK`) returns `INTERLOCKED_HL_OK` and never enters that branch. The traceability tool will report false coverage.

The actual coverage is provided by `InterlockedHL_WaitForValue_timesout_when_elapsed_exceeds_timeout_after_wakeup` (line 432) — only that test should carry the `27_013` tag.

The same mistake is present for:
- `27_014` at line 496 (vs. real coverage at line 598)
- `27_015` at line 662 (vs. real coverage at line 759)
- `27_016` at line 823 (vs. real coverage at line 923)

common/devdoc/interlocked_hl_requirements.md
```diff
@@ -68,7 +68,13 @@ MOCKABLE_FUNCTION_WITH_RETURNS(, INTERLOCKED_HL_RESULT, InterlockedHL_WaitForVal
...
 **SRS_INTERLOCKED_HL_01_003: [** If the value at `address_to_check` is equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_OK`. **]**
 
-**SRS_INTERLOCKED_HL_01_004: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall wait until the value at `address_to_check` changes using `wait_on_address`. **]**
+**SRS_INTERLOCKED_HL_27_001: [** `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` to determine the start time of the wait. **]**
+
+**SRS_INTERLOCKED_HL_27_002: [** `InterlockedHL_WaitForValue` shall call `timer_global_get_elapsed_ms` and compute the elapsed time since the start. **]**
+
+**SRS_INTERLOCKED_HL_27_013: [** If the elapsed time is greater than or equal to `timeout_ms`, `InterlockedHL_WaitForValue` shall return `INTERLOCKED_HL_TIMEOUT`. **]**
+
+**SRS_INTERLOCKED_HL_01_004: [** If the value at `address_to_check` is not equal to `value_to_wait`, `InterlockedHL_WaitForValue` shall wait until the value at `address_to_check` changes using `wait_on_address` with a timeout equal to `timeout_ms` minus the elapsed time. **]**
```
dcristoloveanu, line 81: [MrBot] **[minor | requirements]** Existing SRS items now slightly inaccurate after this change

After this PR, `wait_on_address` is called with the *remaining* time, not the original `timeout_ms`. Several existing SRS items still describe the old behavior and should be updated:

- `SRS_INTERLOCKED_HL_11_001` (line 81)
- `SRS_INTERLOCKED_HL_05_005` (line 107)
- `SRS_INTERLOCKED_HL_11_002` (line 134)
- `SRS_INTERLOCKED_HL_05_011` (line 160)

All four say *"If `wait_on_address` hits the timeout specified in `timeout_ms`, ..."*. They should be reworded to reflect that the timeout passed to `wait_on_address` is now `timeout_ms - elapsed`, e.g. *"If `wait_on_address` hits the timeout passed to it, ... shall return `INTERLOCKED_HL_TIMEOUT`."*

dcristoloveanu (dismissed):
:shipit:

common/src/interlocked_hl.c
flash-sinx: Fixed. Added UINT32_MAX special case: when `timeout_ms == UINT32_MAX`, elapsed-time tracking is skipped and UINT32_MAX is passed directly to `wait_on_address`/`wait_on_address_64`, preserving the wait-indefinitely semantic per SRS_SYNC_43_008. Applied to all 4 functions. New specs: 27_017/018/019/020.

common/src/interlocked_hl.c
flash-sinx: Fixed. Added `if (elapsed_ms < 0.0) elapsed_ms = 0.0;` clamp before the cast in all 4 functions. New specs: 27_021/022/023/024. Added unit tests verifying that when the timer goes backwards, remaining_ms uses the full timeout.

flash-sinx: Acknowledged. The truncation is intentional — `(uint32_t)elapsed_ms` can overshoot by up to ~1ms, but the OS timer resolution on Windows is ~15.6ms anyway, so the imprecision is well within noise. Using `ceil()` would require `#include <math.h>` and `-lm` linkage, which isn't worth the tradeoff.

common/src/interlocked_hl.c
flash-sinx: Agreed this is a valid improvement, but extracting a helper changes the structure significantly and the 4-function duplication is pre-existing (not introduced by this PR). Deferring to a follow-up PR to keep this bug fix focused.

flash-sinx: The UINT32_MAX fix (27_017-020) naturally eliminates the fast-path overhead for the most common production callers (sm.c, completion_port_linux.c, threadpool_linux.c all pass UINT32_MAX). For finite-timeout callers, the cost is one `timer_global_get_elapsed_ms()` call (~50-150ns) which is negligible vs. the wait itself. Deferring lazy-init optimization to a follow-up if profiling shows it matters.

common/tests/interlocked_hl_int/interlocked_hl_int.c
flash-sinx: Fixed. Added `interlocked_hl_wait_for_value_with_spurious_wakeups_returns_timeout_in_bounded_time` integration test: a helper thread calls `wake_by_address_single` every 50ms without changing the value, then asserts that `InterlockedHL_WaitForValue` returns TIMEOUT within 5x the requested timeout (generous bound to avoid flakiness). This test would FAIL on the pre-fix code (~20s) and PASSES now (~1s).

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
flash-sinx: Fixed. Added edge-case unit tests: `timeout_ms == 0` (immediate timeout on mismatch), `elapsed == timeout` boundary (exact equality returns timeout), and negative timer (clamp to zero, remaining = full timeout). The UINT32_MAX case is now implicitly covered by existing UINT32_MAX tests which verify `wait_on_address` receives UINT32_MAX directly (spec 27_017).

common/tests/interlocked_hl_ut/interlocked_hl_ut.c
flash-sinx: Fixed. Removed `27_013/014/015/016` tags from the success-path tests (they never enter the timeout branch). The tags remain correctly on the dedicated timeout tests (`*_timesout_when_elapsed_exceeds_timeout_after_wakeup`). Also removed `27_002/005/008/011` tags from UINT32_MAX success-path tests since they no longer exercise the elapsed-compute code path.

common/devdoc/interlocked_hl_requirements.md
flash-sinx: Fixed. Updated all 4 specs (11_001, 05_005, 11_002, 05_011) from 'hits the timeout specified in timeout_ms' to 'hits the timeout'. Updated corresponding `Codes_SRS_` and `Tests_SRS_` comments to match.

ipg2013 (approved):

RohitG28 (approved):

PR #590, opened 2026-05-07
Fix wait_on_address_64 on Linux when only upper 32 bits change (use futex2)

parth21999:
## Summary

The Linux implementation of `wait_on_address_64` used the legacy `SYS_futex` syscall, which fundamentally compares only **32 bits** of the value at the address. When `*address` differed from `compare_value` only in the upper 32 bits (e.g. both lower 32 bits zero), the kernel's atomic check-before-sleep saw "equal" and slept until timeout, violating the [`wait_on_address_64` contract](https://github.com/Azure/c-pal/blob/master/c_pal_ll/interfaces/devdoc/sync_requirements.md#L25-L29) (SRS_SYNC_43_002).

In multi-threaded scenarios this also caused **lost wakeups** in `InterlockedHL_WaitForNotValue64` (and the other 64-bit `InterlockedHL` waiters): if a writer changed the upper 32 bits of the value between the function's `interlocked_add_64` read and its `wait_on_address_64` call, the kernel's 32-bit check missed the change and the waiter blocked until timeout.

The zrpc multiplexer integration tests hit this race because `SUBSTREAM_ID` is a 64-bit value whose lower 32 bits are an index starting at 0 — see [zrpc PR 15645356](https://msazure.visualstudio.com/One/_git/zrpc/pullrequest/15645356), which worked around the bug at the call site.

## Fix

Switch the 64-bit wait/wake helpers in `c_pal_ll/linux/src/sync_linux.c` to the Linux 6.7+ **futex2** syscalls (`SYS_futex_wait`, `SYS_futex_wake`) with `FUTEX2_SIZE_U64 | FUTEX2_PRIVATE`, which performs a true 64-bit atomic check-before-sleep. The 32-bit helpers are unchanged.

Notable details:
- The futex2 wait deadline is **absolute `CLOCK_MONOTONIC`** time (per the futex2 ABI), not relative.
- A `NULL` deadline is passed when `timeout_ms == UINT32_MAX` so the waiter sleeps indefinitely.
- Both wait and wake use the same `FUTEX2_SIZE_U64` flag so they hash to the same kernel futex queue.
- `#define` fallbacks for `SYS_futex_wake`, `SYS_futex_wait`, `FUTEX2_SIZE_U64`, `FUTEX2_PRIVATE` are added so the code compiles on glibc versions that don't yet expose these constants.

## Regression tests

Two deterministic tests were added that reproduce the bug on the legacy implementation (time out after 5 s) and pass on the new implementation:

| Test | File | What it does |
|---|---|---|
| `wait_on_address_64_returns_immediately_when_only_upper_32_bits_differ` | `c_pal_ll/interfaces/tests/sync_int/sync_int.c` | Single-threaded. Sets `*var = 0x100000000` then asserts `wait_on_address_64(&var, 0, 5000)` returns `WAIT_ON_ADDRESS_OK` promptly. |
| `interlocked_hl_wait_for_not_value_64_with_only_upper_32_bits_change` | `common/tests/interlocked_hl_int/interlocked_hl_int.c` | Replays the inner loop of `InterlockedHL_WaitForNotValue64` deterministically: reads with `interlocked_add_64`, changes the upper 32 bits, then calls `wait_on_address_64` with the original value as `compare_value`. |

## Verification

- ✅ Compiles cleanly on WSL (Ubuntu 22.04, glibc 2.35).
- ✅ `strace` confirms correct syscalls and arguments:
  - `syscall_0x1c7(addr, val=0, mask=~0, flags=0x83, &deadline, clockid=1)` → `SYS_futex_wait` with `FUTEX2_SIZE_U64 | FUTEX2_PRIVATE`, `CLOCK_MONOTONIC`
  - `syscall_0x1c6(addr, mask=~0, nr=1 or INT_MAX, flags=0x83)` → `SYS_futex_wake` with `FUTEX2_SIZE_U64 | FUTEX2_PRIVATE`
- ⚠️ The regression tests can't be exercised on WSL2 today because Microsoft's stable WSL kernel is still **6.6.87.2** (futex2 wait/wake landed in **6.7**). On older kernels the syscall returns `ENOSYS` and `wait_on_address_64` now returns `WAIT_ON_ADDRESS_ERROR`, surfacing the kernel-version requirement explicitly instead of silently sleeping incorrectly. CI/Linux runners on kernel ≥ 6.7 will run the tests for real.

## Behavior change

| Scenario | Before | After |
|---|---|---|
| `wait_on_address_64` with changed upper 32 bits, lower 32 bits unchanged | sleeps to timeout (silently wrong) | returns `OK` immediately (kernel ≥ 6.7) / returns `ERROR` with `ENOSYS` (kernel < 6.7) |
| `wait_on_address_64` with full-64-bit difference | works (lower bits already differ) | works |
| Normal wake-after-sleep paths | works | works |

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/c_pal_ll/interfaces/tests/sync_int/sync_int.c
+++ b/c_pal_ll/interfaces/tests/sync_int/sync_int.c
@@ -312,6 +312,54 @@ TEST_FUNCTION(wait_on_address_64_returns_immediately)
 
 /* Tests_SRS_SYNC_43_001: [ wait_on_address shall atomically compare *address and *compare_address.] */
 /* Tests_SRS_SYNC_43_002: [ wait_on_address shall immediately return true if *address is not equal to *compare_address.] */
+//
+// Regression test: wait_on_address_64 must compare the FULL 64-bit value at
+// *address against compare_value, not just the lower 32 bits.
+//
+// On Linux, wait_on_address_64 is implemented using the futex syscall, which
+// fundamentally compares only 32 bits of the value at the address (the lower
+// 32 bits on little-endian systems). When *address differs from compare_value
+// only in the upper 32 bits, the kernel sees the lower 32 bits as equal and
+// incorrectly sleeps until timeout instead of returning immediately.
+//
+// This bug also breaks InterlockedHL_WaitForNotValue64 (and the other 64-bit
+// InterlockedHL waiters) in race conditions, because those functions read the
+// full 64-bit value via interlocked_add_64, then call wait_on_address_64. If
+// another thread changes the upper 32 bits between the read and the syscall,
+// the kernel's atomic check-before-sleep fails to detect it, leading to a lost
+// wakeup that blocks until timeout.
+//
+// Concrete example from the multiplexer integration tests: a SUBSTREAM_ID is a
+// 64-bit value where the lower 32 bits are an index. The first substream has
+// index = 0, so the lower 32 bits of its SUBSTREAM_ID match the initial value
+// (0) used as compare_value, even though the upper 32 bits are non-zero.
+TEST_FUNCTION(wait_on_address_64_returns_immediately_when_only_upper_32_bits_differ)
+{
+    //arrange
+    volatile_atomic int64_t var;
+    // 0x100000000 has upper 32 bits = 1 and lower 32 bits = 0.
+    int64_t differing_value = 0x100000000LL;
+    (void)interlocked_exchange_64(&var, differing_value);
+    int64_t compare_value = 0; // Lower 32 bits match var's lower 32 bits but full 64 bits differ.
+    int32_t timeout = 5000;
+    double tolerance_factor = 0.1;
+
+    //act
+    double start_time = timer_global_get_elapsed_ms();
+    WAIT_ON_ADDRESS_RESULT return_val = wait_on_address_64(&var, compare_value, timeout);
+    double time_elapsed = timer_global_get_elapsed_ms() - start_time;
+
+    //assert
+    ASSERT_ARE_EQUAL(WAIT_ON_ADDRESS_RESULT, WAIT_ON_ADDRESS_OK, return_val,
+        "wait_on_address_64 must return WAIT_ON_ADDRESS_OK when *var (0x%" PRIx64 ") != compare_value (0x%" PRIx64 "), but it slept for %lf ms",
+        (uint64_t)differing_value, (uint64_t)compare_value, time_elapsed);
+    ASSERT_IS_TRUE(time_elapsed < timeout * tolerance_factor,
+        "wait_on_address_64 took too long: %lf ms (max expected %lf ms). It likely slept due to comparing only the lower 32 bits of the value.",
+        time_elapsed, timeout * tolerance_factor);
+}
+
+/* Tests_SRS_SYNC_43_001: [ wait_on_address shall atomically compare *address and *compare_address.] */
+/* Tests_SRS_SYNC_43_002: [ wait_on_address shall immediately return true if *address is not equal to *compare_address.] */
 /* Tests_SRS_SYNC_43_009: [ If timeout_ms milliseconds elapse, wait_on_address shall return false. ] */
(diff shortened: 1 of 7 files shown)


c_pal_ll/linux/src/sync_linux.c
```diff
@@ -17,6 +17,25 @@
...
+#endif
+#ifndef SYS_futex_wait
+#define SYS_futex_wait 455
+#endif
+#ifndef FUTEX2_SIZE_U64
+#define FUTEX2_SIZE_U64 0x03
+#endif
+#ifndef FUTEX2_PRIVATE
+#define FUTEX2_PRIVATE FUTEX_PRIVATE_FLAG /* 128 */
+#endif
```
parth21999: Why are these needed? where are they coming from?

c_pal_ll/linux/src/sync_linux.c
```diff
@@ -68,11 +87,34 @@ WAIT_ON_ADDRESS_RESULT wait_on_address_64(volatile_atomic int64_t* address, int6
...
+        {
+            deadline.tv_sec += 1;
+            deadline.tv_nsec -= 1000000000L;
+        }
+        deadline_p = &deadline;
+    }
 
-    /* Codes_SRS_SYNC_LINUX_05_002: [ wait_on_address_64 shall call syscall to wait on value at address to change to a value different than the one provided in compare_value. ] */
-    int syscall_result = syscall(SYS_futex, address, FUTEX_WAIT_PRIVATE, compare_value, &timeout, NULL, 0);
+    /* Codes_SRS_SYNC_LINUX_05_002: [ wait_on_address_64 shall call syscall(SYS_futex_wait) with FUTEX2_SIZE_U64 | FUTEX2_PRIVATE and a CLOCK_MONOTONIC absolute deadline, performing a true 64-bit atomic check-before-sleep. ] */
```
parth21999: >_05_002 [](http://example.com/codeflow?start=27&length=7)

All specs that are changed should be retagged with my devid 43. Changes are needed in devdoc, code and tests.

make sure repo_validation passes

c_pal_ll/interfaces/tests/sync_int/sync_int.c
```diff
@@ -310,6 +310,54 @@ TEST_FUNCTION(wait_on_address_64_returns_immediately)
...
+// InterlockedHL waiters) in race conditions, because those functions read the
+// full 64-bit value via interlocked_add_64, then call wait_on_address_64. If
+// another thread changes the upper 32 bits between the read and the syscall,
+// the kernel's atomic check-before-sleep fails to detect it, leading to a lost
+// wakeup that blocks until timeout.
+//
+// Concrete example from the multiplexer integration tests: a SUBSTREAM_ID is a
+// 64-bit value where the lower 32 bits are an index. The first substream has
+// index = 0, so the lower 32 bits of its SUBSTREAM_ID match the initial value
+// (0) used as compare_value, even though the upper 32 bits are non-zero.
```
parth21999: make this comment smaller.

c_pal_ll/interfaces/tests/sync_int/sync_int.c
```diff
@@ -310,6 +310,54 @@ TEST_FUNCTION(wait_on_address_64_returns_immediately)
...
+    // 0x100000000 has upper 32 bits = 1 and lower 32 bits = 0.
+    int64_t differing_value = 0x100000000LL;
+    (void)interlocked_exchange_64(&var, differing_value);
+    int64_t compare_value = 0; // Lower 32 bits match var's lower 32 bits but full 64 bits differ.
+    int32_t timeout = 5000;
+    double tolerance_factor = 0.1;
+
+    //act
+    double start_time = timer_global_get_elapsed_ms();
+    WAIT_ON_ADDRESS_RESULT return_val = wait_on_address_64(&var, compare_value, timeout);
```
parth21999: remove this timer. The result is enough

PR #592, opened 2026-05-15
Add helgrind suppressions for InterlockedHL_WaitForXXX functions

flash-sinx:
## Problem

PR #587 (InterlockedHL_WaitForXXX timeout accounting fix) restructured the loop body in the 4 WaitForXXX functions, which shifted the instruction layout enough to trigger a helgrind false positive on the atomic \interlocked_add\ / \interlocked_exchange\ pair used by \InterlockedHL_WaitForValue\ and \InterlockedHL_SetAndWake\.

The failure was observed in downstream zrpc propagation build (build 164315648), test \zrpc_v2_tcp_io_int_helgrind\:

\\\
Possible data race during write of size 4 by thread #1
   at InterlockedHL_WaitForValue (interlocked_hl.c:120)
This conflicts with a previous read of size 4 by thread #9
   at interlocked_exchange (interlocked_linux.c:206)
   by InterlockedHL_SetAndWake (interlocked_hl.c:519)
ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 40449 from 76)
\\\

All 6 tests passed — only the helgrind error count caused CTest to mark it failed. This is the same class of false positive already suppressed for other c-pal functions (e.g. \<incorrect interlocked flagging>\ for \interlocked_compare_exchange\).

## Fix

Added helgrind suppressions for all 4 \InterlockedHL_WaitForXXX\ functions to \c-pal_suppressions.sup\. These propagate to downstream repos via \VALGRIND_SUPPRESSIONS_FILE_EXTRA_PARAMETER\.

### DIFF
--- a/build_functions/c-pal_suppressions.sup
+++ b/build_functions/c-pal_suppressions.sup
@@ -63,3 +63,27 @@
    fun:pthread_create@@GLIBC_*
    ...
 }
+{
+   <incorrect interlocked flagging in InterlockedHL_WaitForValue>
+   Helgrind:Race
+   fun:InterlockedHL_WaitForValue
+   ...
+}
+{
+   <incorrect interlocked flagging in InterlockedHL_WaitForValue64>
+   Helgrind:Race
+   fun:InterlockedHL_WaitForValue64
+   ...
+}
+{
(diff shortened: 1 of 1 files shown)


dcristoloveanu (changes requested):
:clock1:

PR #612, opened 2026-06-23
jemalloc priming

anporumb:


win32/src/gballoc_ll_jemalloc.c
```diff
@@ -21,9 +21,24 @@ MU_STATIC_ASSERT(sizeof(uint32_t) == sizeof(unsigned));
...
     (void)params;
-    return 0;
+    /*Codes_SRS_GBALLOC_LL_JEMALLOC_02_011: [ gballoc_ll_init shall force jemalloc's one-time initialization to run on the calling thread by calling je_malloc. ]*/
+    void* prime = je_malloc(1);
+    if (prime == NULL)
+    {
+        /*Codes_SRS_GBALLOC_LL_JEMALLOC_02_012: [ If je_malloc fails then gballoc_ll_init shall fail and return a non-zero value. ]*/
+        LogError("failure in je_malloc(1)");
+        result = MU_FAILURE;
+    }
```
dcristoloveanu, line 33: This is funny. We did the same fix for C#. It might be possible to remove the C# "priming" allocation now if we have it in native.
We should check so that we have everything cleaner.

anporumb: created https://msazure.visualstudio.com/One/_sprints/taskboard/Azure%20Messaging%20Store/One/Custom/AzureMessaging/Rubidium/June2026?workitem=38538562

win32/tests/gballoc_ll_jemalloc_init_race_unprimed_int/gballoc_ll_jemalloc_init_race_unprimed_int.c
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// Demonstration of the defect that the jemalloc init-time priming fixes. Spawned children do NOT prime
+// jemalloc, so racing the process's first allocation hits jemalloc's lazy init concurrently and a
+// child crashes or hangs almost immediately. The shared driver spawns children until one fails, which
+// fails this test; it is registered WILL_FAIL (CTest inverts the expected failure to a pass). The
+// shared driver lives in gballoc_ll_jemalloc_init_race_common.c; the ONLY difference from
+// gballoc_ll_jemalloc_init_race_primed_int is the prime argument passed below.
+
```
dcristoloveanu, line 10: This is an interesting test. In a way it feels a bit overboard. As in we usually don't add tests which prove that without the fix something fails.

I could understand this for this case to sort of sniff when jemalloc would fix this issue. But that will be never, as we know.

So I am a little torn if we should add it since it feels a bit like dead weight, but, on the other hand it can't be that expensive...

dcristoloveanu (approved):
:shipit:

win32/tests/gballoc_ll_jemalloc_init_race_unprimed_int/gballoc_ll_jemalloc_init_race_unprimed_int.c
anporumb: It serves as a differentiator marker. Otherwise nobody would believe that jemalloc need "priming" and why now after so many years when we've been using it "as it was". So in a way archeological reasons. But yeah, it does fail fast, so cost is almost nothing.

PR #613, opened 2026-06-25
[MrBot] Update vcpkg to 2026.06.01; remove jemalloc priming (fixed in 5.3.1) and add init-race regression test

dcristoloveanu:
Updates `deps/vcpkg` to `2026.06.01` (jemalloc 5.3.0 -> 5.3.1) and handles the fallout from that bump:

- **vcpkg bump** - `deps/vcpkg` -> `2026.06.01` (`f3e1065`).
- **C4068 fix** - jemalloc 5.3.1's installed `jemalloc.h` uses `#pragma GCC`, which MSVC flags as `warning C4068` -> `error C2220` under `/WX`. The three `jemalloc.h` include sites are wrapped with a scoped `#pragma warning(disable: 4068)`.
- **Removed the jemalloc priming workaround** - jemalloc 5.3.1 makes the Windows/MSVC TSD statically thread-local (`__declspec(thread)`), so the per-thread wrapper is no longer allocated on first access and concurrent first allocations no longer race jemalloc's lazy init. `gballoc_ll_init` no longer primes; `SRS_GBALLOC_LL_JEMALLOC_02_011/012/013` and the priming init unit tests are removed.
- **Added `gballoc_ll_jemalloc_init_race_int`** - a regression test that spawns fresh child processes, each racing the process's first jemalloc allocation across multiple threads, and fails if any child crashes, hangs, or returns non-zero. Guards against a future jemalloc regression of the lazy-init race.

Work item 38196464.


win32/tests/gballoc_ll_jemalloc_init_race_int/gballoc_ll_jemalloc_init_race_int.c
```diff
@@ -91,11 +105,17 @@ static void run_one_first_allocation_race(void)
 
     (void)interlocked_exchange(&start_gate, 1);
 
+    int overall_result = 0;
     for (uint32_t index = 0; index < RACE_THREAD_COUNT; index++)
     {
-        int dont_care;
-        (void)ThreadAPI_Join(threads[index], &dont_care);
+        int thread_result;
+        (void)ThreadAPI_Join(threads[index], &thread_result);
```
dcristoloveanu: >ThreadAPI_Join [](http://example.com/codeflow?start=14&length=14)

If it returns something we should assert

dcristoloveanu: 🤖 MrBot: Good catch - ThreadAPI_Join now asserts it returns THREADAPI_OK (matching the ThreadAPI_Create check), and the join out-param (thread_result) is used for the per-thread result. See commit 562ddd4.

mattdurak (approved):
:shipit:

